PR #2, opened 2022-05-12
Initial commit of py_gtfs_rt_ingestion project

mzappitello:
The project was initialized with `poetry new` and will be the place to
store our pythonic ingestion of the gtfs json files.

We're initially launching with a pytest dev dependency that will run our
tests. It has the base initial test that checks version information.


py_gtfs_rt_ingestion/poetry.lock
```diff
@@ -0,0 +1,208 @@
+[[package]]
```
mzappitello, line 1: this file ensures that all developers are using the same versions of python and package dependencies.

py_gtfs_rt_ingestion/pyproject.toml
```diff
@@ -0,0 +1,16 @@
+[tool.poetry]
```
mzappitello, line 1: this file describes the project, its dependencies, and scripts that we'll want to run as we go.

from my understanding, it is largely edited using the `poetry` command line tool?

py_gtfs_rt_ingestion/tests/test_py_gtfs_rt_ingestion.py
```diff
@@ -0,0 +1,5 @@
+from py_gtfs_rt_ingestion import __version__
```
mzappitello, line 1: super basic initial test that just checks version information.

as the library is built, we will add tests here that do things like create temp parquet files to read and make assertions on.

.gitignore
```diff
@@ -8,3 +8,6 @@ erl_crash.dump
 *.beam
 /config/*.secret.exs
 .elixir_ls/
+
+
+/py_gtfs_rt_ingestion/*/__pycache__/
```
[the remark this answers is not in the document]
rymarczy: just having __pychache__ should catch any pycache folders in the whole project

PR #3, opened 2022-05-13
Vehicle transform test

mzappitello:
Add testing suite on top of @rymarczy 's `vehicle-transform` branch.

I moved a few things around a touch, but its largely the same. A few tests are added that we need to expand on and put into github actions, but the basic structure is there.


.tool-versions
```diff
@@ -1,2 +1,2 @@
 poetry 1.1.13
-python 3.7.12
+python 3.9.12
```
mzappitello, line 2: bumped py version to the latest supported by lambda. (this will also get rid of my pyarrow nightmares)

py_gtfs_rt_ingestion/ingest.py
```diff
@@ -0,0 +1,78 @@
...
+    parser.add_argument(
+        '--fname',
+        dest='fname',
+        type=str,
+        required=True,
+        help='provide filename to ingest')
+
+    parsed_args = parser.parse_args(args)
+
+    return parsed_args
```
mzappitello, line 27: Moved this into its own method. This way we can test our argparser all on its own and expand on it. It also means that the `run` method is taking in a filename rather than a list of sysvar args that need parsing.

py_gtfs_rt_ingestion/ingest.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def run(filename) -> None:
```
mzappitello, line 30: now that this method is taking a filename, it should be easier to test on its own.

py_gtfs_rt_ingestion/ingest.py
```diff
@@ -0,0 +1,78 @@
...
+    if 'https_cdn.mbta.com_realtime_VehiclePositions_enhanced' not in filename:
+        raise Exception('Requires VehiclePostitions file.')
+
+    fname = Path(filename)
```
mzappitello, line 34: this validation stuff should probably be moved into `parseArgs` but we can get to that as we're going.

py_gtfs_rt_ingestion/ingest.py
```diff
@@ -0,0 +1,78 @@
...
+    schema = get_vehicle_schema()
+
+    pa_table_dict = {k.name:[] for k in schema}
+
+    with gzip.open(fname, 'rb') as f:
```
mzappitello, line 39: all of this looks good to me. i think at some point we're going to want to make it more generic so that we can use the same method on multiple input types, but this is a fine starting point.

py_gtfs_rt_ingestion/ingest.py
```diff
@@ -0,0 +1,78 @@
...
+    pq.write_to_dataset(
+        pa_table,
+        root_path='./sample_data',
+        partition_cols=['year','month','day','hour']
+    )
+
+
+if __name__ == '__main__':
+    args = parseArgs(sys.argv[1:])
+    run(args)
```
mzappitello, line 78: explicitly parse args here.

py_gtfs_rt_ingestion/poetry.lock
```diff
@@ -53,6 +53,14 @@ category = "dev"
 optional = false
 python-versions = ">=3.5"
 
+[[package]]
```
mzappitello, line 56: this file is going to get update each time we update the `pyproject.toml` file. i think best practices say to keep it commited so that we know we're on the same version of everything.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/helpers.py
```diff
@@ -0,0 +1,42 @@
+from datetime import datetime
```
mzappitello, line 1: all LGTM.

mzappitello: i did move it into this dir so that we can access it via this module rather than lib.

i think we should make some naming decisions soon. part of this is my own ignorance on not remembering all of the different acronyms we're supposed to keep track of.

py_gtfs_rt_ingestion/pyproject.toml
```diff
@@ -6,7 +6,8 @@ authors = ["MBTA CTD <developer@mbta.com>",
            "Mike Zappitello <mzappitello@mbta.com>"]
 
 [tool.poetry.dependencies]
-python = "^3.7"
+python = "^3.9"
+pyarrow = "^8.0.0"
```
mzappitello, line 10: update python and add pyarrow to our list of dependencies.

i used to add pyarrow `poetry add pyarrow` on the command line. i think that edits the dot toml file and also invokes `poetry install` which uses `pip` to add pyarrow to the virtualenv, also installing pyarrows dependencies, and adding all of them to the poetry dot lock file.

py_gtfs_rt_ingestion/tests/test_ingest.py
```diff
@@ -0,0 +1,19 @@
+from ingest import parseArgs, run
```
mzappitello, line 1: super duper basic testing stub.

what we want to do here is create a suit of tests that will check that our functions are all doing the things we hope they will.

right now `test_argparse` is just testing out that we're parsing a string in the fname arg. in the future, we'll want to put validation stuff in there, and maybe add some kind of argparse action to let us make selections on schemas based on input filenames. all that will get tested in this block, as well as tests against bad inputs.

py_gtfs_rt_ingestion/tests/test_ingest.py
```diff
@@ -0,0 +1,19 @@
...
+    Test that the custom argparse is working as expected
+    """
+    dummy_file_path = 'my/fake/path'
+
+    dummy_arguments = ['--fname', dummy_file_path]
+    args = parseArgs(dummy_arguments)
+
+    assert args.fname == dummy_file_path
+
+def test_file_conversion():
```
mzappitello, line 14: this is the stub where we're going to do the bulk of our testing. obviously, nothing happening right now, but i hope to build this out in earnest on monday.

PR #5, opened 2022-05-17
Re factor move to lib

rymarczy:
Re-factor Configuration class to standard class. 

config_base.py added to hold Enumeration of ConfigType and ABC for ConfigDetail. 
- Each ConfigType should implement ConfigDetail Class in separate file for import into Configuration class.

Configuration.detail assigned to ConfigDetail ABC Class. 

Class/Methods added to Configuration class as helper to pull information for Configuration.detail ABC class:
- Configuration.config_type -> ConfigType Enum
- Configuration.export_schema -> pyarrow export schema
- Configuration.record_from_entity() -> processed parquet transformation dictionary.

### DIFF
--- a/py_gtfs_rt_ingestion/ingest.py
+++ b/py_gtfs_rt_ingestion/ingest.py
@@ -46,10 +46,9 @@ def convert_json_to_parquet(input_filename: str, output_dir: str) -> None:
     * append to the table for each element the entities list in the json file
     * write the table
     """
-    config = Configuration.from_filename(input_filename)
+    config = Configuration(input_filename)
 
-    schema = config.get_schema()
-    table = {key.name:[] for key in schema}
+    table = {key.name:[] for key in config.export_schema}
 
     with gzip.open(Path(input_filename), 'rb') as f:
         json_data = json.loads(f.read())
@@ -74,7 +73,7 @@ def convert_json_to_parquet(input_filename: str, output_dir: str) -> None:
                 table[key].append(record[key])
 
     pq.write_to_dataset(
-        pa.table(table, schema=schema),
+        pa.table(table, schema=config.export_schema),
         root_path=output_dir,
         partition_cols=['year','month','day','hour']
     )
--- a/py_gtfs_rt_ingestion/ingest.py
+++ b/py_gtfs_rt_ingestion/ingest.py
@@ -11,7 +11,7 @@ import pyarrow.parquet as pq
 from datetime import datetime
 from pathlib import Path
 
-from py_gtfs_rt_ingestion import Configuration
+from py_gtfs_rt_ingestion.configuration import Configuration
 
 DESCRIPTION = "Convert a json file into a parquet file. Used for testing."
 
--- a/py_gtfs_rt_ingestion/ingest.py
+++ b/py_gtfs_rt_ingestion/ingest.py
@@ -11,7 +11,7 @@ import pyarrow.parquet as pq
 from datetime import datetime
 from pathlib import Path
 
-from py_gtfs_rt_ingestion.configuration import Configuration
+from py_gtfs_rt_ingestion import Configuration
 
 DESCRIPTION = "Convert a json file into a parquet file. Used for testing."
 
--- a/py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/__init__.py
+++ b/py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/__init__.py
@@ -2,4 +2,3 @@ __version__ = '0.1.0'
 
 from .configuration import Configuration
 from .config_base import ConfigType
-from .config_base import ConfigDetail
--- a/py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/__init__.py
+++ b/py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/__init__.py
@@ -1,3 +1,5 @@
 __version__ = '0.1.0'
 
 from .configuration import Configuration
+from .config_base import ConfigType
+from .config_base import ConfigDetails
--- a/py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/__init__.py
+++ b/py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/__init__.py
@@ -2,4 +2,4 @@ __version__ = '0.1.0'
 
 from .configuration import Configuration
 from .config_base import ConfigType
-from .config_base import ConfigDetails
+from .config_base import ConfigDetail
--- a/py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/config_base.py
+++ b/py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/config_base.py
@@ -23,7 +23,7 @@ class ConfigDetail(ABC):
     """
     Abstract Base Class for all ConfigDetail implementations.
 
-    ConfigDetails classes must implement all methods and properties that are defined.
+    ConfigDetail classes must implement all methods and properties that are defined.
     """
     @property
     @abstractmethod
--- a/py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/config_base.py
+++ b/py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/config_base.py
@@ -1,0 +1,38 @@
+from abc import ABC
+from abc import abstractmethod
+
+from enum import auto
+from enum import Enum
+
+import pyarrow
+
+class ConfigType(Enum):
+    """
+    ConfigType is and Enumaration taht is inclusive of all 
+    configurations types to be processed by this library
+
+    """
+    RT_ALERTS = auto()
+    RT_TRIP_UPDATES = auto()
+    RT_VEHICLE_POSITIONS = auto()
+    BUS_TRIP_UPDATES = auto()
+    BUS_VEHICLE_POSITIONS = auto()
+    VEHICLE_COUNT = auto()
+
+class ConfigDetails(ABC):
+    """
+    Abstract Base Class for all Configuration Details implementations.
+
+    Configuration Details classes must implement all methods and properties
+    as defined by ConfigDetails.
+    """
+    @property
+    @abstractmethod
+    def config_type(self) -> ConfigType: ...
+
+    @property
+    @abstractmethod
+    def export_schema(self) -> pyarrow.schema: ...
+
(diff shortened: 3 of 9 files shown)


py_gtfs_rt_ingestion/ingest.py
```diff
@@ -46,10 +46,9 @@ def convert_json_to_parquet(input_filename: str, output_dir: str) -> None:
     * append to the table for each element the entities list in the json file
     * write the table
     """
-    config = Configuration.from_filename(input_filename)
+    config = Configuration(input_filename)
```
rymarczy, line 49: Pass filename directly into Configuration class for initialization

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/__init__.py
```diff
@@ -1,3 +1,5 @@
 __version__ = '0.1.0'
 
 from .configuration import Configuration
+from .config_base import ConfigType
+from .config_base import ConfigDetail
```
rymarczy, line 5: Probably don't need ConfigDetail in here.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/config_base.py
```diff
@@ -0,0 +1,37 @@
+from abc import ABC
+from abc import abstractmethod
+
+from enum import auto
+from enum import Enum
+
+import pyarrow
+
+class ConfigType(Enum):
```
rymarczy, line 9: ConfigType Enums separated as basic Enum Class.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/config_rt_alerts.py
```diff
@@ -0,0 +1,17 @@
+import pyarrow
+
+from .config_base import ConfigDetail
+from .config_base import ConfigType
+
+class RtAlertsDetail(ConfigDetail):
```
rymarczy, line 6: Barebones ConfigDetail implementation for RT_ALERTS

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/config_rt_trip.py
```diff
@@ -0,0 +1,17 @@
+import pyarrow
+
+from .config_base import ConfigDetail
+from .config_base import ConfigType
+
+class RtTripDetail(ConfigDetail):
```
rymarczy, line 6: Barebones ConfigDetail implementation for RT_TRIP_UPDATES

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/config_rt_vehicle.py
```diff
@@ -0,0 +1,57 @@
+import pyarrow
+
+from .config_base import ConfigDetail
+from .config_base import ConfigType
+
+class RtVehicleDetail(ConfigDetail):
```
rymarczy, line 6: Full ConfigDetail implementation for RT_VEHICLE_POSITIONS.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/configuration.py
```diff
@@ -13,71 +19,32 @@ class Configuration(Enum):
...
-                    ('route_id', pyarrow.string()),
-                    ('direction_id', pyarrow.int8()),
-                    ('schedule_relationship', pyarrow.string()),
-                    ('start_date', pyarrow.string()),
-                    ('start_time', pyarrow.string())])),
-                ('consist_labels', pyarrow.list_(pyarrow.string()))
-            ])
-
-        raise Exception("No Current Schema for %s" % self.name)
+    def __init__(self, filename: str) -> None:
```
rymarczy, line 22: Initialize Configuration object with filename. 

Filename stored as pathlib.Path object. 
filename compared to known implementations of ConfigDetail, if ConfigDetail is found, save as self.detail, else raise Exception.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/configuration.py
```diff
@@ -13,71 +19,32 @@ class Configuration(Enum):
...
+        elif 'mbta.com_realtime_VehiclePositions_enhanced' in filename:
+            self.detail = RtVehicleDetail()
+        else:
+            raise Exception("Bad Configuration from filename %s" % filename)
+
+    @property
+    def config_type(self) -> ConfigType:
+        return self.detail.config_type
+
+    @property
```
rymarczy, line 42: Convenience property for accessing export_schema of self.detail.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/configuration.py
```diff
@@ -13,71 +19,32 @@ class Configuration(Enum):
...
+    def export_schema(self) -> pyarrow.schema:
+        return self.detail.export_schema
 
     def record_from_entity(self, entity: dict) -> dict:
```
rymarczy, line 46: Convenience method for accessing record_from_entity of self.detail.

py_gtfs_rt_ingestion/tests/test_configuration.py
```diff
@@ -2,37 +2,41 @@
...
 from py_gtfs_rt_ingestion import Configuration
+from py_gtfs_rt_ingestion import ConfigType
 
 def test_filname_parsing():
     """
     Check that we are able to get the correct Configuration type for multiple
     filenames
     """
-    trip_updates_type = Configuration.from_filename(
+    trip_updates_type = Configuration(
```
rymarczy, line 12: Update test to work with re-factored Configuration class.

py_gtfs_rt_ingestion/tests/test_configuration.py
```diff
@@ -2,37 +2,41 @@
...
-    config = Configuration.RT_VEHICLE_POSITIONS
-    schema = config.get_schema()
+    config = Configuration(
+        '2022-01-01T00:00:03Z_https_cdn.mbta.com_realtime_VehiclePositions_enhanced.json.gz')
+    schema = config.export_schema
 
-    with pytest.raises(Exception):
-        unimpled_config = Configuration.RT_ALERTS
-        unimpled_schema = unimpled_config.get_schema()
+    # Requires full implemenation of ConfigDetail for ConfigType
```
rymarczy, line 32: Remove test, for Configuration class re-factor. Should expect entire ConfigDetail to be implemented, so no exception raised.

py_gtfs_rt_ingestion/ingest.py
```diff
@@ -46,10 +46,9 @@ def convert_json_to_parquet(input_filename: str, output_dir: str) -> None:
...
 
-    schema = config.get_schema()
-    table = {key.name:[] for key in schema}
+    table = {key.name:[] for key in config.export_schema}
```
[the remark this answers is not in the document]
mzappitello: i like this change.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/configuration.py
```diff
@@ -13,71 +19,32 @@ class Configuration(Enum):
...
-                    ('direction_id', pyarrow.int8()),
-                    ('schedule_relationship', pyarrow.string()),
-                    ('start_date', pyarrow.string()),
-                    ('start_time', pyarrow.string())])),
-                ('consist_labels', pyarrow.list_(pyarrow.string()))
-            ])
-
-        raise Exception("No Current Schema for %s" % self.name)
+    def __init__(self, filename: str) -> None:
+        self.file_path = Path(filename)
```
mzappitello, line 23: i don't think we want a `Confiiguration` object to hold onto the path. i imagine we'll use the came `Configuration` to process a number of input files.

i think its fine to initialize with a filepath. we could also have a static method that derives the `ConfigType` from a filepath and we use that to initialize the `Configuration`.

py_gtfs_rt_ingestion/tests/test_configuration.py
mzappitello: https://github.com/mbta/lamp/pull/5/files#r875077704

this change would make this feel a bit more natural to me as well.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/configuration.py
rymarczy: Agree, no reason to save the file path right now.

We haven't really talked about how we'll implement the processing of multiple files, but we'll need some kind of container object which I imagine will be responsible for keeping a list of files that were downloaded/processed and concatenating all of the data extracted from the gzip files into one object.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/__init__.py
mzappitello: yeah. i think it should be taken out, but there aren't really hard, accepted rules here.

i tend to think of it as "What stuff would I want to use outside of this library". it includes the added benefit of giving us flexibility to change the underlying structure, but keep the API consistent by changing where the imports are coming from. its not a big concern here given we're the only consumers and its unlikely we're going to even touch this module much after we get the lambda stuck kicking off.

my general feelings about `__init__.py` are pretty well captured in this old reddit thread.

https://old.reddit.com/r/Python/comments/1bbbwk/whats_your_opinion_on_what_to_include_in_init_py/

py_gtfs_rt_ingestion/tests/test_configuration.py
rymarczy: Any `ConfigDetail` Class that doesn't implement all of the classes defined in the `config_base` ABC should automatically raise an exception when it is invoked. I'm just not sure how exactly we would structure a test for that.

mzappitello:
For the most part, this looks good to me. My only main concern is using the filename in the `Configuration` class constructor, since this will mean we need to have a file around to make one, and it implies that we have one `Configuration` object per file, when we really ought to have one per file type.

After that, I think this is ready to merge straight into `main`.

mzappitello:
actually, scratch merging this directly into `main`. we'll want to do a "Squash and Merge" into `move_to_lib` so that we can clean up the commit history a bit.

PR #6, opened 2022-05-17
Add convert to parquet tests

mzappitello:
* Load a single vehicle data dot json dot gz file and coonvert it
  parquet.
* check that its columns are correct based on incoming json data
* check that the correct number of records were made
* check that partitioning was done as expected

* add a test files directory to tests that has the file we're converting
* make convert_json_to_parqute take Path objects instead of strings


py_gtfs_rt_ingestion/ingest.py
```diff
@@ -36,7 +36,7 @@ def parseArgs(args) -> dict:
     return parsed_args
 
 
-def convert_json_to_parquet(input_filename: str, output_dir: str) -> None:
+def convert_json_to_parquet(input_filename: Path, output_dir: Path) -> None:
```
mzappitello, line 39: changed func signature.

py_gtfs_rt_ingestion/ingest.py
```diff
@@ -46,11 +46,11 @@ def convert_json_to_parquet(input_filename: str, output_dir: str) -> None:
     * append to the table for each element the entities list in the json file
     * write the table
     """
-    config = Configuration(input_filename)
+    config = Configuration(input_filename.name)
```
mzappitello, line 49: the from filename stuff is expecting a string and i didn't want to change it.

py_gtfs_rt_ingestion/ingest.py
```diff
@@ -81,5 +81,5 @@ def convert_json_to_parquet(input_filename: str, output_dir: str) -> None:
 
 if __name__ == '__main__':
     args = parseArgs(sys.argv[1:])
-    convert_json_to_parquet(input_filename=args.input_file,
-                            output_dir=args.output_dir)
+    convert_json_to_parquet(input_filename=Path(args.input_file),
+                            output_dir=Path(args.output_dir))
```
mzappitello, line 85: this should probably be handled in parseArgs but it felt out of scope.

py_gtfs_rt_ingestion/tests/test_ingest.py
```diff
@@ -1,4 +1,11 @@
-from ingest import parseArgs
+from pathlib import Path
+import logging
+import pyarrow.parquet as pq
+
+from ingest import parseArgs, convert_json_to_parquet
+
+LOGGER = logging.getLogger(__name__)
```
mzappitello, line 7: added this for debugging.

py_gtfs_rt_ingestion/tests/test_ingest.py
```diff
@@ -1,4 +1,11 @@
...
+TEST_FILE_DIR = Path(__file__).parent.joinpath("test_files")
```
mzappitello, line 8: where to find our test files. we might eventually want this to live in a helper that all the tests can see.

py_gtfs_rt_ingestion/tests/test_ingest.py
```diff
@@ -14,9 +21,64 @@ def test_argparse():
     assert args.input_file == dummy_file_path
     assert args.output_dir == dummy_output_dir
 
-def test_file_conversion():
+def test_file_conversion(tmpdir):
```
mzappitello, line 24: using the `tmpdir` factory from pytest so that we can dump data and then have it cleaned up after running tests.

py_gtfs_rt_ingestion/tests/test_ingest.py
```diff
@@ -14,9 +21,64 @@ def test_argparse():
...
     """
     TODO - convert a dummy json data to parquet and check that the new file
     matches expectations
     """
-    assert True
+    rt_vehicle_positions_file = TEST_FILE_DIR.joinpath(
```
mzappitello, line 29: pretty self explanatory i think.

rymarczy (approved):
All looks good to me!

PR #8, opened 2022-05-19
Ingest multiprocess

rymarczy:
Add `gz_to_pyarrow` function. Responsible for threaded conversion of gzip json to pyarrow table.
- Must return either pyarrow table or filename (on conversion failure)

Added `lambda_handler` as entry point for lambda function. Launches multiprocessing conversion is async call to `gz_to_pyarrow`.

### DIFF
--- a/py_gtfs_rt_ingestion/ingest.py
+++ b/py_gtfs_rt_ingestion/ingest.py
@@ -9,11 +9,120 @@ import pyarrow as pa
 import pyarrow.parquet as pq
 
 from datetime import datetime
+from multiprocessing import Pool
+from multiprocessing import Queue
 from pathlib import Path
 
 from py_gtfs_rt_ingestion import Configuration
 
+
 DESCRIPTION = "Convert a json file into a parquet file. Used for testing."
+MULTIPROCESSING_POOL_SIZE = 4
+
+def gz_to_pyarrow(filename: str, config: Configuration, return_queue: Queue) -> None:
+    """
+    Accepts filename as string for gzip json to pyarrow table parsing. 
+
+    Converted pyarrow table sent back to main process using return_queue.
+    If conversion failes, returns filename of failed conversion.
+    """
+    # Enclose entire function in try block, Process can't fail silently.
+    try:
+        with gzip.open(filename, 'rb') as f:
+            json_data = json.loads(f.read())
+
+        # parse timestamp info out of the header
+        header = json_data['header']
+        feed_timestamp = header['timestamp']
+        timestamp = datetime.utcfromtimestamp(feed_timestamp)
+
+        # create dict of lists 
+        ret_obj = {key.name:[] for key in config.export_schema}
+
+        # for each entity in the list, create a record and add it to the table
+        for entity in json_data['entity']:
+            record = config.record_from_entity(entity=entity)
+            record.update({
+                'year': timestamp.year,
+                'month': timestamp.month,
+                'day': timestamp.day,
+                'hour': timestamp.hour,
+                'feed_timestamp': feed_timestamp
+            })
+
+            for key in record:
+                ret_obj[key].append(record[key])
+
+        ret_obj = pa.table(ret_obj, schema=config.export_schema)
+
+    # Catch Process failure, send back filename for handling
+    except Exception as e:
+        print(e, flush=True)
+        ret_obj = filename
+
+    # Send pyarrow table or None back to main Process for concatenation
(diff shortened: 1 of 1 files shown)


py_gtfs_rt_ingestion/ingest.py
```diff
@@ -9,11 +9,121 @@
...
+def gz_to_pyarrow(filename: str, config: Configuration, return_queue: Queue) -> None:
+    """
+    Accepts filename as string. Converts gzipped json -> pyarrow table. 
+
+    parrow table sent back to main process using return_queue.
+    If conversion fails, returns filename.
+    """
+    # Enclose entire function in try block, Process can't fail silently.
+    # Must return either pyarrow table or filename
+    try:
```
rymarczy, line 31: Put all logic of function in `try` block. Multiprocess must return either pyarrow table or filename (if conversion failed).

py_gtfs_rt_ingestion/ingest.py
```diff
@@ -9,11 +9,121 @@
...
+                'month': timestamp.month,
+                'day': timestamp.day,
+                'hour': timestamp.hour,
+                'feed_timestamp': feed_timestamp
+            })
+
+            for key in record:
+                ret_obj[key].append(record[key])
+
+        ret_obj = pa.table(ret_obj, schema=config.export_schema)
```
rymarczy, line 57: Convert to pyarrow table in child process.

py_gtfs_rt_ingestion/ingest.py
```diff
@@ -9,11 +9,121 @@
...
+
+    # Catch failure, send back filename for handling
+    except Exception as e:
+        print(e, flush=True)
+        ret_obj = filename
```
rymarczy, line 62: On failure, send filename back through `return_queue`

py_gtfs_rt_ingestion/ingest.py
```diff
@@ -9,11 +9,121 @@
...
+        raise Exception("Lambda event missing 'file_batch' key.")
+
+    file_batch = event['file_batch']
+
+    config = Configuration(file_batch[0])
+
+    return_queue = Queue()
+
+    pool = Pool(MULTIPROCESSING_POOL_SIZE)
+    pool.starmap_async(gz_to_pyarrow, [(f, config, return_queue) for f in file_batch])
```
rymarczy, line 95: Set up multiprocessing, create Queue for return data and Pool object. 

Call `starmap_async` so process work does not block this process collecting data from `return_queue`.

py_gtfs_rt_ingestion/ingest.py
```diff
@@ -9,11 +9,121 @@
...
+    pa_table = None
+    failed_ingestion = []
+    for _ in range(len(file_batch)):
+        ret_obj = return_queue.get()
+        if pa_table is None and not isinstance(ret_obj, str):
+            pa_table = ret_obj
+        elif not isinstance(ret_obj, str):
+            pa.concat_tables([pa_table,ret_obj])
+        else:
+            failed_ingestion.append(ret_obj)
```
rymarczy, line 107: Collect data from return Queue, save filenames of failed conversion for later.

Concat all pyarrow tables.

py_gtfs_rt_ingestion/ingest.py
```diff
@@ -9,11 +9,121 @@
...
+    # TODO:
+    # 1. Implement output directory as env variable for lamda function.
+    # 2. Do something with failed conversion filenames.
+    OUTPUT_DIR = ''
+
+    pq.write_to_dataset(
+        pa_table,
+        root_path=OUTPUT_DIR,
+        partition_cols=['year','month','day','hour']
+    )
```
rymarczy, line 124: Write out concat pyarrow table to `OUTPUT_DIR`.

mzappitello:
This looks good to me.

I think we might want to move some of it to the library, but we can do that later.

py_gtfs_rt_ingestion/ingest.py
```diff
@@ -9,11 +9,121 @@
...
+    # Catch failure, send back filename for handling
+    except Exception as e:
+        print(e, flush=True)
+        ret_obj = filename
+
+    # Send pyarrow table or filename back to main Process for concatenation
+    return_queue.put(ret_obj)
+    return_queue.close()
+
+def lambda_handler(event: dict, context) -> None:
```
mzappitello, line 68: it would be kind of cool if this is what we were calling when we invoke the script too.

PR #9, opened 2022-05-19
add batching script utility

mzappitello:
Add `Batch` class and a function to build them out of an generator of filename, filesize tuples.

Add a script that uses that method and can do batch processing using this method. It can take in either local file arguments or s3 arguments.

### DIFF
--- a/py_gtfs_rt_ingestion/batch_files.py
+++ b/py_gtfs_rt_ingestion/batch_files.py
@@ -1,7 +1,10 @@
 #!/usr/bin/env python
 
 import argparse
+import boto3
 import sys
+
+from collections.abc import Iterable
 
 from py_gtfs_rt_ingestion import batch_files
 
@@ -13,14 +16,18 @@ def parseArgs(args) -> dict:
         '--input',
         dest='input_file',
         type=str,
-        required=True,
         help='text file of AWS S3 ls call')
+
+    parser.add_argument(
+        '--s3-prefix',
+        dest='s3_prefix',
+        type=str,
+        help='prefix to files in the mbta-gtfs-s3 bucket')
 
     parser.add_argument(
         '--threshold',
         dest='filesize_threshold',
         type=int,
-        required=True,
         default=100000,
         help='filesize threshold for batch sizes')
 
@@ -28,23 +35,38 @@ def parseArgs(args) -> dict:
 
     return parsed_args
 
-def file_list_from_file(input_file):
+def file_list_from_file(input_file: str) -> Iterable[(str, int)]:
     """
-    take a file that is the output of `aws s3 ls <dir>` and yield out objects
-    one at a time
+    take a file that is the output of `aws s3 ls <dir>` and yield out filename,
+    filesize tuples.
     """
-    for line in open(input_file):
-        yield line
+    for file_info in open(input_file):
+        (date, time, size, filename) = file_info.split()
+        yield (filename, sie)
 
-def make_file_list(args):
+def file_list_from_s3(file_prefix) -> Iterable[(str, int)]:
+    """
+    look in the mbta-gtfs-s3 bucket for files begining with this prefix and
+    yield out filename, filesize tuples.
+    """
+    s3 = boto3.resource('s3')
+    bucket = s3.Bucket('mbta-gtfs-s3')
+
+    for bucket_object in bucket.objects.filter(Prefix=file_prefix):
+        yield (bucket_object.key, bucket_object.size)
+
+def make_file_list(args: dict) -> Iterable[(str, int)]:
+    """
+    based on args, yield out filename, filesize tuples either from a local file
+    or from our s3 bucket.
+    """
     if args.input_file:
         return file_list_from_file(args.input_file)
 
-    if args.s3_filepath:
-        raise Exception("haven't impl'ed yet")
+    if args.s3_prefix:
+        return file_list_from_s3(args.s3_prefix)
 
     raise Exception("shouldn't get here")
-
 
 if __name__ == '__main__':
     args = parseArgs(sys.argv[1:])
--- a/py_gtfs_rt_ingestion/batch_files.py
+++ b/py_gtfs_rt_ingestion/batch_files.py
@@ -1,0 +1,54 @@
+#!/usr/bin/env python
+
+import argparse
+import sys
+
+from py_gtfs_rt_ingestion import batch_files
+
+DESCRIPTION = "Generate batches of json files that should be processesd"
+
+def parseArgs(args) -> dict:
+    parser = argparse.ArgumentParser(description=DESCRIPTION)
+    parser.add_argument(
+        '--input',
+        dest='input_file',
+        type=str,
+        required=True,
+        help='text file of AWS S3 ls call')
+
+    parser.add_argument(
+        '--threshold',
+        dest='filesize_threshold',
+        type=int,
+        required=True,
+        default=100000,
+        help='filesize threshold for batch sizes')
+
+    parsed_args = parser.parse_args(args)
+
+    return parsed_args
+
+def file_list_from_file(input_file):
+    """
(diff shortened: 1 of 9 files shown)


py_gtfs_rt_ingestion/batch_files.py
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env python
```
mzappitello, line 1: new script at the top level. this is going to eventually become the lambada function for scheduling things.

py_gtfs_rt_ingestion/batch_files.py
```diff
@@ -0,0 +1,76 @@
...
+def file_list_from_s3(file_prefix) -> Iterable[(str, int)]:
+    """
+    look in the mbta-gtfs-s3 bucket for files begining with this prefix and
+    yield out filename, filesize tuples.
+    """
+    s3 = boto3.resource('s3')
+    bucket = s3.Bucket('mbta-gtfs-s3')
+
+    for bucket_object in bucket.objects.filter(Prefix=file_prefix):
+        yield (bucket_object.key, bucket_object.size)
```
mzappitello, line 56: generator functions that produce tuples of filenames and filesizes that will be batched.

py_gtfs_rt_ingestion/poetry.lock
```diff
@@ -20,6 +20,38 @@ docs = ["furo", "sphinx", "zope.interface", "sphinx-notfound-page"]
 tests = ["coverage[toml] (>=5.0.2)", "hypothesis", "pympler", "pytest (>=4.3.0)", "six", "mypy", "pytest-mypy-plugins", "zope.interface", "cloudpickle"]
 tests_no_zope = ["coverage[toml] (>=5.0.2)", "hypothesis", "pympler", "pytest (>=4.3.0)", "six", "mypy", "pytest-mypy-plugins", "cloudpickle"]
 
+[[package]]
```
mzappitello, line 23: add boto3 dependency.

mzappitello: you'll want to run `poetry install` to get boto3 and its dependencies.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/batcher.py
```diff
@@ -0,0 +1,66 @@
...
+        return "%d bytes across %d files of type %s" % (self.total_size,
+                                                        len(self.filenames),
+                                                        self.config_type)
+
+    def add_file(self, filename: str, filesize: int) -> None:
+        self.filenames.append(filename)
+        self.total_size += filesize
+
+    def trigger_lambda(self) -> None:
+        raise Exception("not impled yet")
```
mzappitello, line 24: helpful little class to store stuff.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/batcher.py
```diff
@@ -0,0 +1,66 @@
...
+
+def batch_files(files: Iterable[(str, int)], threshold: int) -> list[Batch]:
```
mzappitello, line 26: hopefully a straight forward enough method.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/config_base.py
```diff
@@ -19,6 +19,24 @@ class ConfigType(Enum):
...
+        if 'com_prod_TripUpdates_enhanced' in filename:
+            return cls.BUS_TRIP_UPDATES
+        if 'com_prod_VehiclePositions_enhanced' in filename:
+            return cls.BUS_VEHICLE_POSITIONS
+        if 'net_vehicleCount' in filename:
+            return cls.VEHICLE_COUNT
+
+        raise Exception("Bad Configuration Type from filename %s" % filename)
+
+
```
mzappitello, line 39: i brought this back when it was clear we'd want it for batching.

i didn't want to create an entire `Configuration` object to figure out which type a file was.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/config_base.py
```diff
@@ -34,4 +52,4 @@ def config_type(self) -> ConfigType: ...
     def export_schema(self) -> pyarrow.schema: ...
 
     @abstractmethod
-    def record_from_entity(self, entity: dict) -> dict: ...
\ No newline at end of file
+    def record_from_entity(self, entity: dict) -> dict: ...
```
mzappitello: we should decide if we want this newline at the end of our files or not.

i think our editors are different and its causing this diff to appear.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/configuration.py
```diff
@@ -20,19 +20,24 @@ class Configuration:
     https_mbta_busloc_s3.s3.amazonaws.com_prod_VehiclePositions_enhanced.json.gz
     https_mbta_integration.mybluemix.net_vehicleCount.gz
     """
-    def __init__(self, filename: str) -> None:
+    def __init__(self,
+                 config_type: ConfigType=None,
+                 filename: str=None) -> None:
```
mzappitello, line 25: i added this so that we can create a `Configuration` either from config type or filename.

if you prefer something else, i'm happy to switch it.

py_gtfs_rt_ingestion/pyproject.toml
```diff
@@ -8,6 +8,7 @@ authors = ["MBTA CTD <developer@mbta.com>",
 [tool.poetry.dependencies]
 python = "^3.9"
 pyarrow = "^8.0.0"
+boto3 = "^1.23.3"
```
mzappitello, line 11: new dependency

py_gtfs_rt_ingestion/tests/test_configuration.py
```diff
@@ -4,29 +4,34 @@
...
 
+update_filename = \
+    '2022-01-01T00:00:02Z_https_cdn.mbta.com_realtime_TripUpdates_enhanced.json.gz'
+
+vehicle_positions_filename = \
+    '2022-01-01T00:00:03Z_https_cdn.mbta.com_realtime_VehiclePositions_enhanced.json.gz'
+
+alerts_filename = \
+    '2022-01-01T00:00:38Z_https_cdn.mbta.com_realtime_Alerts_enhanced.json.gz'
+
```
mzappitello, line 15: pulled these out so they weren't creating giant lines in most of the code.

py_gtfs_rt_ingestion/tests/test_configuration.py
```diff
@@ -4,29 +4,34 @@
...
 def test_filname_parsing():
     """
     Check that we are able to get the correct Configuration type for multiple
     filenames
     """
-    trip_updates_type = Configuration(
```
mzappitello, line 12: updated to use ConfigType rather than Configuration.

py_gtfs_rt_ingestion/batch_files.py
rymarczy: It might not matter for how we're going to be using this, but I think all standard `filter` methods are limited to returning 1000 items. 

[S3.Bucket.objects](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html?highlight=bucket#S3.Bucket.objects)
 > MaxKeys (integer) -- Sets the maximum number of keys returned in the response. By default the action returns up to 1,000 key names. The response might contain fewer keys but will never contain more.

`Paginators` are recomended for any call that could return more than 1000 results: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/paginators.html

mzappitello:
replaced by https://github.com/mbta/lamp/pull/11

PR #12, opened 2022-05-23
Add Error Handling and Logging

mzappitello:
Add a file of custom library exceptions called error.py. It will contain
a base class for our custom errors and specific ones that inherit from
it. Use them where appropriate around the library.

Add logging to the library, replacing any print statements and ensure
that it is used at the info level in the two main scripts.

### DIFF
--- a/py_gtfs_rt_ingestion/batch_files.py
+++ b/py_gtfs_rt_ingestion/batch_files.py
@@ -9,6 +9,9 @@ from collections.abc import Iterable
 
 from py_gtfs_rt_ingestion import batch_files, file_list_from_s3
 
+import logging
+logging.basicConfig(level=logging.INFO)
+
 DESCRIPTION = "Generate batches of json files that should be processesd"
 
 def parseArgs(args) -> dict:
@@ -33,11 +36,16 @@ def parseArgs(args) -> dict:
         help='filesize threshold for batch sizes')
 
     parser.add_argument(
+        '--pretty',
+        dest='pretty',
+        action='store_true',
+        help='print out a pretty summary of the batches')
+
+    parser.add_argument(
         '--output',
-        dest='output_dir',
+        dest='output',
         type=str,
-        required=True,
-        help='provide a directory to output')
+        help='filepath to dump out batches as event json list')
 
     parsed_args = parser.parse_args(args)
 
@@ -70,4 +78,19 @@ if __name__ == '__main__':
     file_list = make_file_list(args)
     batches = batch_files(file_list, args.filesize_threshold)
 
-    print(json.dumps([b.create_event() for b in batches], indent=2))
+    if args.pretty:
+        total_bytes = 0
+        total_files = 0
+        for batch in batches:
+            total_bytes += batch.total_size
+            total_files += len(batch.filenames)
+
+        total_gigs = total_bytes / 1000000000
+        logging.info("Batched %d gigs across %d files" % (total_gigs,
+                                                          total_files))
+
+    if args.output:
+        with open(args.output, 'w') as output_file:
+            json.dump([b.create_event() for b in batches],
+                      output_file,
+                      indent=2)
--- a/py_gtfs_rt_ingestion/ingest.py
+++ b/py_gtfs_rt_ingestion/ingest.py
@@ -9,7 +9,12 @@ import tempfile
 
 from pathlib import Path
 
-from py_gtfs_rt_ingestion import Configuration, convert_files, download_file_from_s3
+from py_gtfs_rt_ingestion import (Configuration,
+                                  convert_files,
+                                  download_file_from_s3)
+
+import logging
+logging.basicConfig(level=logging.INFO)
 
 DESCRIPTION = "Convert a json file into a parquet file. Used for testing."
 
@@ -64,12 +69,20 @@ def lambda_handler(event: dict, context) -> None:
              an HTTP response. The 'statusCode' field is the HTTP status code
              and the 'body' field is the body of the response.
 
-    expected event structure:
+    expected event structure is either
     {
-        file_batch: [file_name_1, file_name_2, ...],
+        files: [file_name_1, file_name_2, ...],
     }
+
+    or
+    {
+        s3_files: [file_name_1, file_name_2, ...],
+        bucket: bucket_name
+    }
+
     batch files should all be of same ConfigType
     """
+    logging.info("Processing event:\n%s" % json.dumps(event, indent=2))
     temp_dir = None
     if 'files' not in event:
         temp_dir = tempfile.TemporaryDirectory()
@@ -84,9 +97,7 @@ def lambda_handler(event: dict, context) -> None:
     pa_table = convert_files(files, config)
 
     if pa_table is not None:
-        # TODO:
-        # 1. Implement output directory as env variable for lamda function.
-        # 2. Do something with failed conversion filenames.
+        logging.info("Writing Table for %s" % config.config_type)
         pq.write_to_dataset(
             pa_table,
             root_path=os.environ['OUTPUT_DIR'],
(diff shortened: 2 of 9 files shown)


py_gtfs_rt_ingestion/batch_files.py
```diff
@@ -9,6 +9,9 @@
 
 from py_gtfs_rt_ingestion import batch_files, file_list_from_s3
 
+import logging
+logging.basicConfig(level=logging.INFO)
+
```
mzappitello, line 14: this ensures we log all info events in addition to warnings and errors.

this will log everything to stderr by default, but we can change it to stdout if we want (i think ian said thats what they do for splunk monitoring).

we can also add formatting, and change the level via input arguments if we want.

py_gtfs_rt_ingestion/batch_files.py
```diff
@@ -32,12 +35,17 @@ def parseArgs(args) -> dict:
...
+        help='print out a pretty summary of the batches')
+
     parser.add_argument(
         '--output',
-        dest='output_dir',
+        dest='output',
         type=str,
-        required=True,
-        help='provide a directory to output')
+        help='filepath to dump out batches as event json list')
```
mzappitello, line 48: i changed the output format to dump event json out to file and print things out if `--pretty` is defined.

py_gtfs_rt_ingestion/ingest.py
```diff
@@ -84,9 +97,7 @@ def lambda_handler(event: dict, context) -> None:
     pa_table = convert_files(files, config)
 
     if pa_table is not None:
-        # TODO:
-        # 1. Implement output directory as env variable for lamda function.
-        # 2. Do something with failed conversion filenames.
```
mzappitello, line 89: we're setting this in the env now.

py_gtfs_rt_ingestion/ingest.py
```diff
@@ -9,7 +9,12 @@
 
 from pathlib import Path
 
-from py_gtfs_rt_ingestion import Configuration, convert_files, download_file_from_s3
+from py_gtfs_rt_ingestion import (Configuration,
+                                  convert_files,
+                                  download_file_from_s3)
+
+import logging
+logging.basicConfig(level=logging.INFO)
```
mzappitello, line 17: add logging.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/batcher.py
```diff
@@ -1,4 +1,7 @@
+import logging
+
 from .config_base import ConfigType
+from .error import ConfigTypeFromFilenameException, NoImplException
```
mzappitello, line 4: add logging and use custom exceptions.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/batcher.py
```diff
@@ -50,17 +53,21 @@ def batch_files(files: Iterable[(str, int)], threshold: int) -> list[Batch]:
...
         try:
             config_type = ConfigType.from_filename(filename)
-        except:
-            # for now just print out the error.
-            print("Encontered Unknown File Type: %s" % filename)
+        except ConfigTypeFromFilenameException as config_exception:
+            logging.warning(config_exception)
+            continue
+        except Exception as exception:
+            loging.excption(exception)
```
mzappitello, line 64: not sure if we want to re-raise here.

basically, if we encounter a filename we cant handle, do we want to:

A) log it and move on. it won't get picked up in a batch, and it will remain in the incoming s3 bucket until sentry yells at us and we fix it

OR

B) re-raise here and nothing gets batched until we fix the issue or remove the file.

I'm leaning A

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/config_base.py
```diff
@@ -19,6 +21,9 @@ class ConfigType(Enum):
     BUS_VEHICLE_POSITIONS = auto()
     VEHICLE_COUNT = auto()
 
+    def __str__(self):
+        return self.name
+
```
mzappitello, line 26: prettier string imho.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/convert.py
```diff
@@ -45,13 +48,15 @@ def gz_to_pyarrow(filename: Path, config: Configuration):
         ret_obj = pa.table(table, schema=config.export_schema)
 
     except Exception as e:
-        print(e, flush=True)
+        logging.exception(
+            "Error occured converting %s to Parquet File." % filename.name)
```
mzappitello, line 52: `logging.exception` will also tell us about the exception that got us here.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/convert.py
```diff
@@ -69,7 +74,11 @@ def convert_files(filepaths: list[Path], config: Configuration) -> pa.Table:
...
             failed_ingestion.append(ret_obj)
 
-    pool.close()
-    pool.join()
+    logging.info(
+        "Completed converting %d files with config %s" % (len(filepaths),
+                                                          config.config_type))
+
+    if len(failed_ingestion) > 0:
+        logging.warning("Unable to process %d files" % len(failed_ingestion))
```
mzappitello, line 82: should figure out a better way to handle this other than just logging?

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/error.py
```diff
@@ -0,0 +1,32 @@
+class GTFSIngestException(Exception):
```
mzappitello, line 1: new exceptions.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/s3_utils.py
```diff
@@ -26,13 +31,19 @@ def download_file_from_s3(bucket_name: str,
...
+
+        # have to convert filenames, deliminating with . instead of /, or else the
+        # move method in the download_file method errors out with "no directory".
+        deliminated_filename = filename.replace("/", ".")
+
+        destination = os.path.join(destination_dir, deliminated_filename)
+        s3.Bucket(bucket_name).download_file(filename, destination)
+        return Path(destination)
+    except Exception as exception:
+        raise S3Exception("Failed to download %s" % filename) from exception
```
mzappitello, line 49: decided to re-raise here.

if we're not downloading things for some reason, i think we'll want our lambda jobs to fail.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/batcher.py
rymarczy: Agree that A is the way to go. Seems like we would be able to make a dashboard on splunk and watch for this type of thing.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/convert.py
rymarczy: I think this will depend on how we utilize the logging that is being sent to splunk and how frequently (if ever) this occurs.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/s3_utils.py
rymarczy: I'm not sure we would want to raise here, as it would cause the whole batch to fail because of an error on one file. 

To keep the batching approach as fault-tolerant as possible, it would probably be advantageous to have a structure where errors caused by singular files have little to no impact on the batch as a whole.

PR #16, opened 2022-05-26
Batch lambda

mzappitello:
Add lambda handler method to batch files script

* Have parse args generate an event dict that will be parsed to create a
  BatchArgs named tuple. All arguments will go through there, giving us
  some level of safety.
* Remove the option to read file names / sizes from a text file. That
  was useful when getting the script stood up, but no longer.
* Move pretty info printing into lambda handler method, along with
  printing out event json (which is no longer dumped to file)
* `batch_files`is a generator function now, rather than returning a
  list.
* remove default values from `file_list_from_s3` method. this method
  _should_ throw without these strings being provided.
  
* Add try / except to lambda handlers
In the lambda_handler methods of ingest and batch files scripts, wrap
the main functionality inside of a try except block and log any raised
exceptions. When exceptions are caught, log them with
`logging.exception` and re raise them so that lambda can pick them up.
This way failures will give a full trace back in the logs and lambda
will still error out.

### DIFF
--- a/py_gtfs_rt_ingestion/batch_files.py
+++ b/py_gtfs_rt_ingestion/batch_files.py
@@ -7,7 +7,6 @@ import sys
 from collections.abc import Iterable
 from typing import NamedTuple
 
-from py_gtfs_rt_ingestion import ArgumentException
 from py_gtfs_rt_ingestion import batch_files
 from py_gtfs_rt_ingestion import file_list_from_s3
 
@@ -53,6 +52,29 @@ def parseArgs(args) -> dict:
 
     return vars(parser.parse_args(args))
 
+def main(batch_args: BatchArgs) -> None:
+    file_list = file_list_from_s3(bucket_name=batch_args.s3_bucket,
+                                  file_prefix=batch_args.s3_prefix)
+
+    batches = batch_files(file_list, batch_args.filesize_threshold)
+
+    # TODO use boto3 to launch ingestion jobs from each batch
+
+    # log out a summary of whats been batched.
+    if batch_args.pretty:
+        total_bytes = 0
+        total_files = 0
+        for batch in batches:
+            total_bytes += batch.total_size
+            total_files += len(batch.filenames)
+
+        total_gigs = total_bytes / 1000000000
+        logging.info("Batched %d gigs across %d files" % (total_gigs,
+                                                          total_files))
+    else:
+        print(json.dumps([b.create_event() for b in batches], indent=2))
+
+
 def lambda_handler(event: dict, context) -> None:
     """
     AWS Lambda Python handled function as described in AWS Developer Guide:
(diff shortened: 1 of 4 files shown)


py_gtfs_rt_ingestion/batch_files.py
```diff
@@ -7,32 +7,34 @@
...
 logging.basicConfig(level=logging.INFO)
 
 DESCRIPTION = "Generate batches of json files that should be processed"
 
 class BatchArgs(NamedTuple):
-    input_file: str
     s3_prefix: str
+    s3_bucket: str
     filesize_threshold: int
-    output: str
```
mzappitello, line 21: no longer doing input files or output files here.

leaving in an option to print out events if invoked without `pretty` but wouldn't be surprised to see that go at some point too.

py_gtfs_rt_ingestion/batch_files.py
```diff
@@ -7,32 +7,34 @@
...
     pretty: bool
 
-def parseArgs(args) -> BatchArgs:
+def parseArgs(args) -> dict:
```
mzappitello, line 24: return an event dict. pipe it directly into `lambda_handler`

same adjustments as mentioned above.

py_gtfs_rt_ingestion/batch_files.py
```diff
@@ -48,44 +50,18 @@ def parseArgs(args) -> BatchArgs:
...
-    based on args, yield out filename, filesize tuples either from a local file
-    or from our s3 bucket.
-    """
-    if args.input_file:
-        return file_list_from_file(args.input_file)
+    return vars(parser.parse_args(args))
 
-    if args.s3_prefix:
-        return file_list_from_s3(file_prefix=args.s3_prefix)
+def main(batch_args: BatchArgs) -> None:
```
mzappitello, line 55: really recommend looking at this diff in Split view.

this main function gets the file names and sizes from s3, passes them to the batching function, logs out a summary of them or prints them to std::out (so you can capture with `./batch_files.py > events.json` if you like). eventually it should trigger the ingestion jobs.

py_gtfs_rt_ingestion/batch_files.py
```diff
@@ -95,9 +71,42 @@ def make_file_list(args: BatchArgs) -> Iterable[(str, int)]:
...
+    """
+    logging.info("Processing event:\n%s" % json.dumps(event, indent=2))
+    try:
+        batch_args = BatchArgs(**event)
+        logging.info(batch_args)
+        main(batch_args)
+    except Exception as e:
+        # log if something goes wrong and let lambda recatch the exception
+        logging.exception(e)
+        raise e
```
mzappitello, line 107: keeping this function slim. convert the args from a dict to a BatchArgs object (which will throw if the event isn't 1:1 with the members of a BatchArgs instance) and run main.

i wanted to wrap this in a try / except so that our logs will get a full traceback of any error. i re-raise for now to let lambda handle the exception and communicate back to us. we might want to change this.

py_gtfs_rt_ingestion/ingest.py
```diff
@@ -67,34 +67,7 @@ def parseArgs(args) -> dict:
...
-    logging.info("Processing event:\n%s" % json.dumps(event, indent=2))
-
-    # get files and function to read them based on the event. for local files,
-    # use gzip reading, for s3 files use pyarrow to read directly from s3
-    if 'files' not in event:
-        raise ArgumentException(
-            "Unable to find 'files' in event json")
-    files = event['files']
-
+def main(files: list[str]) -> None:
```
mzappitello, line 70: same treatment here where main does the work but is called by the `lambda_handler` method and handles the try except stuff.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/batcher.py
```diff
@@ -33,7 +33,7 @@ def create_event(self) -> dict:
             'files': self.filenames,
         }
 
-def batch_files(files: Iterable[(str, int)], threshold: int) -> list[Batch]:
+def batch_files(files: Iterable[(str, int)], threshold: int) -> Iterable[Batch]:
```
mzappitello, line 36: made this a generator function. it felt more right when i did a large test run today.

this way we can trigger jobs as we know about them rather than triggering a bunch at the end.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/s3_utils.py
```diff
@@ -3,8 +3,8 @@
 
 from collections.abc import Iterable
 
-def file_list_from_s3(bucket_name: str='mbta-gtfs-s3',
-                      file_prefix: str=None) -> Iterable[(str, int)]:
+def file_list_from_s3(bucket_name: str,
+                      file_prefix: str) -> Iterable[(str, int)]:
```
mzappitello, line 7: make these required arguments.

rymarczy:
All looks good!

PR #18, opened 2022-06-03
Add Batch.trigger_lambda functionality

mzappitello:
The batch lambda function needs to be able to trigger the ingest lambda
for each batch that it creates. The method usesboto3, and an environment
variable that contains the ingest lambdas ARN. The actual call itself is
async and run inside of the s3_utils library. There are a handful of
logging statements and exceptions to make clear whats going on / wrong.

The batch lambda script has new flags that can be used to do a dry run
(not triggering the lambda) and printing out events as json formatted
strings.

### DIFF
--- a/py_gtfs_rt_ingestion/batch_files.py
+++ b/py_gtfs_rt_ingestion/batch_files.py
@@ -2,6 +2,7 @@
 
 import argparse
 import json
+import os
 import sys
 
 from collections.abc import Iterable
@@ -19,7 +20,8 @@ class BatchArgs(NamedTuple):
     s3_prefix: str
     s3_bucket: str
     filesize_threshold: int
-    pretty: bool
+    print_events: bool = False
+    dry_run: bool = False
 
 def parseArgs(args) -> dict:
     parser = argparse.ArgumentParser(description=DESCRIPTION)
@@ -45,10 +47,16 @@ def parseArgs(args) -> dict:
         help='filesize threshold for batch sizes')
 
     parser.add_argument(
-        '--pretty',
-        dest='pretty',
+        '--dry-run',
+        dest='dry_run',
         action='store_true',
-        help='print out a pretty summary of the batches')
+        help='do not invoke ingest lambda function')
+
+    parser.add_argument(
+        '--print-events',
+        dest='print_events',
+        action='store_true',
+        help='print out each event as json to stdout')
 
     return vars(parser.parse_args(args))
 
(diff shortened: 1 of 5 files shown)


py_gtfs_rt_ingestion/batch_files.py
```diff
@@ -19,7 +20,8 @@ class BatchArgs(NamedTuple):
     s3_prefix: str
     s3_bucket: str
     filesize_threshold: int
-    pretty: bool
+    print_events: bool = False
+    dry_run: bool = False
```
mzappitello, line 24: i decided we should always log out the stuff that was behind pretty. the print events is used for local testing and the dry run stops us from triggering the lambda (which will raise an exception if the lambda arn isn't defined)

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/batcher.py
```diff
@@ -26,7 +30,14 @@ def add_file(self, filename: str, filesize: int) -> None:
...
     def trigger_lambda(self) -> None:
-        raise NoImplException("Cannot Trigger Lambda on a Batch")
+        try:
+            function_arn = os.environ["INGEST_FUNCTION_ARN"]
+            invoke_async_lambda(function_arn=function_arn,
+                                event=self.create_event())
+        except KeyError as e:
+            raise ArgumentException("Ingest Func Arn not Defined") from e
+        except Exception as e:
+            raise AWSException("Unable To Trigger Ingest Lambda") from e
```
mzappitello, line 40: trigger this using a new s3 util that will use boto3 to do this async.

catch errors here and and raise new ones from them to improve logging.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/error.py
```diff
@@ -21,3 +21,9 @@ class NoImplException(GTFSIngestException):
     General Error for things we haven't done yet.
     """
     pass
+
+class AWSException(GTFSIngestException):
```
mzappitello, line 25: pretty straight forward.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/s3_utils.py
```diff
@@ -29,3 +29,10 @@ def file_list_from_s3(bucket_name: str,
             continue
         for obj in page['Contents']:
             yield (obj['Key'], obj['Size'])
+
+def invoke_async_lambda(function_arn: str, event: dict) -> None:
+    lambda_client = boto3.client('lambda')
+    logging.info("Invoking Lambda: %s" % function_arn)
+    lambda_client.invoke(FunctionName=function_arn,
+                         InvocationType='Event',
+                         Payload=json.dumps(event))
```
mzappitello, line 38: not sure how to stub this, but it follows the patterns i saw in documentation and examples.

the `InvocationType='Event'` makes this async.

py_gtfs_rt_ingestion/tests/test_batcher.py
```diff
@@ -39,7 +39,7 @@ def test_batch_class(capfd):
     assert out == f"Batch of {sum(files.values())} bytes in {len(files)} {config_type} files\n" 
 
     # Calling `trigger_lambda` method raises exception
-    with pytest.raises(NoImplException):
+    with pytest.raises(ArgumentException):
```
mzappitello, line 42: again, not sure how to stub this.

probably one of those things that makes the most sense to catch on end to end testing.

py_gtfs_rt_ingestion/batch_files.py
```diff
@@ -45,35 +47,44 @@ def parseArgs(args) -> dict:
...
 def main(batch_args: BatchArgs) -> None:
     file_list = file_list_from_s3(bucket_name=batch_args.s3_bucket,
                                   file_prefix=batch_args.s3_prefix)
 
-    batches = batch_files(file_list, batch_args.filesize_threshold)
 
-    # TODO use boto3 to launch ingestion jobs from each batch
+    total_bytes = 0
+    total_files = 0
+    events = []
```
mzappitello, line 70: always keep track of this, regardless of flags.

py_gtfs_rt_ingestion/batch_files.py
```diff
@@ -45,35 +47,44 @@ def parseArgs(args) -> dict:
...
-        total_bytes = 0
-        total_files = 0
-        for batch in batches:
-            total_bytes += batch.total_size
-            total_files += len(batch.filenames)
+        if not batch_args.dry_run:
+            try:
+                batch.trigger_lambda()
+            except Exception as e:
+                logging.exception("Unable to trigger ingest lambda.\n%s" % e)
```
mzappitello, line 80: catch an exception here so that other batches have a chance of getting kicked off.

we'll probably want to figure out how to get splunk to watch for these errors specifically, since that would prevent ingestion from ever happening.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/s3_utils.py
rymarczy: Do we want to do anything with the return object from `.invoke`? 

It offers a http `StatusCode` and `FunctionError` string that we might want to do something with.

Not sure if an invocation failure would be caught / reported anywhere else.

mzappitello: if we're waiting for a response, then we can run into cases where the batch lambda is timing out because its waiting for responses from the ingest lambda. it is a good point though that this is an error who's details we'll want to capture and log. i made a ticket to do some digging into that.

https://app.asana.com/0/1189492770004753/1202389697849560/f

PR #21, opened 2022-06-08
Dev test setup script

rymarczy:
Initial attempt at creating a python script to help with S3 dev bucket setup. 

Script provides prompts to clear out all lamp dev buckets and copy random selection of from from gtfs bucket to incoming bucket.


py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,132 @@
...
+    """
+    parse input args from the command line and generate an event dict in the
+    format the lambda handler is expecting
+    """
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '--dev-prefix',
+        dest='dev_prefix',
+        type=str,
+        default='lamp/',
```
mzappitello, line 37: i think you can drop the trailing slash here.

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,132 @@
...
+        DEV_EXPORT_BUCKET,
+        DEV_ARCHIVE_BUCKET,
+        DEV_ERROR_BUCKET,
+    )
+    
+    for bucket in bucket_list:
+        uri_root = f"s3://{os.path.join(bucket, args.dev_prefix)}"
+        action = None
+        print(f"Checking for objects to delete in {uri_root}...")
+        files_to_delete = tuple(uri for (uri,size) in file_list_from_s3(bucket, args.dev_prefix) if size > 0)
```
mzappitello, line 68: why the if condition?

mzappitello: and does this capture folders as well?

mzappitello:
a few questions and suggestions.

py_gtfs_rt_ingestion/dev_test_setup.py
rymarczy: It works better for the script to leave it on, but I moved this into a constant variable instead of a command line argument, since it should never really change.

py_gtfs_rt_ingestion/dev_test_setup.py
rymarczy: New version does capture folders (except the root lamp folder), so this check is not used anymore.

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,132 @@
...
+        if len(files_to_delete) == 0:
+            print(f"No objects found... skipping bucket.")
+        else: 
+            while action is None or action not in ('n','no'):
```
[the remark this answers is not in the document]
rymarczy: Incorporated, `-f` flag for deleting all bucket objects without asking.

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,132 @@
...
+                    s3_client.delete_objects(
+                        Bucket=bucket,
+                        Delete=delete_objs,
+                    )
+                    break
+                print(f"{len(files_to_delete)} objects found, delete?\nyes(y) / no(n) / list(ls)")
+                action = input()
+
+def copy_gfts_to_ingest(args:SetupArgs):
+    count_objs_to_pull = args.objs_to_copy * 10
```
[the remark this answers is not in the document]
rymarczy: This has changed to 1,000 objects or 10x the amount being pulled from the folder (which ever is larger). This is so a random sample can be selected from.

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,298 @@
...
+        help="number of objects to move into dev environment",
+    )
+
+    parser.add_argument(
+        "-f",
+        dest="force_delete",
+        default=False,
+        action="store_true",
+        help="delete objects from dev buckets without asking",
+    )
```
rymarczy, line 66: Add argument to delete all objects from dev buckets without asking, should this also apply to the object copy?

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,298 @@
...
+    }
+    try:
+        s3_client.delete_objects(
+            Bucket=bucket,
+            Delete=delete_objs,
+        )
+    except Exception as e:
+        logging.exception(e)
+        return 0
+    return len(del_list)
```
rymarczy, line 89: Function to delete batches of objects. The boto3 library limits the size of these to 1,000 max.

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,298 @@
...
+    Spin up Multiprocessing jobs to delete batches of objects
+    """
+    chunk = 750
+    with Pool(os.cpu_count()) as pool:
+        l_t = [
+            (files[i : i + chunk], bucket) for i in range(0, len(files), chunk)
+        ]
+        results = pool.starmap_async(del_objs, l_t)
+        del_count = sum(r for r in results.get())
+    return del_count
```
rymarczy, line 103: Function to create workers for object deletion, size of chunks depends on `chunk` value. Boto3 limits this to 1,000 max, 750 seems to work pretty well without throwing too many errors

Number of objects deleted is returned by each worker, sum value of returned values determines if all objects were deleted.

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,298 @@
...
+                # Proceed with object deletion
+                elif action in ("y", "yes") or args.force_delete:
+                    delete_count = make_del_jobs(files_to_delete, bucket)
+                    # If not all objects deleted, retry
+                    if delete_count < len(files_to_delete):
+                        print(
+                            f"Only {delete_count:,} of {len(files_to_delete):,}"
+                            " deleted... will retry."
+                        )
+                        files_to_delete = get_del_obj_list(bucket, uri_root)
```
rymarczy, line 153: Semi-automatic retry logic if not all objects are deleted.

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,298 @@
...
+
+
+def copy_obj(prefix: str, num_to_copy: int) -> int:
+    """
+    Copy random objects from prefix to dev Ingest bucket
+    """
+    count_objs_to_pull = max(num_to_copy * 10, 1_000)
+    uri_copy_set = set()
+    src_uri_root = f"s3://{os.path.join(GTFS_BUCKET, prefix)}"
+    skip_uri = "https_mbta_busloc_s3.s3.amazonaws.com_prod_TripUpdates_enhanced"
```
rymarczy, line 173: `GTFS_BUCKET` still has permission issues with busloc TripUpdates files, so they are completely skipped.

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,298 @@
...
+    print(
+        f"Pulling list of {count_objs_to_pull:,} objects from "
+        f"{src_uri_root} for random sample..."
+    )
+    try:
+        for (uri, size) in file_list_from_s3(GTFS_BUCKET, prefix):
+            # Skip busloc TripUpdates because of S3 permission issues
+            if size > 0 and skip_uri not in uri:
+                uri_copy_set.add(uri)
```
rymarczy, line 182: Avoids copying folders and skips busloc TripUpdate files.

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,298 @@
...
+    success_count = 0
+    print(
+        f"Starting copy of {num_to_copy} random objects from {src_uri_root} ..."
+    )
+    for uri in random.sample(tuple(uri_copy_set), num_to_copy):
+        key = str(uri).replace(f"s3://{GTFS_BUCKET}/", "")
+        copy_source = {
+            "Bucket": GTFS_BUCKET,
+            "Key": key,
+        }
```
rymarczy, line 200: Copy random sample of objects pulled. `uri_copy_set` will be a minimum of 1,000 objects and up to 10x the number to be copied.

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,298 @@
...
+    Enumerate folders in specified bucket prefix combination
+    """
+    response = client.list_objects_v2(
+        Bucket=bucket, Delimiter="/", Prefix=prefix, MaxKeys=45
+    )
+    if "CommonPrefixes" not in response:
+        yield prefix
+    else:
+        for new_prefix in response["CommonPrefixes"]:
+            yield from drill_s3_folders(client, bucket, new_prefix["Prefix"])
```
rymarczy, line 228: generator to enumerate folders in root copy folder. This allows equal number of files to be pulled from each nested folder.

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,298 @@
...
+    # Combine list of folders and number of files to pull from each folder
+    folder_pull_cnt = tuple(
+        zip(
+            folders,
+            [
+                per_folder + 1 if x < remain else per_folder
+                for x in range(len(folders))
+            ],
+        )
+    )
```
rymarczy, line 260: Determines the number of objects to be pulled from each enumerated folder.

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,298 @@
...
+            with Pool(pool_size) as pool:
+                results = pool.starmap_async(copy_obj, folder_pull_cnt)
+                success_count = 0
+                for result in results.get():
+                    success_count += result
+            print(
+                f"{success_count:,} objects copied, "
+                f"{args.objs_to_copy - success_count} errors."
+            )
+            break
```
rymarczy, line 279: Create object copy workers, will create 2x he number of cpu's found on the machine. These are low computation but high latency operations, so more workers == more better, I think...

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,298 @@
...
+        ]
+    }
+    try:
+        s3_client.delete_objects(
+            Bucket=bucket,
+            Delete=delete_objs,
+        )
+    except Exception as e:
+        logging.exception(e)
+        return 0
```
mzappitello, line 88: probably not a big deal given the scope of this script, but does `delete_objects` ever delete some of them and then raise an exception?

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,298 @@
...
+    return len(del_list)
+
+
+def make_del_jobs(files: list[str], bucket: str) -> int:
+    """
+    Spin up Multiprocessing jobs to delete batches of objects
+    """
+    chunk = 750
+    with Pool(os.cpu_count()) as pool:
+        l_t = [
```
mzappitello, line 98: what does `l_t` stand for?

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,298 @@
...
+        DEV_ARCHIVE_BUCKET,
+        DEV_ERROR_BUCKET,
+    )
+
+    for bucket in bucket_list:
+        action = None
+        uri_root = f"s3://{os.path.join(bucket, LAMP_PREFIX)}"
+        files_to_delete = get_del_obj_list(bucket, uri_root)
+        if len(files_to_delete) == 0:
+            print("No objects found... skipping bucket.")
```
mzappitello, line 135: should be `logging` not `print`

py_gtfs_rt_ingestion/dev_test_setup.py
mzappitello, line 153: was this a case that you were regularly seeing?

also, `logging` again rather than print.

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,298 @@
...
+from py_gtfs_rt_ingestion import file_list_from_s3
+
+GTFS_BUCKET = "mbta-gtfs-s3"
+DEV_INGEST_BUCKET = "mbta-ctd-dataplatform-dev-incoming"
+DEV_EXPORT_BUCKET = "mbta-ctd-dataplatform-dev-springboard"
+DEV_ARCHIVE_BUCKET = "mbta-ctd-dataplatform-dev-archive"
+DEV_ERROR_BUCKET = "mbta-ctd-dataplatform-dev-error"
+LAMP_PREFIX = "lamp/"
+
+logging.basicConfig(level=logging.WARNING)
```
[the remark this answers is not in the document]
mzappitello: and since this script is user interactive, you might consider setting the level based on a CLI Arg.

https://stackoverflow.com/a/60679407

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,298 @@
...
+                    delete_count = make_del_jobs(files_to_delete, bucket)
+                    # If not all objects deleted, retry
+                    if delete_count < len(files_to_delete):
+                        print(
+                            f"Only {delete_count:,} of {len(files_to_delete):,}"
+                            " deleted... will retry."
+                        )
+                        files_to_delete = get_del_obj_list(bucket, uri_root)
+                    else:
+                        print(f"All {len(files_to_delete)} deleted")
```
mzappitello, line 155: `logging` again.

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,298 @@
...
+                        break
+
+                print(
+                    f"{len(files_to_delete):,} objects found, delete? "
+                    "\n"
+                    "yes(y) / no(n) / list(ls)"
+                )
```
mzappitello, line 162: i think this stays as print since its a prompt for the user.

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,298 @@
...
+    Copy random objects from prefix to dev Ingest bucket
+    """
+    count_objs_to_pull = max(num_to_copy * 10, 1_000)
+    uri_copy_set = set()
+    src_uri_root = f"s3://{os.path.join(GTFS_BUCKET, prefix)}"
+    skip_uri = "https_mbta_busloc_s3.s3.amazonaws.com_prod_TripUpdates_enhanced"
+    print(
+        f"Pulling list of {count_objs_to_pull:,} objects from "
+        f"{src_uri_root} for random sample..."
+    )
```
mzappitello, line 177: `logging` again.

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,298 @@
...
+    try:
+        for (uri, size) in file_list_from_s3(GTFS_BUCKET, prefix):
+            # Skip busloc TripUpdates because of S3 permission issues
```
mzappitello, line 180: what was the deal with this again?

they have different permissions set even though they are in the same bucket?

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,298 @@
...
+    except Exception as e:
+        logging.error("Unable to pull file list from %s", src_uri_root)
+        logging.exception(e)
+        return 0
+
+    s3_client = boto3.client("s3")
+    success_count = 0
+    print(
+        f"Starting copy of {num_to_copy} random objects from {src_uri_root} ..."
+    )
```
mzappitello, line 194: `logging`

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,298 @@
...
+
+def copy_gfts_to_ingest(args: SetupArgs) -> None:
+    """
+    Enumerate folders in GTFS_BUCKET prefix and create workers to copy
+    objects to development Ingest bucket
+    """
+    src_uri_root = f"s3://{os.path.join(GTFS_BUCKET, args.src_prefix)}"
+    dest_urc_root = f"s3://{os.path.join(DEV_INGEST_BUCKET, LAMP_PREFIX)}"
+
+    print(f"Enumerating folders in ({src_uri_root})...")
```
mzappitello, line 239: `logging`

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,298 @@
...
+    s3_client = boto3.client("s3")
```
mzappitello, line 240: does it make sense to create one client when the script is kicked off and use that everywhere?

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,298 @@
...
+    folders = list(drill_s3_folders(s3_client, GTFS_BUCKET, args.src_prefix))
+    # Get number of files to pull per folder
+    per_folder = args.objs_to_copy // len(folders)
```
mzappitello, line 243: it feels like `objs_to_copy` has to be quite large for this to not always be 1 given how many subdirs exist in the gtfs bucket, right?

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,298 @@
...
+    # Number of folders that will have extra file
+    remain = args.objs_to_copy % len(folders)
+    print(
+        f"{len(folders):,} folders found, "
+        f"will copy ~{per_folder} objects from each folder."
+    )
```
mzappitello, line 249: `logging`

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,298 @@
...
+                pool_size *= 2
+            with Pool(pool_size) as pool:
+                results = pool.starmap_async(copy_obj, folder_pull_cnt)
+                success_count = 0
+                for result in results.get():
+                    success_count += result
+            print(
+                f"{success_count:,} objects copied, "
+                f"{args.objs_to_copy - success_count} errors."
+            )
```
mzappitello, line 278: `logging`

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,298 @@
...
+            break
+
+        print(
+            f"Copy random selection of {args.objs_to_copy} objects "
+            f"from {src_uri_root} to {dest_urc_root} ?\n"
+            "yes(y) / no(n)"
+        )
```
mzappitello, line 285: `logging`

mzappitello:
this mostly looks good to me.

i think there is some decision making about `logging` vs `print` that should be considered, especially if you want to add a log level to the arg parse to cut back on how much is actually output.

i also think its worth considering making a single s3 client that gets used everywhere.

py_gtfs_rt_ingestion/dev_test_setup.py
rymarczy: If you are doing the whole bucket, yes. The script works best if you use it with a specific year prefix. Each year maxes out at 365 subdirs, one for each day.

py_gtfs_rt_ingestion/dev_test_setup.py
rymarczy: Incorporated

py_gtfs_rt_ingestion/dev_test_setup.py
rymarczy: Incorporated

py_gtfs_rt_ingestion/dev_test_setup.py
rymarczy: Incorporated

py_gtfs_rt_ingestion/dev_test_setup.py
rymarczy: Incorporated

py_gtfs_rt_ingestion/dev_test_setup.py
rymarczy: Incorporated

py_gtfs_rt_ingestion/dev_test_setup.py
rymarczy: updated to `chunks_list`

py_gtfs_rt_ingestion/dev_test_setup.py
rymarczy: Yup, it's a known issue https://app.asana.com/0/1200152089785426/1200192016253762/f

py_gtfs_rt_ingestion/dev_test_setup.py
rymarczy: This is the only client that is created in the main thread. All of the other client's are created in multiprocessing Pools, which I think is the most performant approach. We could try to pass one client around between processes, but I think that adds an extra level of complexity that is not ideal for this type of script.

py_gtfs_rt_ingestion/dev_test_setup.py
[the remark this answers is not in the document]
rymarczy: Incorporated

py_gtfs_rt_ingestion/dev_test_setup.py
rymarczy: It is intermittent depending on the number of objects that are being deleted and also if boto3 just randomly craps out because of a connection issue. For less than 5,000 objects it almost never happens but for 20,000+ it seems to happen about 20% of the time.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/s3_utils.py
```diff
@@ -34,7 +34,7 @@ def file_list_from_s3(
             continue
         for obj in page["Contents"]:
             uri = os.path.join("s3://", bucket_name, obj["Key"])
-            logging.info(uri)
+            logging.debug(uri)
```
rymarczy, line 37: Update to debug logging to avoid lots of INFO printing

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,330 @@
...
+    """
+    Enumerate folders in specified bucket prefix combination
+    """
+    try:
+        response = client.list_objects_v2(
+            Bucket=bucket, Delimiter="/", Prefix=prefix, MaxKeys=45
+        )
+    except Exception as e:
+        logging.exception(e)
+        yield None
```
rymarczy, line 239: Add try except block around this boto3 call to avoid it nuking the script

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -0,0 +1,330 @@
...
+        "CRITICAL",
+        "ERROR",
+        "WARNING",
+        "INFO",
+        "DEBUG",
+        "NOTSET",
+    )
+    if args.log_level not in log_levels:
+        raise KeyError(f"{args.log_level} not a valid log_level")
+    logging.basicConfig(level=logging.getLevelName(args.log_level))
```
rymarczy, line 323: Add log level setting from CL argument

py_gtfs_rt_ingestion/dev_test_setup.py
mzappitello: i would think so.

having a "non user interactive" mode feels like something we would eventually want.

py_gtfs_rt_ingestion/dev_test_setup.py
mzappitello: oh thats right. i remember running into this problem before.

PR #28, opened 2022-06-17
Bug Fix  Move Lambda Types to gtfsrt ingest Module

mzappitello:
The Lambda Types file was added in to provide type hints to the
parameters to lambda handler functions in both of our batch files and
ingest scripts. It was put at the top level next to both of these
scripts, but that causes problems when zipping up the project for
deployments. To avoid this, move it to the module itself so that it will
always be bundled.


py_gtfs_rt_ingestion/batch_files.py
```diff
@@ -7,12 +7,12 @@
 import sys
 
 from typing import NamedTuple
-from lambda_types import LambdaDict, LambdaContext
 
 from py_gtfs_rt_ingestion import ArgumentException
 from py_gtfs_rt_ingestion import DEFAULT_S3_PREFIX
 from py_gtfs_rt_ingestion import batch_files
 from py_gtfs_rt_ingestion import file_list_from_s3
+from py_gtfs_rt_ingestion.lambda_types import LambdaDict, LambdaContext
```
rymarczy, line 15: This import pattern is different from the other items being brought in from the library, but I'm fine with it if you are.

rymarczy:
LGTM, other than the varied import pattern.

mzappitello:
changed the import back. i'm going to merge now.

PR #30, opened 2022-06-17
Fix: Batch event parsing needs default filesize threshold

rymarczy:
Asana Task: https://app.asana.com/0/1202413225616890/1202464608350249

- adds default batch `filesize_threshold`
- add conditional for lambda event payload size in `batcher.py`

### DIFF
--- a/py_gtfs_rt_ingestion/batch_files.py
+++ b/py_gtfs_rt_ingestion/batch_files.py
@@ -24,7 +24,7 @@ DESCRIPTION = "Generate batches of json files that should be processed"
 class BatchArgs(NamedTuple):
     """wrapper for arguments to main method"""
 
-    filesize_threshold: int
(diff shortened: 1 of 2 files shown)


py_gtfs_rt_ingestion/batch_files.py
```diff
@@ -144,7 +144,10 @@ def lambda_handler(event: LambdaDict, context: LambdaContext) -> None:
     logging.info("Context:\n%s", context)
 
     try:
-        batch_args = BatchArgs(**event)
+        if "filesize_threshold" in event:
+            batch_args = BatchArgs(**event)
+        else:
+            batch_args = BatchArgs()
```
mzappitello, line 150: i think this can just stay as it was now that each member of `BatchArgs` has a default value.

PR #32, opened 2022-06-24
Add Handling of Schedule Zip Data

mzappitello:
* Add new ConfigType for zipped up schedule data
* Create convert_gtfs dot py that will take a gtfs zip file, pull out
  date time information, and generate pyarrow tables for each of the
  different csv's inside.
* add new s3 util that will allow us to read contents of dot zip files
  from s3 using python's ZipFile module without downloading them.
* update ingest to have a track for handling dot zip files and a track
  for handling dot json dot gz files.
* rename convert dot py to convert_gtfs_rt dot py
* Add tests and test file for schedule gtfs work

Asana Task: https://app.asana.com/0/1202413225616901/1202183746758634/f


py_gtfs_rt_ingestion/ingest.py
```diff
@@ -17,11 +17,13 @@
 from py_gtfs_rt_ingestion import ArgumentException
 from py_gtfs_rt_ingestion import ConfigTypeFromFilenameException
 from py_gtfs_rt_ingestion import Configuration
+from py_gtfs_rt_ingestion import ConfigType
```
mzappitello, line 20: need to use config type in order to choose between gtfs_rt and gtfs data.

py_gtfs_rt_ingestion/ingest.py
```diff
@@ -17,11 +17,13 @@
...
 from py_gtfs_rt_ingestion import DEFAULT_S3_PREFIX
 from py_gtfs_rt_ingestion import LambdaContext
 from py_gtfs_rt_ingestion import LambdaDict
 from py_gtfs_rt_ingestion import NoImplException
 from py_gtfs_rt_ingestion import gz_to_pyarrow
+from py_gtfs_rt_ingestion import zip_to_pyarrow
```
mzappitello, line 26: method to convert gtfs data to pyarrow table.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/config_base.py
```diff
@@ -27,6 +27,7 @@ class ConfigType(Enum):
     BUS_TRIP_UPDATES = auto()
     BUS_VEHICLE_POSITIONS = auto()
     VEHICLE_COUNT = auto()
+    SCHEDULE = auto()
```
mzappitello, line 30: new `ConfigType` for schedule data.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/convert_gtfs.py
```diff
@@ -0,0 +1,53 @@
+import zipfile
```
mzappitello, line 1: should be a pretty straight forward file.

get the `ZipFile` either locally or from s3. parse out the start date timestamp, generate a table from each text file in the zip, adding in start time columns to each one.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/s3_utils.py
```diff
@@ -12,6 +13,20 @@ def get_s3_client() -> boto3.client:
     return boto3.client("s3")
 
 
+def get_zip_buffer(filename: str) -> BytesIO:
```
mzappitello, line 16: read without downloading.

py_gtfs_rt_ingestion/tests/test_convert.py
```diff
@@ -313,3 +314,315 @@ def test_bus_vehicle_positions_file_conversion() -> None:
             assert upper == str(np_df[col].max())
         if lower != "nan":
             assert lower == str(np_df[col].min())
+
+
+def test_schedule_conversion() -> None:
```
mzappitello, line 319: i added a test zip file and ensured it was parsing things out the way I expected. field names came from mbta's github and lenghts were found via inspection.

py_gtfs_rt_ingestion/ingest.py
```diff
@@ -106,43 +108,61 @@ def main(files: list[str]) -> None:
...
+                config.config_type,
+            )
+
+            logging.info("Writing Table to %s", export_bucket)
+            pq.write_to_dataset(
+                table=pa_table,
+                root_path=os.path.join(export_bucket, str(config.config_type)),
+                filesystem=s3_filesystem,
+                partition_cols=["year", "month", "day", "hour"],
+            )
```
mzappitello, line 165: two track conversion here now.  one track for gtfs and one for gtfs_rt.

if we ever have more sources, we probably want to abstract this a bit and move the logic into the library.

tbh, i kind of feel like we should do that now.

rymarczy:
I think this all looks good for the approach we discussed last week. But I have been thinking about edge cases where this could break. 

The main one would be if two schedule zip files happen to be issued on the same day. We are basically using they `feed_start_date` as a primary key for these data tables. It might be a case of over engineering, but I'm thinking we might want to generate and GUID for each zip file we process, add that as a column to each table we're writing and then just join everything together at a later date how we see fit.

paulswartz:
> It might be a case of over engineering, but I'm thinking we might want to generate and GUID for each zip file we process, add that as a column to each table we're writing and then just join everything together at a later date how we see fit.

This has happened before. Instead of a GUID, another option (which is also sortable) would be the `Last-Modified` date of the GTFS file.

PR #36, opened 2022-07-05
Filter out s3 files based on last modified date

mzappitello:
The Batching lambda is scheduled to run at the start of every hour.
However, due to start up times, the actual run time is a few minutes
after the start of the hour. This causes most batches to have files that
were created between 3 minutes past the previous hour up until 3 minutes
past the current hour. When the parquet files are generated from these
ingested files, we partition them, leaving us with two files. One is a
large file for all data created during the previous hour while the other
is quite small, containing only data created during the first three
minutes of the current hour.

Update `file_list_from_s3` utility to filter out files modified after
passed in datetime. This feature is used in batch files dot py to
prevent the files that have been uploaded since the start of the current
hour from being batched, which should help us avoid the splits happening
on partition.

Tests were updated to catch this behavior.

pyproject dot toml was also updated to allow log statements to come
through on failure, which is useful for debugging code both in the
module and in the tests.

Asana Task: https://app.asana.com/0/1189492770004753/1202506607268445/f

### DIFF
--- a/py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/s3_utils.py
+++ b/py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/s3_utils.py
@@ -3,6 +3,7 @@ import logging
 import os
 
 from collections.abc import Iterable
+from datetime import datetime
 from typing import IO
 from io import BytesIO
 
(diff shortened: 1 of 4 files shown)


py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/s3_utils.py
```diff
@@ -34,28 +35,35 @@ def get_zip_buffer(filename: str) -> tuple[IO[bytes], int]:
...
 
     :param bucket_name: the name of the bucket to look inside of
     :param file_prefix: prefix for files to generate
+    :param until: filter out any files modified after this datetime
 
     :yield filename, filesize tuples from inside of the bucket
     """
     logging.info(
-        "Getting files with prefix %s from %s", file_prefix, bucket_name
+        'Getting files with prefix "%s" from %s%s',
```
rymarczy, line 50:
```suggestion
        'Getting files with prefix "%s" from %s %s',
```

Missing a space?

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/s3_utils.py
```diff
@@ -34,28 +35,35 @@ def get_zip_buffer(filename: str) -> tuple[IO[bytes], int]:
...
     pages = paginator.paginate(Bucket=bucket_name, Prefix=file_prefix)
     for page in pages:
         if page["KeyCount"] == 0:
             continue
+
         for obj in page["Contents"]:
             uri = os.path.join("s3://", bucket_name, obj["Key"])
-            logging.debug(uri)
+            logging.debug("URI %s last modified %s", uri, obj["LastModified"])
+            if until is not None and obj["LastModified"] > until:
```
rymarczy, line 65:
```suggestion
            if until is not None and obj["LastModified"] >= until:
```

Skip over anything from exactly on the hour?

rymarczy:
Couple of comments, everything else LGTM.

mzappitello:
replaced by
https://github.com/mbta/lamp/pull/37

PR #37, opened 2022-07-06
Filter Out GTFS RT Ingestion based on timestamp

mzappitello:
The Batching lambda is scheduled to run at the start of every hour.
However, due to start up times, the actual run time is a few minutes
after the start of the hour. This causes most batches to have files that
were created between 3 minutes past the previous hour up until 3 minutes
past the current hour. When the parquet files are generated from these
ingested files, we partition them, leaving us with two files. One is a
large file for all data created during the previous hour while the other
is quite small, containing only data created during the first three
minutes of the current hour. Doing the filtering during batch would be
cleaner, but would require reading the gtfs_rt files themselves to the
their internal timestamps that we derive year, month, day, and hour
information from. Instead, filter out based on time during ingest.
Files that are too recent to neither archive nor error, so they will
remain in the incoming s3 bucket for the next hours pass.

* add `start_of_hour` member to `GtfsRtConverter` class that will be
  used to filter files that are too recent.
* in `gz_to_pyarrow` return None if the files internal timestamp is
  passed the start of the hour.
* add tests to check that `start_of_hour` has no minute, second, or
  microsecond information and that the filtering works as expected. a
  new vehicle positions file was added for this purpose.
* update pytest in pyproject dot toml to allow log statements to come
  through on failed tests, aiding debugging.

BUG FIX:
While working on this ticket I noticed that `GtfsRtConverter` was adding
files to the archive list when catching exceptions. That behavior was
patched and tests were updated to ensure everything was working as
expected.

Asana Task: https://app.asana.com/0/1189492770004753/1202506607268445/f

This PR replaces https://github.com/mbta/lamp/pull/36, which had a flawed assumption.


py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/convert_gtfs_rt.py
```diff
@@ -55,6 +55,11 @@ def __init__(self, config_type: ConfigType) -> None:
         else:
             raise NoImplException(f"No Specialization for {config_type}")
 
+        # get the start of the current hour. any files written after this will
+        # not be ingested until the next go around.
+        now = datetime.now(tz=timezone.utc)
+        self.start_of_hour = now.replace(minute=0, second=0, microsecond=0)
+
```
mzappitello, line 62: at a start of hour member that we'll use to filter out files that are too recent.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/convert_gtfs_rt.py
```diff
@@ -130,7 +135,15 @@ def gz_to_pyarrow(self, filename: str) -> Optional[str]:
 
             # parse timestamp info out of the header
             feed_timestamp = json_data["header"]["timestamp"]
-            timestamp = datetime.utcfromtimestamp(feed_timestamp)
+            timestamp = datetime.fromtimestamp(feed_timestamp, timezone.utc)
```
mzappitello, line 138: use from timestamp with a utc timezone rather than utcfromtimestamp.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/convert_gtfs_rt.py
```diff
@@ -154,7 +167,7 @@ def gz_to_pyarrow(self, filename: str) -> Optional[str]:
         except Exception as exception:
             logging.error("Error converting %s", filename)
             logging.exception(exception)
-            self.archive_files.append(filename)
+            self.error_files.append(filename)
```
mzappitello, line 170: caught this bug.

py_gtfs_rt_ingestion/pyproject.toml
```diff
@@ -44,6 +44,11 @@ good-names = ["e"]
 max-line-length = 80
 min-similarity-lines = 10
 
+[tool.pytest]
+log_cli = true
+log_cli_level = "DEBUG"
+verbose = true
+
```
mzappitello, line 51: helpful for debugging tests.

py_gtfs_rt_ingestion/tests/test_gtfs_rt_converter.py
```diff
@@ -9,20 +10,45 @@
...
+
+    assert len(tables) == 1
+
+    # archive files should be empty, error files should have the bad file in it
+    assert len(converter.archive_files) == 0
+    assert converter.error_files == ["s3://badfile"]
+
+    # get the table and ensure its empty
+    _, table = tables[0]
+    assert table.num_rows == 0
```
mzappitello, line 51: split this tests into two tests. one for local one for s3.

now check that error files and archive files lists are correct and that convert produces empty tables with no rows.

py_gtfs_rt_ingestion/tests/test_gtfs_rt_converter.py
```diff
@@ -402,3 +438,53 @@ def test_bus_trip_updates_file_conversion() -> None:
             assert upper == str(np_df[col].max())
         if lower != "nan":
             assert lower == str(np_df[col].min())
+
+
+def test_start_of_hour() -> None:
```
mzappitello, line 443: new test to check that time filtering works as expected.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/convert_gtfs_rt.py
```diff
@@ -130,7 +135,14 @@ def gz_to_pyarrow(self, filename: str) -> Optional[str]:
 
             # parse timestamp info out of the header
             feed_timestamp = json_data["header"]["timestamp"]
-            timestamp = datetime.utcfromtimestamp(feed_timestamp)
+            timestamp = datetime.fromtimestamp(feed_timestamp, timezone.utc)
+
+            # skip files that have been generated after the start of the current
+            # hour. don't add them to archive or error so that they are picked
+            # up next go around.
+            if timestamp > self.start_of_hour:
```
rymarczy, line 143:
```suggestion
            if timestamp >= self.start_of_hour:
```

Don't want to process files with timestamps equal to the hour.

rymarczy:
Just one comment. LGTM otherwise.

PR #38, opened 2022-07-14
add logging for when batcher event size is over aws limit

rymarczy:
Currently we don't know when a new batch is being triggered by the event size being over the AWS 256KB limit. This change adds a log message when a batch is created by the event size being too large.

Asana Task: https://app.asana.com/0/1202535370823256/1202604738421227


mzappitello:
looks good to me.

PR #42, opened 2022-07-26
Use Boto3 Sessions for Threading S3 File Moves

mzappitello:
Using a boto3 client to execute s3 file moves runs into backend race
conditions that were throwing exceptions. To avoid this, the s3 file
moving utility now uses a new Session for each move and gets a resource
from that session.

### DIFF
--- a/py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/s3_utils.py
+++ b/py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/s3_utils.py
@@ -80,9 +80,17 @@ def invoke_async_lambda(function_arn: str, event: dict) -> None:
 
 
 def _move_s3_object(filename: str, destination: str) -> None:
-    # filename is expected as 's3://my_bucket/the/path/to/the/file.json
+    """
+    move a single s3 file to the destination bucket. break the incoming s3 path
+    intco parts that are used for copying. each execution will spin up a session
+    and get a resource from that session to avoid threading issues.
+
+    :param filename - expected as 's3://my_bucket/the/path/to/the/file.json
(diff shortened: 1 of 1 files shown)


py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/s3_utils.py
```diff
@@ -80,9 +80,17 @@ def invoke_async_lambda(function_arn: str, event: dict) -> None:
...
+    intco parts that are used for copying. each execution will spin up a session
+    and get a resource from that session to avoid threading issues.
+
+    :param filename - expected as 's3://my_bucket/the/path/to/the/file.json
+    :param destination bucket name
+    """
     try:
-        s3_client = get_s3_client()
+        session = boto3.session.Session()
+        s3_resource = session.resource('s3')
```
mzappitello, line 93: this seems to be the way to do threading.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/s3_utils.py
```diff
@@ -98,16 +106,21 @@ def _move_s3_object(filename: str, destination: str) -> None:
         logging.debug("copy key 1: %s", copy_key)
 
         # json args for cop
-        copy_source = {
-            "Bucket": source,
-            "Key": copy_key,
-        }
+        destination_bucket = s3_resource.Bucket(destination)
```
mzappitello, line 109: use resource api rather than client one. the concept of

`move = copy + delete` remains.

rymarczy:
LGTM

PR #43, opened 2022-08-02
Create Performance Manager Infra

mzappitello:
* Create `performance_manager` directory that will contain the code that
  runs on our ECS tasks and describes its dependencies. It will
  eventually house tests that can be run on deployment.
* Create a `Dockerfile` to describe the image that will be deployed to
  ECS. It uses python:3.9-slim as its base, pulls in performance manager
  and its installs its dependencies using poetry.
* Create a `docker-compose.yml` file that will spin up a PostgresDB and
  a Performance Manager Image.
* Add `thing_to_run.py` file that tests out that all of the dependencies
  can be imported in the container and that the db can be connected to.
  This will eventually become the main entry point to the Performance
  Manager.
* Update README.md to describe how to setup test environment locally.

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,6 @@
+# database
+DB_HOST=local_rds
+DB_PORT=5432
+DB_NAME=performance_manager
+DB_USER=postgres
+DB_PASSWORD=postgres
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,19 @@
+FROM python:3.9-slim
+
+# Keeps Python from generating .pyc files in the container
+ENV PYTHONDONTWRITEBYTECODE 1
+# Turns off buffering for easier container logging
+ENV PYTHONUNBUFFERED 1
+
+# Install dev dependencies
+RUN apt-get update
+RUN apt-get install -y libpq-dev gcc
+
+# Install poetry
+RUN pip install -U pip
+RUN pip install "poetry==1.1.14"
+
+COPY ./performance_manager /performance_manager/
+WORKDIR /performance_manager/
+RUN poetry config virtualenvs.create false
+RUN poetry install --no-dev --no-interaction --no-ansi
--- a/README.md
+++ b/README.md
@@ -35,13 +35,10 @@ Information on the parquet table format for these file types can be found
 [here](parquet_schemas.md).
 
 ### Parquet to Relational Database
-AWS Glue, or a similar service, will be used to concatenate and join the parquet
-files. This will create intermediate data representations that will be loaded
-into an RDS for low latency access.
-
-### Relational Database Analyzed in ECS
-The relational database of all the historical data will be analyzed via Elixir.
-
+Parquet files are analyzed by the Performance Manager application to compare
+static schedule data with the real time positioning of vehicles in the field.
+The application is run inside of python3.9 image described in a Dockerfile,
+executing code found in the performance_manager directory.
 
 ## Developer Usage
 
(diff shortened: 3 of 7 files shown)


.env
```diff
@@ -0,0 +1,6 @@
+# database
```
mzappitello, line 1: new .env file that describes the db to stand up locally.

we can also add the s3 filepaths in here to use with our lambda scripts if we want.

Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM python:3.9-slim
```
mzappitello, line 1: describes the performance manager image that we'll deploy to ECS. we use this to test out locally.

it will pull in the performance manager files and install dependencies.

README.md
```diff
@@ -35,13 +35,10 @@ Information on the parquet table format for these file types can be found
...
-files. This will create intermediate data representations that will be loaded
-into an RDS for low latency access.
-
-### Relational Database Analyzed in ECS
-The relational database of all the historical data will be analyzed via Elixir.
-
+Parquet files are analyzed by the Performance Manager application to compare
+static schedule data with the real time positioning of vehicles in the field.
+The application is run inside of python3.9 image described in a Dockerfile,
+executing code found in the performance_manager directory.
```
mzappitello, line 41: update this with a more accurate description.

README.md
```diff
@@ -104,3 +101,30 @@ poetry run pylint py_gtfs_ingestion tests *.py
 # Ru pytest to run unit tests
 poetry run pytest
 ``
+
+### performance manager
```
mzappitello, line 105: steps for standing up performance manager for local testing.

docker-compose.yml
```diff
@@ -0,0 +1,26 @@
+
```
mzappitello, line 1: stand up postgres container and performance manager container for local testing.

performance_manager/pyproject.toml
```diff
@@ -0,0 +1,21 @@
+[tool.poetry]
```
mzappitello, line 1: dependencies for performance manager application. i started with things based on postgres reading and the notebook @rymarczy posted.

performance_manager/thing_to_run.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python
```
mzappitello, line 1: test file to make sure everything is working locally.

mzappitello (commented):
Here is my initial pass at creating a docker image that we'll push up to ECS and setting up a method for testing it out locally.

Obviously, lots of TODOs and unknowns still, bet a good step in the right direction.

README.md
```diff
@@ -104,3 +101,30 @@ poetry run pylint py_gtfs_ingestion tests *.py
 # Ru pytest to run unit tests
 poetry run pytest
 ``
```
rymarczy, line 103: Missing one `

Dockerfile
rymarczy: Would it be worth using an `alpine` or `debian-slim` image in combination with `asdf` and `poetry` give us a little more flexibility/control in terms of install configuration in the future? 

I think the `python-3.9-slim` is already based on a debian bullseye image, so this might just be a matter of semantics.

Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+RUN apt-get install -y libpq-dev gcc
+
+# Install poetry
+RUN pip install -U pip
+RUN pip install "poetry==1.1.14"
+
+COPY ./performance_manager /performance_manager/
+WORKDIR /performance_manager/
+RUN poetry config virtualenvs.create false
+RUN poetry install --no-dev --no-interaction --no-ansi
```
rymarczy, line 19: without a `CMD` or `ENTRYPOINT` the container exits immediately after starting. Is this the expected behavior?

rymarczy:
A few minor comments. But everything built and ran well on my end.

Dockerfile
mzappitello: for now yeah. after you stand the containers up, you run

`docker-compose run --rm performance_manager python thing_to_run.py` and it will kick off the job. once i get a better sense of how ECS Tasks work, I'll get this portion lined up correctly.

README.md
mzappitello: derp.

PR #44, opened 2022-08-10
Add sqlalchemy to performance manager

mzappitello:
* Add SQLAlchemy dependency to manage connections and calls to DataBase
      while using psycopg2 as the DBAPI layer.
    * Create lib to store all the code for performance manager that will be
      used by top level scripts.
    * Create an engine that connects to the local db for dev testing.
    * Bootstrap env variables in thing to run for faster development.
    * process static schedule data from s3 parquet files using ryans biz
      logic.
    * Create SqlBase class that all table objects will inherit from
      * Create Static Subway Headways table

### DIFF
--- a/performance_manager/lib/s3_utils.py
+++ b/performance_manager/lib/s3_utils.py
@@ -1,12 +1,21 @@
-import pyarrow.parquet as pq
-from pyarrow import fs
-from collections.abc import Iterable
-import boto3
 import logging
 import os
+from collections.abc import Iterable
+from typing import Optional, List, Tuple, Union
+
+import boto3
+import pandas
+import pyarrow.parquet as pq
+from pyarrow import fs
 
 
-def read_parquet(filename, filters=None):
+def read_parquet(
+    filename: str,
+    filters: Optional[Union[List[Tuple], List[List[Tuple]]]] = None,
+) -> pandas.core.frame.DataFrame:
+    """
+    read a parquet file from s3 and return it as a pandas dataframe
+    """
     active_fs = fs.S3FileSystem()
     file_to_load = str(filename).replace("s3://", "")
 
@@ -38,6 +47,9 @@ def file_list_from_s3(bucket_name: str, file_prefix: str) -> Iterable[str]:
         if page["KeyCount"] == 0:
             continue
         for obj in page["Contents"]:
+            # skip if this object is a "directory"
+            if obj["Size"] == 0:
+                continue
             uri = os.path.join("s3://", bucket_name, obj["Key"])
             logging.debug(uri)
             yield uri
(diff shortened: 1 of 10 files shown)


performance_manager/thing_to_run.py
```diff
@@ -1,52 +1,108 @@
...
+    get_local_engine,
+    process_static_schedule,
+)
 
-def main():
+logging.getLogger().setLevel("INFO")
+DESCRIPTION = """Entry Point to RDS Manipulating Performance Manager Scripts"""
+
+
+def load_environment() -> None:
```
mzappitello, line 23: helper function to bring in .env file if running outside of docker.

performance_manager/thing_to_run.py
```diff
@@ -1,52 +1,108 @@
...
+    """parse args for running this entrypoint script"""
+    parser = argparse.ArgumentParser(description=DESCRIPTION)
+    parser.add_argument(
+        "--experimental",
+        action="store_true",
+        dest="experimental",
+        help="if set, use a sqllite engine for quicker development",
+    )
+
+    return parser.parse_args(args)
```
mzappitello, line 61: right now its just choosing which engine to use. i'm still not sure what the best way to do this is still.

it seems like the engine should just be an attribute of the lib itself but then its hard to mess with from the command line.

performance_manager/thing_to_run.py
```diff
@@ -1,52 +1,108 @@
...
+
+
+def get_static_schedule() -> dict:
```
mzappitello, line 64: test out our static schedule parsing.

performance_manager/thing_to_run.py
```diff
@@ -1,52 +1,108 @@
...
+    things out.
+    """
+    gtfs_subway_headways = process_static_schedule("1659568605")
+
+    # The orient='records' is the key of this, it allows to align with the
+    # format mentioned in the doc to insert in bulks.
+    return gtfs_subway_headways.to_dict(orient="records")
+
+
+def write_from_dict(input_dictionary: dict, experimental: bool) -> None:
```
mzappitello, line 76: write the static subway headways to the rds via sql alchemy session.

performance_manager/lib/s3_utils.py
```diff
@@ -0,0 +1,51 @@
...
+import os
+from collections.abc import Iterable
+
+import boto3
+import pandas
+import pyarrow.parquet as pq
+from pyarrow import fs
+
+
+def read_parquet(filename: str) -> pandas.core.frame.DataFrame:
```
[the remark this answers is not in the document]
rymarczy: Can we add a `filters` parameter to this function? https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetDataset.html

I think we will not always want the whole parquet file returned as a data-frame and this would probably be the most efficient way to limit what is returned.

rymarczy:
Just one comment on the parquet to dataframe loading. Other than that, looks good to me.

PR #47, opened 2022-08-15
new script to load `VEHICLE_POSITION` events to db

rymarczy:
Workflow for this script is the following:
- check `metadataLog` table for paths to process
- group paths by same hourly folder
- process each group of paths in an hourly folder
- INSERT/UPDATE/DELETE event records in `eventsVehiclePositions` table
- if successful, UPDATE `processed' flag for paths in `metadataLog` table

Asana Task: https://app.asana.com/0/1202672148025243/1202784843755687

### DIFF
--- a/performance_manager/lib/postgres_utils.py
+++ b/performance_manager/lib/postgres_utils.py
@@ -46,7 +46,10 @@ def get_experimental_engine(
     database to be stood up. great for testing from within the shell.
     """
     engine = sa.create_engine(
-        "sqlite+pysqlite:///:memory:", echo=echo, future=True
+        "sqlite+pysqlite:///./test.db",
+        echo=echo,
+        future=True
+        # "sqlite+pysqlite:///:memory:", echo=echo, future=True
     )
     return engine
 
@@ -90,3 +93,32 @@ class StaticSubHeadway(SqlBase):  # pylint: disable=too-few-public-methods
     def __repr__(self) -> str:
         """this is just a helper string for debugging."""
         return f"( id='{self.id}', trip_id='{self.trip_id}', arrival_time='{self.arrival_time}', departure_time='{self.departure_time}', stop_id='{self.stop_id}', stop_sequence='{self.stop_sequence}', pickup_type='{self.pickup_type}', drop_off_type='{self.drop_off_type}', timepoint='{self.timepoint}', checkpoint_id='{self.checkpoint_id}', route_type='{self.route_type}', route_id='{self.route_id}', service_id='{self.service_id}', direction_id='{self.direction_id}', parent_station='{self.parent_station}', departure_time_sec='{self.departure_time_sec}', prev_departure_time_sec='{self.prev_departure_time_sec}', head_way='{self.head_way}')"
+
+
+class VehiclePositionEvents(SqlBase):  # pylint: disable=too-few-public-methods
+    """Table for GTFS-RT Vehicle Position Events"""
+
+    __tablename__ = "eventsVehiclePositions"
+
+    id = sa.Column(sa.Integer, primary_key=True)
+    current_status = sa.Column(sa.Integer, nullable=False)
+    current_stop_sequence = sa.Column(sa.Integer, nullable=True)
+    stop_id = sa.Column(sa.String(60), nullable=True)
+    timestamp_start = sa.Column(sa.Integer, nullable=False)
+    timestamp_end = sa.Column(sa.Integer, nullable=False)
+    direction_id = sa.Column(sa.Integer, nullable=True)
+    route_id = sa.Column(sa.String(60), nullable=True)
+    start_date = sa.Column(sa.Integer, nullable=True)
+    start_time = sa.Column(sa.Integer, nullable=True)
+    vehicle_id = sa.Column(sa.String(60), nullable=False)
+    hash = sa.Column(sa.Integer, nullable=False)
+
+
+class MetadataLog(SqlBase):
+    """Table for keeping track of parquet files in S3"""
+
+    __tablename__ = "metadataLog"
+
+    id = sa.Column(sa.Integer, primary_key=True)
+    processed = sa.Column(sa.Boolean, default=False)
+    path = sa.Column(sa.String(256), nullable=False, unique=True)
--- a/performance_manager/lib/postgres_utils.py
+++ b/performance_manager/lib/postgres_utils.py
@@ -108,8 +108,8 @@ class VehiclePositionEvents(SqlBase):  # pylint: disable=too-few-public-methods
     timestamp_end = sa.Column(sa.Integer, nullable=False)
     direction_id = sa.Column(sa.SmallInteger, nullable=True)
     route_id = sa.Column(sa.String(60), nullable=True)
-    start_date = sa.Column(sa.SmallInteger, nullable=True)
-    start_time = sa.Column(sa.SmallInteger, nullable=True)
+    start_date = sa.Column(sa.Integer, nullable=True)
+    start_time = sa.Column(sa.Integer, nullable=True)
     vehicle_id = sa.Column(sa.String(60), nullable=False)
     hash = sa.Column(sa.BigInteger, nullable=False)
     updated_on = sa.Column(
--- a/performance_manager/lib/postgres_utils.py
+++ b/performance_manager/lib/postgres_utils.py
@@ -100,18 +100,21 @@ class VehiclePositionEvents(SqlBase):  # pylint: disable=too-few-public-methods
 
     __tablename__ = "eventsVehiclePositions"
 
-    id = sa.Column(sa.Integer, primary_key=True)
-    current_status = sa.Column(sa.Integer, nullable=False)
-    current_stop_sequence = sa.Column(sa.Integer, nullable=True)
+    pk_id = sa.Column(sa.Integer, primary_key=True)
+    is_moving = sa.Column(sa.Boolean)
+    current_stop_sequence = sa.Column(sa.SmallInteger, nullable=True)
     stop_id = sa.Column(sa.String(60), nullable=True)
     timestamp_start = sa.Column(sa.Integer, nullable=False)
     timestamp_end = sa.Column(sa.Integer, nullable=False)
-    direction_id = sa.Column(sa.Integer, nullable=True)
+    direction_id = sa.Column(sa.SmallInteger, nullable=True)
     route_id = sa.Column(sa.String(60), nullable=True)
-    start_date = sa.Column(sa.Integer, nullable=True)
-    start_time = sa.Column(sa.Integer, nullable=True)
+    start_date = sa.Column(sa.SmallInteger, nullable=True)
+    start_time = sa.Column(sa.SmallInteger, nullable=True)
     vehicle_id = sa.Column(sa.String(60), nullable=False)
-    hash = sa.Column(sa.Integer, nullable=False)
+    hash = sa.Column(sa.BigInteger, nullable=False)
+    updated_on = sa.Column(
+        sa.DateTime, server_default=sa.func.now(), server_onupdate=sa.func.now()
+    )
 
 
 class MetadataLog(SqlBase):
@@ -119,6 +122,12 @@ class MetadataLog(SqlBase):
 
     __tablename__ = "metadataLog"
 
-    id = sa.Column(sa.Integer, primary_key=True)
+    pk_id = sa.Column(sa.Integer, primary_key=True)
     processed = sa.Column(sa.Boolean, default=False)
     path = sa.Column(sa.String(256), nullable=False, unique=True)
+    updated_on = sa.Column(
+        sa.TIMESTAMP,
+        server_default=sa.func.now(),
+        server_onupdate=sa.func.now(),
+    )
+    created_on = sa.Column(sa.TIMESTAMP, server_default=sa.func.now())
--- a/performance_manager/lib/rt_vehicle_positions.py
+++ b/performance_manager/lib/rt_vehicle_positions.py
@@ -1,0 +1,277 @@
+import logging
+
+import pandas
+import numpy
+import sqlalchemy
+import re
+import datetime
+
+from sqlalchemy import select
+from sqlalchemy import update
+from sqlalchemy import insert
+from sqlalchemy import delete
+from sqlalchemy.orm import sessionmaker
+
+from typing import Optional
+from typing import Union
+from typing import List
+
+from .s3_utils import read_parquet
+from .postgres_utils import VehiclePositionEvents
+
+
+def get_vp_dataframe(to_load: Union[str, List[str]]) -> pandas.DataFrame:
+    """
+    return a dataframe from a vehicle position parquet file (or list of files)
+    with expected columns
+    """
+    vehicle_position_cols = [
+        "current_status",
+        "current_stop_sequence",
+        "stop_id",
+        "vehicle_timestamp",
+        "direction_id",
+        "route_id",
+        "start_date",
+        "start_time",
+        "vehicle_id",
+    ]
+
+    return read_parquet(to_load).loc[:, vehicle_position_cols]
+
+
+def start_time_to_seconds(
+    time: Optional[str],
+) -> Optional[float]:
+    """
+    transform time string in HH:MM:SS format to seconds integer
+    """
+    if time is None:
+        return time
+    (hour, min, sec) = time.split(":")
+    return int(hour) * 3600 + int(min) * 60 + int(sec)
+
+
+def transform_vp_dtyes(df: pandas.DataFrame) -> pandas.DataFrame:
+    """
+    ingest dataframe of vehicle position data from parquet file and transform
+    column datatypes
+    """
+    # current_staus: 1 = MOVING, 0 = STOPPED_AT
+    df["current_status"] = numpy.where(
+        df["current_status"] != "STOPPED_AT", 1, 0
+    ).astype(numpy.byte)
+
+    # store start_date as Int64 [nullable] instead of string
+    df["start_date"] = pandas.to_numeric(df["start_date"]).astype("Int64")
+
+    # store current_stop_sequence as Int64 [nullable]
+    df["current_stop_sequence"] = pandas.to_numeric(
+        df["current_stop_sequence"]
+    ).astype("Int64")
+
+    # store direction_id as Int64 [nullable]
+    df["direction_id"] = pandas.to_numeric(df["direction_id"]).astype("Int64")
+
+    # store start_time as seconds from start of day (Int64 [nullable])
+    df["start_time"] = (
+        df["start_time"].apply(start_time_to_seconds).astype("Int64")
+    )
+
+    # hash row data, exluding vehicle_timestamp column
+    # This hash is used to identify continuous series of records with the same
+    # categorical data
+    df["hash"] = df[list(set(df.columns) - {"vehicle_timestamp"})].apply(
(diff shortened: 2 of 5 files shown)


performance_manager/lib/postgres_utils.py
```diff
@@ -90,3 +93,32 @@ class StaticSubHeadway(SqlBase):  # pylint: disable=too-few-public-methods
...
+
+
+class MetadataLog(SqlBase):
+    """Table for keeping track of parquet files in S3"""
+
+    __tablename__ = "metadataLog"
+
+    id = sa.Column(sa.Integer, primary_key=True)
+    processed = sa.Column(sa.Boolean, default=False)
+    path = sa.Column(sa.String(256), nullable=False, unique=True)
```
rymarczy, line 124: Temporary metadata table structure, I'm guessing we will want more columns in here eventually

performance_manager/lib/postgres_utils.py
```diff
@@ -90,3 +93,32 @@ class StaticSubHeadway(SqlBase):  # pylint: disable=too-few-public-methods
...
+    current_stop_sequence = sa.Column(sa.Integer, nullable=True)
+    stop_id = sa.Column(sa.String(60), nullable=True)
+    timestamp_start = sa.Column(sa.Integer, nullable=False)
+    timestamp_end = sa.Column(sa.Integer, nullable=False)
+    direction_id = sa.Column(sa.Integer, nullable=True)
+    route_id = sa.Column(sa.String(60), nullable=True)
+    start_date = sa.Column(sa.Integer, nullable=True)
+    start_time = sa.Column(sa.Integer, nullable=True)
+    vehicle_id = sa.Column(sa.String(60), nullable=False)
+    hash = sa.Column(sa.Integer, nullable=False)
```
rymarczy, line 114: This table should have all the columns we need for saving compressed vehicle position events.

performance_manager/lib/rt_vehicle_positions.py
```diff
@@ -0,0 +1,277 @@
...
+        "stop_id",
+        "vehicle_timestamp",
+        "direction_id",
+        "route_id",
+        "start_date",
+        "start_time",
+        "vehicle_id",
+    ]
+
+    return read_parquet(to_load).loc[:, vehicle_position_cols]
```
rymarczy, line 40: Load parquet file, or list of parquet files from `read_parquet` function and limit returned dataframe to relevant columns.

performance_manager/lib/rt_vehicle_positions.py
```diff
@@ -0,0 +1,277 @@
...
+def start_time_to_seconds(
+    time: Optional[str],
+) -> Optional[float]:
+    """
+    transform time string in HH:MM:SS format to seconds integer
+    """
+    if time is None:
+        return time
+    (hour, min, sec) = time.split(":")
+    return int(hour) * 3600 + int(min) * 60 + int(sec)
```
rymarczy, line 52: helper function to convert `start_time` field to numeric value

performance_manager/lib/rt_vehicle_positions.py
```diff
@@ -0,0 +1,277 @@
...
+
+
+def transform_vp_dtyes(df: pandas.DataFrame) -> pandas.DataFrame:
+    """
+    ingest dataframe of vehicle position data from parquet file and transform
+    column datatypes
+    """
```
rymarczy, line 59: Initial dataframe transformation function. Main purpose is to convert datatypes, but also add hash and drops all records with duplicate hash and timestamp values.

performance_manager/lib/rt_vehicle_positions.py
```diff
@@ -0,0 +1,277 @@
...
+    # store start_time as seconds from start of day (Int64 [nullable])
+    df["start_time"] = (
+        df["start_time"].apply(start_time_to_seconds).astype("Int64")
+    )
+
+    # hash row data, exluding vehicle_timestamp column
+    # This hash is used to identify continuous series of records with the same
+    # categorical data
+    df["hash"] = df[list(set(df.columns) - {"vehicle_timestamp"})].apply(
+        lambda row: hash(tuple(row)), axis=1
```
rymarczy, line 85: Hash is currently using the built-in python hash function, which I believe is pulling the object identity hash and sometimes concatenating it. We might want a more specific option for hashing in the future, but this seems to work for now.

performance_manager/lib/rt_vehicle_positions.py
```diff
@@ -0,0 +1,277 @@
...
+
+    return df
+
+
+def transform_vp_timestamps(df: pandas.DataFrame) -> pandas.DataFrame:
+    """
+    convert raw vehicle position data with one timestamp field into
+    vehicle position events with timestamp_start and timestamp_end
+    """
+    df = df.sort_values(by=["vehicle_id", "vehicle_timestamp"])
```
rymarczy, line 99: Convert dataframe with one timestamp column into a dataframe with 2 timestamp columns. Requires the dataframe to be sorted by `vehicle_id` and `vehicle_timestamp` all operations are array based, avoid any iterating.

performance_manager/lib/rt_vehicle_positions.py
```diff
@@ -0,0 +1,277 @@
...
+    for matching consectuive hash events, drop 2nd event because timestamp
+    has been captured in timestamp_end value of previous event
+    """
+    drop_last_mask = df["hash"] - df["hash"].shift(1) != 0
+    df = df.loc[drop_last_mask, :].reset_index(drop=True)
+
+    df.insert(0, "id", numpy.nan)
+
+    # fill all dataframe na values with Python None (for DB writing)
+    df = df.fillna(numpy.nan).replace([numpy.nan], [None])
```
rymarczy, line 135: Seems like `sqlalchemy` prefers to have all python `None` values to represent `NULL`.

performance_manager/lib/rt_vehicle_positions.py
```diff
@@ -0,0 +1,277 @@
...
+    process datetime from s3 path return UTC timestamp
+    """
+    year = int(re.findall(r"year=(\d{4})", path)[0])
+    month = int(re.findall(r"month=(\d{1,2})", path)[0])
+    day = int(re.findall(r"day=(\d{1,2})", path)[0])
+    hour = int(re.findall(r"hour=(\d{1,2})", path)[0])
+    dt = datetime.datetime(
+        year=year, month=month, day=day, hour=hour, tzinfo=datetime.timezone.utc
+    )
+    return datetime.datetime.timestamp(dt)
```
rymarczy, line 151: Get UTC timestamp from file path. This is used to identify the merge window for existing DB records. 5 minutes of either side of the hour from the file path. Might want to check our logic on bucketing/creating these paths to make sure it will align with this approach.

performance_manager/lib/rt_vehicle_positions.py
```diff
@@ -0,0 +1,277 @@
...
+            # Join records from db with records from parquet
+            # db records should contain 'index' column that does not exist in parquet dataset
+            merge_events = pandas.concat(
+                [
+                    pandas.DataFrame(
+                        [r._asdict() for r in s.execute(get_db_events)]
+                    ),
+                    new_events,
+                ]
+            ).sort_values(by=["vehicle_id", "timestamp_start"])
```
rymarczy, line 196: Pull existing records for DB that are inside of the merge window and create a `merge_events` sorted dataframe with new and existing records.

performance_manager/lib/rt_vehicle_positions.py
```diff
@@ -0,0 +1,277 @@
...
+    new_to_insert_mask = (
+        merge_events["id"].isna()
+    ) & ~last_of_consecutive_events
+    logging.info(f"Size of new insert df: {new_to_insert_mask.sum():,}")
+
+    """
+    new events that were used to update existing records, will not be used
+    """
+    new_to_drop_mask = (merge_events["id"].isna()) & last_of_consecutive_events
+    logging.info(f"Size of new being dropped df: {new_to_drop_mask.sum():,}")
```
rymarczy, line 237: All of these masks identify which records will be UPDATED/INSERTED/DELETED in the database.

performance_manager/lib/s3_utils.py
```diff
@@ -10,17 +10,20 @@
 
 
 def read_parquet(
-    filename: str,
+    filename: Union[str, List[str]],
     filters: Optional[Union[List[Tuple], List[List[Tuple]]]] = None,
 ) -> pandas.core.frame.DataFrame:
     """
-    read a parquet file from s3 and return it as a pandas dataframe
+    read parquet file or files from s3 and return it as a pandas dataframe
```
rymarczy, line 17: `ParquetDataset` has the option to ingest list of files and return of table of all those object concatenated. Updated the `read_parquet` function to utilize.

performance_manager/load_vp_events.py
```diff
@@ -0,0 +1,289 @@
...
+    get_experimental_engine,
+    get_local_engine,
+    get_vp_dataframe,
+    transform_vp_dtyes,
+    transform_vp_timestamps,
+    merge_vehicle_position_events,
+)
+
+logging.getLogger().setLevel("INFO")
+DESCRIPTION = """Entry Point to RDS Manipulating Performance Manager Scripts"""
```
rymarczy, line 29: Tried to model this after the existing `thing_to_run` script. Going forward I think we'll need a more standard way of encapsulating these job scripts to be run more uniformly.

performance_manager/load_vp_events.py
```diff
@@ -0,0 +1,289 @@
...
+    {
+        "path": "mbta-ctd-dataplatform-dev-springboard/lamp/RT_VEHICLE_POSITIONS/year=2022/month=7/day=20/hour=8/580de8523f39459486f3e34b2ca40344-0.parquet"
+    },
+    {
+        "path": "mbta-ctd-dataplatform-dev-springboard/lamp/RT_VEHICLE_POSITIONS/year=2022/month=7/day=20/hour=9/647c1c01e703466fb9679a380ad89ac4-0.parquet"
+    },
+    {
+        "path": "mbta-ctd-dataplatform-dev-springboard/lamp/RT_VEHICLE_POSITIONS/year=2022/month=7/day=20/hour=9/6635c9551e4b4c2ca64152bd167f751a-0.parquet"
+    },
+]
```
rymarczy, line 161: Paths to seed the `metadataLog` table. Temporary for debugging/demonstration purposes.

performance_manager/load_vp_events.py
```diff
@@ -0,0 +1,289 @@
...
+    """
+    create database orm objects, return engine for db operations
+    """
+    if experimental:
+        sql_engine = get_experimental_engine(echo=True)
+    else:
+        sql_engine = get_local_engine(echo=True)
+
+    SqlBase.metadata.drop_all(sql_engine)
+    SqlBase.metadata.create_all(sql_engine)
```
rymarczy, line 223: I'm dropping all tables and re-creating, this is very very temporary.

performance_manager/load_vp_events.py
```diff
@@ -0,0 +1,289 @@
...
+
+    db_session = sessionmaker(bind=sql_engine)
+
+    if seed:
+        with db_session.begin() as cur:  # type: ignore
+            cur.execute(insert(MetadataLog.__table__), LOAD_PATHS)
+            cur.commit()
```
rymarczy, line 230: Seed `metadataLog` table if CL flag is set

performance_manager/load_vp_events.py
```diff
@@ -0,0 +1,289 @@
...
+        MetadataLog.processed == 0
+    )
+    with session.begin() as s:  # type: ignore
+        paths_to_load: dict[str, dict[str, list]] = {}
+        for path_id, path in s.execute(read_md_log):
+            path = pathlib.Path(path)
+            if path.parent not in paths_to_load:
+                paths_to_load[path.parent] = {"ids": [], "paths": []}
+            paths_to_load[path.parent]["ids"].append(path_id)
+            paths_to_load[path.parent]["paths"].append(str(path))
```
rymarczy, line 256: Read `metadataLog` table for unprocessed files (processed == 0) and group by path. There's probably a much cleaner way to do this.

performance_manager/load_vp_events.py
```diff
@@ -0,0 +1,289 @@
...
+                "Size of dataframe with updated dtypes is %d",
+                new_events.shape[0],
+            )
+            new_events = transform_vp_timestamps(new_events)
+            logging.info(
+                "Size of dataframe with transformed timestamps is %d",
+                new_events.shape[0],
+            )
+
+            merge_vehicle_position_events(str(folder), new_events, session)
```
rymarczy, line 279: Run ETL logic, if successful then update `metadataLog` table for processed files.

performance_manager/lib/postgres_utils.py
```diff
@@ -46,7 +46,10 @@ def get_experimental_engine(
     database to be stood up. great for testing from within the shell.
     """
     engine = sa.create_engine(
-        "sqlite+pysqlite:///:memory:", echo=echo, future=True
+        "sqlite+pysqlite:///./test.db",
```
mzappitello, line 49: does this create the db file? or do you have to initiate it first?

rymarczy: Ah yes, that just creates the db file on the local disk. I did that for debugging because the db in memory doesn't allow other applications to attach.

performance_manager/lib/rt_vehicle_positions.py
```diff
@@ -0,0 +1,277 @@
...
+import pandas
+import numpy
+import sqlalchemy
+import re
+import datetime
+
+from sqlalchemy import select
+from sqlalchemy import update
+from sqlalchemy import insert
+from sqlalchemy import delete
```
mzappitello, line 12: i started using `import sqalchemy as sa` and prefixing everything i used.

performance_manager/lib/rt_vehicle_positions.py
mzappitello: can't we do this with the `filters` arg on `read_parquet`?

performance_manager/lib/rt_vehicle_positions.py
mzappitello: is this "flow" the only one thats going to need this functionality?

if not it might be better suited in a utils dot py file.

performance_manager/lib/rt_vehicle_positions.py
rymarczy: I thought we would be able to, but based on my testing/reading it seems like filter so meant to do a pre-SELECT on data being pulled from the parquet files. Which I don't think would really help us.

`read_pandas` actually has a `columns` parameter, but I was messing around with that at the end of last week, and that is what was causing my extremely slow download times. 

The implementation I have here seems to download the dataset immediately. The only real downside is if the dataframe size in memory becomes an issue.

performance_manager/lib/rt_vehicle_positions.py
```diff
@@ -0,0 +1,277 @@
...
+
+    # hash row data, exluding vehicle_timestamp column
+    # This hash is used to identify continuous series of records with the same
+    # categorical data
+    df["hash"] = df[list(set(df.columns) - {"vehicle_timestamp"})].apply(
+        lambda row: hash(tuple(row)), axis=1
+    )
+
+    # drop duplicates of hash and vehicle_timestamp columns (extraneous data)
+    df = df.drop_duplicates(subset=["vehicle_timestamp", "hash"])
```
mzappitello, line 89: so we drop all of them save the first one now?

performance_manager/lib/rt_vehicle_positions.py
rymarczy: I think we would also end up using it when we process `TIRP_UPDATES` events

performance_manager/lib/rt_vehicle_positions.py
rymarczy: This drop would be for records that are total duplicates. 

In the parquet files they would have different `feed_timestamp` values, but be conveying no new information from the actual vehicle. 

The transformation where we keep the first and last `vehicle_timestamp` for the same event occurs in the `transform_vp_timestamps` function.

performance_manager/load_vp_events.py
mzappitello: yeah. we def need something more. `thing_to_run` was just a script i was using for grasping around in the dark while learning how these tools worked.

performance_manager/lib/rt_vehicle_positions.py
```diff
@@ -0,0 +1,277 @@
...
+
+
+def start_time_to_seconds(
+    time: Optional[str],
+) -> Optional[float]:
+    """
+    transform time string in HH:MM:SS format to seconds integer
+    """
+    if time is None:
+        return time
```
mzappitello, line 50: this made me lol.

performance_manager/lib/rt_vehicle_positions.py
```diff
@@ -0,0 +1,277 @@
...
+    into one record with timestamp_start and timestamp_end columns
+    if event is not part of consectuive matching events, then 
+    timestamp_start and timestamp_end will be same value
+    """
+    df.rename(columns={"vehicle_timestamp": "timestamp_start"}, inplace=True)
+    df["timestamp_end"] = numpy.where(
+        (df["hash"] - df["hash"].shift(-1) == 0),
+        df["timestamp_start"].shift(-1),
+        df["timestamp_start"],
+    ).astype("int64")
```
mzappitello, line 123: if we've dropped all duplicate hashes above, won't this value always be non zero?

performance_manager/lib/rt_vehicle_positions.py
mzappitello: 👍

performance_manager/lib/rt_vehicle_positions.py
rymarczy: programming haiku

performance_manager/lib/rt_vehicle_positions.py
rymarczy: The duplicates that are dropped in `transform_vp_dtyes` are duplicated where `hash` and `vehicle_timestamp` are both the same. After that drop, we could still have multiple records with the same `hash` and different `vehicle_timestamp` values.
 
Then, right before this operation, we do a drop that keeps only the first and last `vehicle_timestmap` for consecutive records with the same `hash`. This leaves is with, at most, two records for each consecutive run of similar `hash` values. 

This operation creates our `timestamp_end` value from the `timestamp_start` of the 2nd of any records having consecutive hashes.

performance_manager/lib/rt_vehicle_positions.py
mzappitello: got it. nevermind.

performance_manager/load_vp_events.py
```diff
@@ -0,0 +1,289 @@
...
+    parsed_args = parse_args(sys.argv[1:])
+
+    session = get_db_engine(parsed_args.experimental, parsed_args.seed)
+
+    """
+    pull list of objects that need processing from metadata table
+    group objects by similar hourly folders
+    """
+    read_md_log = select((MetadataLog.id, MetadataLog.path)).where(
+        MetadataLog.processed == 0
```
mzappitello, line 247: i had to change this to `== False` on postgres.

rymarczy: Incorporated

performance_manager/lib/rt_vehicle_positions.py
rymarczy: Incorporated

PR #48, opened 2022-08-16
Create startup dot py to launch application with

mzappitello:
startup dot py is the entry point into the performance manager
processing event loop. its main processes schedules repeated calls to a
function that will check the metadata table for unprocessed parquet
files. if it finds them, they are passed to `process_vehicle_positions`,
a function that converts vehicle position gtfs_rt parquet files into row
entries in the vehicle position events table using the same logic that
was in (the now deleted) load_vp_events dot py script. the performance
manager image is now launched with a startup dot py invocation.


performance_manager/lib/rt_vehicle_positions.py
```diff
@@ -275,3 +290,45 @@ def merge_vehicle_position_events(
...
+            logging.info("Error Processing Vehicle Positions")
+            logging.exception(e)
+        else:
+            update_md_log = (
+                sa.update(MetadataLog.__table__)
+                .where(MetadataLog.pk_id.in_(ids))
+                .values(processed=1)
+            )
+            with sql_session.begin() as session:  # type: ignore
+                session.execute(update_md_log)
```
mzappitello, line 334: this does all of the processing. its uses all the above methods and is the only one in lib/__init__.py.

performance_manager/pyproject.toml
```diff
@@ -38,8 +38,10 @@ disable = [
   "broad-except",
   # caught by black
   "line-too-long",
+  # we're logging everything so its "ok"
+  "lost-exception"
```
mzappitello, line 42: seems fine.

performance_manager/pyproject.toml
```diff
@@ -38,8 +38,10 @@ disable = [
...
 ]
-good-names = ["e", "i"]
+good-names = ["e", "i", "s"]
```
mzappitello, line 44: i added this so that we can get by `with session.begin() as s:` but think we should probably remove it when we move a bunch of stuff to postgres_utils.py.

performance_manager/pyproject.toml
```diff
@@ -53,4 +55,7 @@ disallow_untyped_defs = true
 warn_unused_ignores = true
 warn_unreachable = true
 pretty = true
-plugins = "sqlmypy"
+plugins = [
+  "sqlalchemy.ext.mypy.plugin"
+]
```
mzappitello, line 60: i changed which sqalchemy mypy plugin we use from dropbox's custom one to sqlalchemy's shipped one. it solves some problems but creates other. it feels like it will make the 2.0 switch easier whenever that happens.

performance_manager/startup.py
```diff
@@ -0,0 +1,165 @@
...
+    process_vehicle_positions,
+)
+
+logging.getLogger().setLevel("INFO")
+
+DESCRIPTION = """Entry Point to RDS Manipulating Performance Manager Scripts"""
+HERE = os.path.dirname(os.path.abspath(__file__))
+
+
+def load_environment() -> None:
```
mzappitello, line 31: same load function.

performance_manager/startup.py
```diff
@@ -0,0 +1,165 @@
...
+                logging.info("setting %s to %s", key, value)
+                os.environ[key] = value
+
+    except Exception as exception:
+        logging.error("error while trying to bootstrap")
+        logging.exception(exception)
+        raise exception
+
+
+def parse_args(args: list[str]) -> argparse.Namespace:
```
mzappitello, line 58: same parse args plus an interval argument for how frequently to run the event loop.

performance_manager/startup.py
```diff
@@ -0,0 +1,165 @@
...
+        "--seed",
+        action="store_true",
+        dest="seed",
+        help="if set, seed metadataLog table with paths",
+    )
+
+    return parser.parse_args(args)
+
+
+def seed_metadata_table(sql_session: sessionmaker) -> None:
```
mzappitello, line 85: seed metadata function. it would be cool if it took a filepath.

might be a good idea to have this in a different script so we can insert stuff into the db on the fly while startup dot py is running.

performance_manager/startup.py
```diff
@@ -0,0 +1,165 @@
...
+
+        with sql_session.begin() as session:  # type: ignore
+            session.execute(sa.insert(MetadataLog.__table__), load_paths)
+            session.commit()
+    except Exception as e:
+        logging.error("Error cleaning and seeding database")
+        logging.exception(e)
+
+
+def unprocessed_events(sql_session: sessionmaker) -> Dict[str, Dict[str, List]]:
```
mzappitello, line 103: we should probably move this to the library, but it gets paths to process.

performance_manager/startup.py
```diff
@@ -0,0 +1,165 @@
...
+
+    except Exception as e:
+        logging.error("Error searching for unprocessed events")
+        logging.exception(e)
+
+    finally:
+        return paths_to_load
+
+
+def main(args: argparse.Namespace) -> None:
```
mzappitello, line 126: main entrypoint that schedules the work.

performance_manager/tests/july_17_filepaths.json
```diff
@@ -1,31 +1,4 @@
-#!/usr/bin/env python
```
mzappitello, line 1: removed this file now that we no longer need it.

i moved the "files to load" list and put it into a dict in tests.

git thinks its a rename and edit, which, lol.

performance_manager/thing_to_run.py
```diff
@@ -1,108 +0,0 @@
-#!/usr/bin/env python
```
mzappitello, line 1: its purpose fulfilled, `thing_to_run.py` is no more.

mzappitello (commented):
now we have an event loop that checks for unprocessed vehicle position parquet files and processes them!

performance_manager/lib/rt_vehicle_positions.py
```diff
@@ -175,23 +183,31 @@ def merge_vehicle_position_events(
...
+        # Join records from db with records from parquet db records should
+        # contain 'index' column that does not exist in parquet dataset
+        return pandas.concat(
+            [
+                pandas.DataFrame(
+                    [row._asdict() for row in session.execute(get_db_events)]
+                ),
+                new_events,
+            ]
+        ).sort_values(by=["vehicle_id", "timestamp_start"])
```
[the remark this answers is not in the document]
rymarczy: Does this work in postgress? In SQLite I believe if there are no results from a SELECT, a `None` gets returned and this comprehension would throw an error.

performance_manager/lib/rt_vehicle_positions.py
```diff
@@ -143,23 +153,21 @@ def get_utc_timestamp(path: str) -> float:
...
 
 
-def merge_vehicle_position_events(
+def get_event_overlap(
     folder: str,
     new_events: pandas.DataFrame,
-    session: sqlalchemy.orm.session.sessionmaker,
-) -> None:
+    sql_session: sessionmaker,
+) -> pandas.DataFram:
```
rymarczy, line 166: `pandas.DataFrame`

performance_manager/startup.py
```diff
@@ -0,0 +1,165 @@
...
+        read_md_log = sa.select((MetadataLog.pk_id, MetadataLog.path)).where(
+            MetadataLog.processed == sa.false()
+        )
+        with sql_session.begin() as session:  # type: ignore
+            for path_id, path in session.execute(read_md_log):
+                path = pathlib.Path(path)
+                if path.parent not in paths_to_load:
+                    paths_to_load[path.parent] = {"ids": [], "paths": []}
+                paths_to_load[path.parent]["ids"].append(path_id)
+                paths_to_load[path.parent]["paths"].append(str(path))
```
rymarczy, line 116: I think we might want to move this into `lib/rt_vehicle_positions.py`. Each job will probably be performing actions on it's own segment of Metadata paths. So this probably doesn't need to be in the main event loop.

rymarczy:
A few comments on this. Big ones being moving the `MetadataLog` SELECT into the library file and the merge logic on the existing and new events.

performance_manager/startup.py
mzappitello: i moved this to postgres utils. i think its going to end up returning all new files, not just the vehicle positions ones.

performance_manager/lib/rt_vehicle_positions.py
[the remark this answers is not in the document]
mzappitello: it really stinks that there is such a wide discrepancy in dialects. this seems to work correctly in postgres. i would have assumed part of what sqlalchemy is doing is ironing out these differences.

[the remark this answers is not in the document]
mzappitello: that said, i didn't see this error occur on sqlite either?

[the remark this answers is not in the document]
rymarczy: Sorry, just tried this out again and you're correct. If there are no results from the SELECT, the comprehension just returns and empty list. Not sure what I was looking at before where the `None` value was returned.

rymarczy:
LGTM

PR #49, opened 2022-08-19
Trip update event load

rymarczy:
This PR updates `performance_manager` to ingest `RT_TRIP_UPDATES` parquet files, along with the following:
- create `lib/gtfs_utils.py` to hold common gtfs processing functions.
- update `july_17_filepaths.json` with `RT_TRIP_UPDATES` files
- move path selection into event library files

Asana Task: https://app.asana.com/0/1202784843755682/1202798609780869


performance_manager/lib/gtfs_utils.py
```diff
@@ -0,0 +1,43 @@
+from typing import Optional
+
+import pandas
+
+
```
rymarczy, line 5: These functions are common to `RT_VEHICLE_POSITIONS` and `RT_TRIP_UPDATES` so moved into a new file

performance_manager/lib/postgres_utils.py
```diff
@@ -147,6 +146,27 @@ class VehiclePositionEvents(SqlBase):  # pylint: disable=too-few-public-methods
     )
 
 
+class TripUpdateEvents(SqlBase):  # pylint: disable=too-few-public-methods
+    """Table for GTFS-RT Trip Update Predicted Stop Events"""
+
+    __tablename__ = "eventsTripUpdates"
```
rymarczy, line 152: Added trip updates schema

performance_manager/lib/gtfs_utils.py
```diff
@@ -0,0 +1,43 @@
...
+def add_event_hash_column(
```
rymarczy, line 6: Move common functions into utils file

performance_manager/lib/postgres_utils.py
```diff
@@ -147,6 +147,27 @@ class VehiclePositionEvents(SqlBase):  # pylint: disable=too-few-public-methods
...
```
rymarczy, line 153: add trip updates schema

performance_manager/lib/postgres_utils.py
```diff
@@ -163,12 +184,15 @@ class MetadataLog(SqlBase):  # pylint: disable=too-few-public-methods
     created_on = sa.Column(sa.TIMESTAMP, server_default=sa.func.now())
 
 
-def unprocessed_files(sql_session: sessionmaker) -> Dict[str, Dict[str, List]]:
+def get_unprocessed_files(
+    path_contains: str, sql_session: sessionmaker
+) -> Dict[str, Dict[str, List]]:
```
rymarczy, line 189: Gave this function a `path_contains` string to be used by `RT_TRIP_UPDATES` and `RT_VEHICLE_POSITIONS`

performance_manager/lib/rt_trip_updates.py
```diff
@@ -0,0 +1,250 @@
...
+        ("timestamp", ">", 0),
+        ("route_id", "!=", "None"),
+        ("start_date", "!=", "None"),
+        ("start_time", "!=", "None"),
+        ("vehicle_id", "!=", "None"),
+    ]
+
+    return read_parquet(
+        to_load, columns=trip_update_columns, filters=trip_update_filters
+    )
```
rymarczy, line 40: using the `filters` parameter to limit records straight out of the parquet files

performance_manager/lib/rt_trip_updates.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+def explode_stop_time_update(row: pandas.Series) -> Optional[List[dict]]:
+    """
+    function to be used with pandas `apply` dataframe method
+    explode nested list of dicts in stop_time_update column
+    """
```
rymarczy, line 47: This the function that is "applied" to every record on our dataframe from the parquet files. This is what results in high-ish memory usage. We'll have to keep an eye on resources in the ECS, if they become an issue, this is the most likely culprit.

performance_manager/lib/rt_trip_updates.py
```diff
@@ -0,0 +1,250 @@
...
+    return return_list
+
+
+def unwrap_tu_dataframe(events: pandas.DataFrame) -> pandas.DataFrame:
+    """
+    unwrap and explode trip updates records from parquet files
+    parquet files contain stop_time_update field that is saved as list of dicts
+    stop_time_update must have fields extracted and flattened to create
+    predicted trip update stop events
+    """
```
rymarczy, line 88: This function gets the dataframe ready for merging with existing records in the ECS. Very similar to work flow of `RT_VEHICLE_POSITIONS` events.

performance_manager/lib/rt_trip_updates.py
```diff
@@ -0,0 +1,250 @@
...
+    return events
+
+
+def merge_trip_update_events(
+    new_events: pandas.DataFrame, session: sa.orm.session.sessionmaker
+) -> None:
+    """
+    merge new trip update evetns with existing events found in database
+    merge performed on hash of records
+    """
```
rymarczy, line 137: This is also very similar to the `RT_VEHICLE_POSITIONS` logic, however these events only have `timestamp_start` no `timestamp_end` and they are all "stopped_at" events.

performance_manager/lib/rt_trip_updates.py
```diff
@@ -0,0 +1,250 @@
...
+            logging.info("Error Processing Trip Updates")
+            logging.exception(e)
+        else:
+            update_md_log = (
+                sa.update(MetadataLog.__table__)
+                .where(MetadataLog.pk_id.in_(ids))
+                .values(processed=1)
+            )
+            with sql_session.begin() as cursor:  # type: ignore
+                cursor.execute(update_md_log)
```
rymarczy, line 250: Entry point for processing trip update files, just needs an `sql_session`

performance_manager/lib/rt_vehicle_positions.py
```diff
@@ -33,18 +38,6 @@ def get_vp_dataframe(to_load: Union[str, List[str]]) -> pandas.DataFrame:
...
-) -> Optional[float]:
-    """
-    transform time string in HH:MM:SS format to seconds integer
-    """
-    if time is None:
-        return time
-    (hour, minute, second) = time.split(":")
-    return int(hour) * 3600 + int(minute) * 60 + int(second)
-
-
```
rymarczy, line 47: Moved to `gtfs_utils.py`

performance_manager/lib/rt_vehicle_positions.py
```diff
@@ -61,10 +54,13 @@ def transform_vp_dtyes(vehicle_positions: pandas.DataFrame) -> pandas.DataFrame:
...
 
-    # store current_stop_sequence as Int64 [nullable]
-    vehicle_positions["current_stop_sequence"] = pandas.to_numeric(
+    # store current_stop_sequence as stop_sequence Int64 [nullable]
+    vehicle_positions["stop_sequence"] = pandas.to_numeric(
         vehicle_positions["current_stop_sequence"]
     ).astype("Int64")
+    vehicle_positions = vehicle_positions.drop(
+        columns=["current_stop_sequence"]
+    )
```
rymarczy, line 63: Changed column name to work with `add_event_hash_column` function.

performance_manager/lib/rt_vehicle_positions.py
```diff
@@ -81,9 +77,7 @@ def transform_vp_dtyes(vehicle_positions: pandas.DataFrame) -> pandas.DataFrame:
     # hash row data, exluding vehicle_timestamp column
     # This hash is used to identify continuous series of records with the same
     # categorical data
-    vehicle_positions["hash"] = vehicle_positions[
-        list(set(vehicle_positions.columns) - {"vehicle_timestamp"})
-    ].apply(lambda row: hash(tuple(row)), axis=1)
+    vehicle_positions = add_event_hash_column(vehicle_positions)
```
rymarczy, line 80: replace in-line `apply` with `add_event_hash_column` function

performance_manager/lib/rt_vehicle_positions.py
```diff
@@ -292,14 +286,16 @@ def merge_vehicle_position_events(
...
-) -> None:
+def process_vehicle_positions(sql_session: sessionmaker) -> None:
     """
     process a bunch of vehicle position files
     create events for them
     merge those events with existing events
     """
+
+    # check metadata table for unprocessed parquet files
+    paths_to_load = get_unprocessed_files("RT_VEHICLE_POSITIONS", sql_session)
```
rymarczy, line 297: moved `get_unprocessed_files` call into library entry point

performance_manager/lib/s3_utils.py
```diff
@@ -12,7 +12,7 @@
 def read_parquet(
     filename: Union[str, List[str]],
     columns: Union[List[str], slice] = slice(None),
-    filters: Optional[Union[List[Tuple], List[List[Tuple]]]] = None,
+    filters: Optional[Union[Sequence[Tuple], Sequence[List[Tuple]]]] = None,
```
rymarczy, line 15: mypy or pylint complained about this, can't remember which one

performance_manager/startup.py
```diff
@@ -125,8 +125,8 @@ def iteration() -> None:
         """function to invoke on a scheduled routine"""
         logging.info("Entering Iteration")
 
-        paths = unprocessed_files(sql_session)
-        process_vehicle_positions(paths, sql_session)
+        process_vehicle_positions(sql_session)
+        process_trip_updates(sql_session)
```
rymarczy, line 129: just call each entry point and give it an sql_session

performance_manager/lib/gtfs_utils.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def start_time_to_seconds(
+    time: Optional[str],
+) -> Optional[float]:
+    """
+    transform time string in HH:MM:SS format to seconds
+    """
+    if time is None:
+        return time
```
mzappitello, line 41: still makes me lol.

performance_manager/lib/postgres_utils.py
mzappitello: 👍

performance_manager/lib/postgres_utils.py
mzappitello: i kinda feel like we should move the sql base stuff to its own file. schema stuff in one place, utilities in another.

something to keep in mind moving forward.

performance_manager/lib/rt_trip_updates.py
mzappitello: 👍

performance_manager/lib/rt_trip_updates.py
mzappitello: is there a way that we can log the memory usage we're getting after applying this explode func?

rymarczy: Our best bet would probably be to use python's `subprocess` library and grep `free` output???

performance_manager/lib/rt_trip_updates.py
```diff
@@ -0,0 +1,250 @@
...
+    for record in row["stop_time_update"]:
+        try:
+            arrival_time = int(record["arrival"]["time"])
+        except (TypeError, KeyError):
+            continue
+        if (
+            arrival_time - row["timestamp"] < 0
+            or arrival_time - row["timestamp"] > 120
+        ):
+            continue
```
mzappitello, line 66: a comment on why the time has to be in this window would be helpful for future us.

performance_manager/lib/rt_trip_updates.py
```diff
@@ -0,0 +1,250 @@
...
+            {
+                "stop_id": record.get("stop_id"),
+                "stop_sequence": record.get("stop_sequence"),
+                "timestamp_start": arrival_time,
+            }
+        )
+        return_list.append(append_dict.copy())
+
+    if len(return_list) == 0:
+        return None
```
mzappitello, line 77: why return `None` instead of the empty list?

performance_manager/lib/rt_trip_updates.py
```diff
@@ -0,0 +1,250 @@
...
+        "start_date": row["start_date"],
+        "start_time": row["start_time"],
+        "vehicle_id": row["vehicle_id"],
+    }
+    return_list: List[Dict[str, Any]] = []
+    for record in row["stop_time_update"]:
+        try:
+            arrival_time = int(record["arrival"]["time"])
+        except (TypeError, KeyError):
+            continue
```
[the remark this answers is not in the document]
rymarczy: `TypeError` because `record["arrival"]["time"]` can be None, which would throw a `TypeError` on the int conversion.

performance_manager/lib/rt_trip_updates.py
rymarczy: The way I have it currently set up, this `apply` function returns a Series and `.dropna()` is immediately called to get rid of the records that returned `None`.  Returning the empty list doesn't allow for the use of `dropna` and also causes some issues with the next step in the business logic where we explode this series and call `json_normalize` to turn the dictionaries into columns.

performance_manager/lib/rt_trip_updates.py
rymarczy: Will add a comment on this, the 120 value selection is kind of arbitrary. We're just filtering out predictions that are farther into the future and unlikely to be used as the final stop event prediction. 2 minutes seemed like a conservative value.

performance_manager/lib/rt_trip_updates.py
rymarczy: We would also have to thread the `apply` call and whatever method we have for logging the memory usage, since they would have to run concurrently.

performance_manager/lib/rt_trip_updates.py
mzappitello: i think its fine to keep these as different flows. if we end up having a third that has commonality, it might be worth doing some abstraction, but this all reads clear enough to me.

performance_manager/lib/rt_vehicle_positions.py
mzappitello: 👍

performance_manager/lib/rt_vehicle_positions.py
mzappitello: weird that there isn't built in functionality to change the name.

performance_manager/lib/rt_vehicle_positions.py
mzappitello: 👍

mzappitello:
i had a few small things, but LGTM. i had a few small things in there and think a few comments could be really helpful for future us.

performance_manager/lib/rt_vehicle_positions.py
rymarczy: There is, that would be more memory efficient so I will incorporate

performance_manager/lib/rt_trip_updates.py
```diff
@@ -0,0 +1,283 @@
...
+    """
+    while True:
+        mb_memory_used = int(psutil.virtual_memory().used / 1_000_000)
+        logging.info("%d MB of memory currently used", mb_memory_used)
+        try:
+            complete_queue.get(timeout=interval)
+        except Empty:
+            continue
+        else:
+            break
```
rymarczy, line 101: function to be threaded for logging memory usage

performance_manager/lib/rt_trip_updates.py
```diff
@@ -0,0 +1,283 @@
...
+    events["start_time"] = (
+        events["start_time"].apply(start_time_to_seconds).astype("int64")
+    )
+
+    # start thread to log memory usage during apply operation
+    complete_queue: SimpleQueue = SimpleQueue()
+    log_thread = threading.Thread(
+        target=log_memory_usage, args=(complete_queue,)
+    )
+    log_thread.start()
```
rymarczy, line 132: create and start thread that will be used for logging memory usage during `apply` method

performance_manager/lib/rt_trip_updates.py
```diff
@@ -0,0 +1,283 @@
...
+
+    # expand and filter stop_time_update column
+    # this will return a Series with values being list of dicts
+    events = events.apply(explode_stop_time_update, axis=1).dropna()
+
+    # send signal to stop memory usage logging and join thread
+    complete_queue.put(True)
+    log_thread.join()
```
rymarczy, line 140: clean up logging thread

performance_manager/pyproject.toml
```diff
@@ -13,6 +13,7 @@ pandas = "^1.4.3"
 numpy = "^1.23.1"
 psycopg2 = "^2.9.3"
 SQLAlchemy = "^1.4.39"
+psutil = "^5.9.1"
```
rymarczy, line 16: added psutil requirement for logging memory usage

PR #50, opened 2022-08-23
Write Processed Parquet Files to Metadata Table

mzappitello:
The performance manager application will check metadata tables for new
parquet files that can then be converted from gtfs position data into
events data used to compute dwell times, headways, and travel times.

* Replace `parquet.write_to_dataset` call with
  `pyarrow.dataset.write_dataset` call. There is a known bug with the
  former that prevents the `file_visitor` call back function from being
  executed properly. That function is used to keep track of what parquet
  files are written and to update the metadata table appropriately.
* Add psycopg2 to dependencies and add postgres utils dot py to the
  library. It contains a function to build a connection to our database
  and another to insert a filepath into the metadata table.
* Add load_environment function to py_gtfs_rt_ingestion project and use
  it on both the batch and ingest scripts for easier local testing. This
  reduces the reliance on command line options for passing in
  environment variables.

### DIFF
--- a/.env
+++ b/.env
@@ -12,4 +12,4 @@ DB_PASSWORD=postgres
 EXPORT_BUCKET=mbta-ctd-dataplatform-dev-springboard
 ARCHIVE_BUCKET=mbta-ctd-dataplatform-dev-archive
 ERROR_BUCKET=mbta-ctd-dataplatform-dev-error
-IMPORT_BUCKET=mbta-ctd-dataplatform-dev-incoming
+INGEST_BUCKET=mbta-ctd-dataplatform-dev-incoming
--- a/py_gtfs_rt_ingestion/batch_files.py
+++ b/py_gtfs_rt_ingestion/batch_files.py
@@ -14,6 +14,7 @@ from py_gtfs_rt_ingestion import LambdaContext
 from py_gtfs_rt_ingestion import LambdaDict
 from py_gtfs_rt_ingestion import batch_files
 from py_gtfs_rt_ingestion import file_list_from_s3
+from py_gtfs_rt_ingestion import load_environment
 
 
 logging.getLogger().setLevel("INFO")
@@ -45,14 +46,6 @@ def parse_args(args: list[str]) -> dict:
     )
 
     parser.add_argument(
-        "--s3-bucket",
-        dest="s3_bucket",
-        type=str,
-        default="mbta-gtfs-s3",
-        help="prefix to files in the mbta-gtfs-s3 bucket",
-    )
-
-    parser.add_argument(
         "--threshold",
         dest="filesize_threshold",
         type=int,
@@ -74,14 +67,7 @@ def parse_args(args: list[str]) -> dict:
         help="print out each event as json to stdout",
     )
 
-    parsed_args = parser.parse_args(args)
-    event = vars(parsed_args)
-
-    if parsed_args.s3_bucket is not None:
-        os.environ["INGEST_BUCKET"] = parsed_args.s3_bucket
-        del event["s3_bucket"]
-
-    return event
+    return vars(parser.parse_args(args))
 
 
 def main(batch_args: BatchArgs) -> None:
@@ -157,6 +143,7 @@ def lambda_handler(event: LambdaDict, context: LambdaContext) -> None:
 
 
 if __name__ == "__main__":
+    load_environment()
     parsed_event = parse_args(sys.argv[1:])
     parsed_context = LambdaContext()
     lambda_handler(parsed_event, parsed_context)
--- a/py_gtfs_rt_ingestion/ingest.py
+++ b/py_gtfs_rt_ingestion/ingest.py
@@ -8,9 +8,9 @@ import sys
 
 from typing import Dict, Union
 
-import pyarrow.parquet as pq
-
+import pyarrow.dataset as ds
 from pyarrow import fs
+from pyarrow.util import guid
 
 from py_gtfs_rt_ingestion import ArgumentException
 from py_gtfs_rt_ingestion import ConfigType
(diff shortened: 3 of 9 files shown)


.env
```diff
@@ -12,4 +12,4 @@ DB_PASSWORD=postgres
 EXPORT_BUCKET=mbta-ctd-dataplatform-dev-springboard
 ARCHIVE_BUCKET=mbta-ctd-dataplatform-dev-archive
 ERROR_BUCKET=mbta-ctd-dataplatform-dev-error
-IMPORT_BUCKET=mbta-ctd-dataplatform-dev-incoming
+INGEST_BUCKET=mbta-ctd-dataplatform-dev-incoming
```
mzappitello, line 15: keep us consistent with whats going on in the lambda side of things.

we should eventually standardize these.

py_gtfs_rt_ingestion/batch_files.py
```diff
@@ -44,14 +45,6 @@ def parse_args(args: list[str]) -> dict:
...
     )
 
-    parser.add_argument(
-        "--s3-bucket",
-        dest="s3_bucket",
-        type=str,
-        default="mbta-gtfs-s3",
-        help="prefix to files in the mbta-gtfs-s3 bucket",
-    )
-
```
mzappitello, line 54: get env vars from `.env` file now.

py_gtfs_rt_ingestion/batch_files.py
```diff
@@ -74,14 +67,7 @@ def parse_args(args: list[str]) -> dict:
...
 
-    parsed_args = parser.parse_args(args)
-    event = vars(parsed_args)
-
-    if parsed_args.s3_bucket is not None:
-        os.environ["INGEST_BUCKET"] = parsed_args.s3_bucket
-        del event["s3_bucket"]
-
-    return event
+    return vars(parser.parse_args(args))
```
mzappitello, line 70: 👋

py_gtfs_rt_ingestion/batch_files.py
```diff
@@ -157,6 +143,7 @@ def lambda_handler(event: LambdaDict, context: LambdaContext) -> None:
 
 
 if __name__ == "__main__":
+    load_environment()
```
mzappitello, line 146: do this now instead.

py_gtfs_rt_ingestion/ingest.py
```diff
@@ -8,16 +8,18 @@
 
 from typing import Dict, Union
 
-import pyarrow.parquet as pq
-
+import pyarrow.dataset as ds
 from pyarrow import fs
+from pyarrow.util import guid
```
mzappitello, line 13: replacing how we call things.

py_gtfs_rt_ingestion/ingest.py
```diff
@@ -46,33 +48,8 @@ def parse_args(args: list[str]) -> Union[LambdaDict, list[LambdaDict]]:
...
-        help="where to archive ingested files to",
-    )
-
-    parser.add_argument(
-        "--error",
-        dest="error_dir",
-        type=str,
-        help="where to move unconverted files to",
-    )
-
```
mzappitello, line 66: get these from .env file.

py_gtfs_rt_ingestion/ingest.py
```diff
@@ -110,19 +87,42 @@ def main(event: Dict) -> None:
...
         for s3_prefix, table in converter.convert(files):
+            # output path is the springboard bucket plus the filetype
             s3_path = os.path.join(export_bucket, s3_prefix)
-            logging.info("Writing Table to %s", s3_path)
-            pq.write_to_dataset(
-                table=table,
-                root_path=s3_path,
-                filesystem=s3_filesystem,
-                partition_cols=converter.partition_cols,
```
mzappitello, line 123: this function does not work correctly with the `file_visitor` argument.

https://issues.apache.org/jira/browse/ARROW-17068

(they claim its fixed, but its not after looking at source code)

instead, mimic what it was already doing (creating a `Partitioning` object, select defaults for `basename_template`, etc) and call `write_dataset` instead.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/postgres_utils.py
```diff
@@ -0,0 +1,55 @@
+import logging
```
mzappitello, line 1: this should all be pretty straight forward.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/utils.py
```diff
@@ -0,0 +1,28 @@
+import os
```
mzappitello, line 1: we're already using this in performance manager. not sure what the best place for it is since its useful in multiple places.

py_gtfs_rt_ingestion/ingest.py
```diff
@@ -110,19 +87,42 @@ def main(event: Dict) -> None:
...
+                insert_metadata(written_file.path)
+
+            ds.write_dataset(
+                data=table,
+                base_dir=s3_path,
+                filesystem=fs.S3FileSystem(),
+                format=ds.ParquetFileFormat(),
+                partitioning=partitioning,
+                file_visitor=post_processing,
+                basename_template=guid() + "-{i}.parquet",
```
rymarczy, line 120: Any specific reason for the `guid()` and `-{i}`?

mzappitello: this is the default way `write_to_dataset` was generating the basenames for parquet files. if we had a different way we wanted to do this, it would also be fine, but whatever we do would probably be similar.

PR #52, opened 2022-08-26
Deploy Lambda using Psycopg2 with libpq

mzappitello:
Lambda containers don't come with libpq (the postgres library) installed
by default. As psycopg2 is pretty much a wrapper around this c library,
it throws errors when you try to import it, and our lambda cannot run.

When packaging our zip file for deployment, use a github project that
builds the libpq library into psycopy2 instead.

### DIFF
--- a/.github/actions/lambda_deploy/action.yaml
+++ b/.github/actions/lambda_deploy/action.yaml
@@ -24,11 +24,24 @@ runs:
       with:
         poetry-dir: py_gtfs_rt_ingestion
 
+    - name: Download psycopg2 with libpq installed
+      uses: actions/checkout@v2
+      with:
+        # this repo has a bunch of stars and bundles psycopg with libpq
+        repository: jkehler/awslambda-psycopg2
+        path: psycopg_with_libpq
+
     - name: Create build folder
       working-directory: py_gtfs_rt_ingestion
       run: |
+        # build the ingestion modules and install its dependencies to package
         poetry build
         poetry run pip install --upgrade -t package dist/*.whl
+        # remove the shipped version of psycopg2 and replace it with the one
+        # with libpq installed
+        rm -r package/psycopg2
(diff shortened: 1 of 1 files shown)


.github/actions/lambda_deploy/action.yaml
```diff
@@ -24,11 +24,24 @@ runs:
       with:
         poetry-dir: py_gtfs_rt_ingestion
 
+    - name: Download psycopg2 with libpq installed
+      uses: actions/checkout@v2
+      with:
+        # this repo has a bunch of stars and bundles psycopg with libpq
+        repository: jkehler/awslambda-psycopg2
+        path: psycopg_with_libpq
+
```
mzappitello, line 33: i tried for a while to build the thing ourselves so that we could select what versions of postgres and psycopg we're building with, but was unsuccesful as the containers on github actions are missing some things we would need. we could try to build our own artifacts on ubuntu later and use them instead. this seems to work ok for now though.

.github/actions/lambda_deploy/action.yaml
```diff
@@ -24,11 +24,24 @@ runs:
...
     - name: Create build folder
       working-directory: py_gtfs_rt_ingestion
       run: |
+        # build the ingestion modules and install its dependencies to package
         poetry build
         poetry run pip install --upgrade -t package dist/*.whl
+        # remove the shipped version of psycopg2 and replace it with the one
+        # with libpq installed
+        rm -r package/psycopg2
+        mv ../psycopg_with_libpq/psycopg2-3.9 package/psycopg2
```
mzappitello, line 43: this is the big change. use this version thats bundled with libpq instead of the one we install with pip.

mzappitello: we're probably wasting some time installing psycopg2 with pip here, but i didn't want to waste more time mucking around with the .whl file to remove it.

PR #53, opened 2022-08-30
Generate DB Auth Tokens for AWS RDS Connections

mzappitello:
Infra has added an IAM user that we can use to connect to our database
from inside the ECS container and Lambda functions. Get the token for
this user and use it to make a connection to the RDS.

* update lambda deploy action to install the ssl cert into the zip file
* add logic to batching lambda to debug connection issues

### DIFF
--- a/.github/actions/lambda_deploy/action.yaml
+++ b/.github/actions/lambda_deploy/action.yaml
@@ -24,6 +24,17 @@ runs:
       with:
         poetry-dir: py_gtfs_rt_ingestion
 
+    - name: Create build folder
+      working-directory: py_gtfs_rt_ingestion
+      run: |
+        # build the ingestion modules and install its dependencies to package
+        poetry build
+        poetry run pip install --upgrade -t package dist/*.whl
+        # add the batch and ingest scripts to the package
+        cp -v batch_files.py package/batch_files.py
+        cp -v ingest.py package/ingest.py
+      shell: bash
+
     - name: Download psycopg2 with libpq installed
       uses: actions/checkout@v2
       with:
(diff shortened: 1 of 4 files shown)


.github/actions/lambda_deploy/action.yaml
```diff
@@ -24,26 +24,38 @@ runs:
...
-        poetry build
-        poetry run pip install --upgrade -t package dist/*.whl
         # remove the shipped version of psycopg2 and replace it with the one
         # with libpq installed
         rm -r package/psycopg2
         mv ../psycopg_with_libpq/psycopg2-3.9 package/psycopg2
-        # add the batch and ingest scripts to the package
-        cp -v batch_files.py package/batch_files.py
-        cp -v ingest.py package/ingest.py
+      shell: bash
```
mzappitello, line 52: i reordered this stuff and put it in different sets of instructions in a way that felt a bit better.

.github/actions/lambda_deploy/action.yaml
```diff
@@ -24,26 +24,38 @@ runs:
...
+
+    - name: Install AWS SSL Cert
+      working-directory: py_gtfs_rt_ingestion/package
+      run: |
+        curl https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem -o aws-cert-bundle.pem
+        echo "acaf8712f8d71c05f85503c6b90fd0127e95ff0091bf094a22a650119684a08e  aws-cert-bundle.pem" | sha256sum -c -
```
mzappitello, line 58: only new bit is adding the pem file and making sure checksums match.

py_gtfs_rt_ingestion/batch_files.py
```diff
@@ -29,6 +30,7 @@ class BatchArgs(NamedTuple):
     s3_prefix: str = DEFAULT_S3_PREFIX
     print_events: bool = False
     dry_run: bool = False
+    debug_rds_connection: bool = False
```
mzappitello, line 33: new bool to add to try debugging connections.

not a _huge_ fan of debugging this way, but it was all i could think to do.

py_gtfs_rt_ingestion/batch_files.py
```diff
@@ -130,12 +132,13 @@ def lambda_handler(event: LambdaDict, context: LambdaContext) -> None:
     logging.info("Context:\n%s", context)
 
     try:
-        if len(set(event) - set(BatchArgs()._fields)) == 0:
-            batch_args = BatchArgs(**event)
-        else:
-            batch_args = BatchArgs()
```
mzappitello, line 136: i don't think we need this logic any more.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/postgres_utils.py
```diff
@@ -8,23 +8,41 @@
...
+                DBUsername=db_user,
+                Region=db_region
+            )
+
+            # set the ssl cert path to the file that should be added to the
+            # lambda function at deploy time
+            db_ssl_cert = os.path.join(os.path.abspath(__file),
+                                       ".."
+                                       "aws-cert-bundle.pem")
+
```
mzappitello, line 38: use aws to generate a password if one isn't in the env.

py_gtfs_rt_ingestion/py_gtfs_rt_ingestion/postgres_utils.py
```diff
@@ -8,23 +8,41 @@
...
         conn = psycopg2.connect(
             dbname=db_name,
             password=db_password,
             user=db_user,
             host=db_host,
             port=db_port,
+            sslrootcert=db_ssl_cert)
```
mzappitello, line 45: ssl cert will be provided if using aws rds.

.github/actions/lambda_deploy/action.yaml
rymarczy: So this catches the output of sha256sum and errors if not "OK"?

rymarczy:
LGTM

PR #57, opened 2022-09-14
Update Logging on ECS Container

mzappitello:
* Copy logging utils into performance manager library
* Use ProcessLogger throughout performance manager
* Stop allowing `e` to be a variable name. Use `exception` instead.

Asana Task: https://app.asana.com/0/1202841571879768/1202940849760974/f

### DIFF
--- a/performance_manager/lib/gtfs_static_table.py
+++ b/performance_manager/lib/gtfs_static_table.py
@@ -2,12 +2,12 @@ from typing import List, Optional
 from dataclasses import dataclass
 
 import os
-import logging
 import pandas
 import numpy
 
 import sqlalchemy as sa
 
+from .logging_utils import ProcessLogger
 from .s3_utils import read_parquet, file_list_from_s3
 from .postgres_utils import MetadataLog, get_unprocessed_files, DatabaseManager
 from .postgres_schema import (
@@ -247,11 +247,17 @@ def process_static_tables(db_manager: DatabaseManager) -> None:
     """
     process gtfs static table files from metadataLog table
     """
+    process_logger = ProcessLogger("load_static_tables")
+    process_logger.log_start()
 
     # pull list of objects that need processing from metadata table
     paths_to_load = get_unprocessed_files("FEED_INFO", db_manager.get_session())
+    process_logger.add_metadata(filecount=len(paths_to_load))
 
     for folder, folder_data in paths_to_load.items():
+        individual_logger = ProcessLogger("load_static_table", folder=folder)
+        individual_logger.log_start()
+
         ids = folder_data["ids"]
         folder = str(folder)
 
@@ -261,9 +267,8 @@ def process_static_tables(db_manager: DatabaseManager) -> None:
             load_parquet_files(static_tables, folder)
             transform_data_tables(static_tables)
             insert_data_tables(static_tables, db_manager)
-        except Exception as e:
-            logging.info("Error Processing Static GTFS Schedules")
-            logging.exception(e)
+        except Exception as exception:
+            individual_logger.log_failure(exception)
         else:
             update_md_log = (
                 sa.update(MetadataLog.__table__)
@@ -271,3 +276,6 @@ def process_static_tables(db_manager: DatabaseManager) -> None:
                 .values(processed=1)
             )
             db_manager.execute(update_md_log)
+            individual_logger.log_complete()
+
+    process_logger.log_complete()
--- a/performance_manager/lib/gtfs_static_table.py
+++ b/performance_manager/lib/gtfs_static_table.py
@@ -247,7 +247,7 @@ def process_static_tables(db_manager: DatabaseManager) -> None:
     """
     process gtfs static table files from metadataLog table
     """
-    process_logger = ProcessLogger("load_static_tables")
+    process_logger = ProcessLogger("load_static_table_sets")
     process_logger.log_start()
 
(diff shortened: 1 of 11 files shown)


performance_manager/lib/gtfs_static_table.py
```diff
@@ -247,11 +247,17 @@ def process_static_tables(db_manager: DatabaseManager) -> None:
...
+    process_logger.log_start()
 
     # pull list of objects that need processing from metadata table
     paths_to_load = get_unprocessed_files("FEED_INFO", db_manager.get_session())
+    process_logger.add_metadata(filecount=len(paths_to_load))
 
     for folder, folder_data in paths_to_load.items():
+        individual_logger = ProcessLogger("load_static_table", folder=folder)
+        individual_logger.log_start()
+
```
rymarczy, line 260: This for loop with technically be loading a whole set of static tables. So the `individual_logger` would he loading a whole set of static tables with the same timestamp. And then the `process_logger` would be loading multiple sets of static tables with different timestamps (this won't actually happen in normal operation)

performance_manager/lib/logging_utils.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def __init__(self, process_name: str, **metadata: MdValues) -> None:
+        """
+        create a process logger with a name and optional metadata. a start time
+        and uuid will be created for timing and unique identification
+        """
+        self.process_name = process_name
+        self.uuid = uuid.uuid4()
+        self.start_time = time.time()
+        self.metadata = metadata
```
rymarczy, line 23: Should this ( and ingestion logging) have some hard coded field to indicate it's part of performance manager?

performance_manager/lib/rt_trip_updates.py
```diff
@@ -253,34 +263,54 @@ def merge_trip_update_events(
...
 
+        subprocess_logger = ProcessLogger(
+            "process_tu_dir", file_count=len(paths)
+        )
+        subprocess_logger.log_start()
+
         try:
             new_events = get_tu_dataframe(paths)
-            new_events = unwrap_tu_dataframe(new_events)
+            new_events_unwrapped = unwrap_tu_dataframe(new_events)
```
rymarczy, line 294: keeping both these dataframe variables in scope at the same time could be an issue for memory usage

performance_manager/lib/rt_vehicle_positions.py
```diff
@@ -286,49 +276,57 @@ def merge_vehicle_position_events(
...
 
+    process_logger.add_metadata(paths_to_load=len(paths_to_load))
+
     for folder, folder_data in paths_to_load.items():
         ids = folder_data["ids"]
         paths = folder_data["paths"]
 
+        subprocess_logger = ProcessLogger(
+            "process_tu_dir", folder=folder, file_count=len(paths)
+        )
```
rymarczy, line 304: should be `process_vp_dir`

performance_manager/lib/rt_vehicle_positions.py
```diff
@@ -286,49 +276,57 @@ def merge_vehicle_position_events(
...
-            logging.info(
-                "Size of dataframe with updated dtypes is %d",
-                new_events.shape[0],
-            )
-            new_events = transform_vp_timestamps(new_events)
-            logging.info(
-                "Size of dataframe with transformed timestamps is %d",
-                new_events.shape[0],
+            new_events_updated = transform_vp_dtyes(new_events)
+            new_events_timestamps = transform_vp_timestamps(new_events_updated)
```
rymarczy, line 310: same as with trip updates (but likely not as much of an issue) keeping all of these dataframe variables in scope at the same time could cause memory use issues

rymarczy:
Few comments on this one. I think we might want to try to think of some standardization for the naming of the logger processes so they're a little easier to lookup/debug between splunk and the codebase.  

Right now, in splunk, I think it will be a bit of a chore to group and/or aggregate logs from inside of another larger event loop.

performance_manager/lib/logging_utils.py
mzappitello: i think thats a good idea to do now. i would like both of the libraries to be pulling in shared lamp utils at some point, but we can do that at a later date.

performance_manager/lib/l1_full_trip_events.py
```diff
@@ -231,16 +244,21 @@ def update_with_new_events(
     alternate approach is used here where just hashes are selected that represent
     update records and each column is updated seperatly from insert_dataframe
     """
+    process_logger = ProcessLogger("merge_events")
+    process_logger.log_start()
+
```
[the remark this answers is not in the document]
mzappitello: yeah. i'll update the process names for clarity using vp and tu.

performance_manager/lib/rt_vehicle_positions.py
mzappitello: good catch.

performance_manager/lib/rt_trip_updates.py
mzappitello: ah. that makes a lot of sense. reverting.

performance_manager/lib/rt_vehicle_positions.py
mzappitello: same

mzappitello:
updated.

rymarczy:
LGTM

PR #61, opened 2022-09-19
Add update_on columns to database

rymarczy:
this commit adds functions for update_on triggers in database

sqlalchemy does not support triggers natively, trigger syntax must be passed to the database by executing sqlalchemy DDL objects

Asana Task: (https://app.asana.com/0/1202841571879768/1203007791590472)

### DIFF
--- a/performance_manager/lib/postgres_schema.py
+++ b/performance_manager/lib/postgres_schema.py
@@ -11,7 +11,7 @@ class VehiclePositionEvents(SqlBase):  # pylint: disable=too-few-public-methods
 
     __tablename__ = "eventsVehiclePositions"
 
-    pk_id = sa.Column(sa.Integer, primary_key=True, index=True)
+    pk_id = sa.Column(sa.Integer, primary_key=True)
     is_moving = sa.Column(sa.Boolean)
     stop_sequence = sa.Column(sa.SmallInteger, nullable=True)
     stop_id = sa.Column(sa.String(60), nullable=True)
@@ -25,6 +25,7 @@ class VehiclePositionEvents(SqlBase):  # pylint: disable=too-few-public-methods
     hash = sa.Column(
         sa.LargeBinary(16), nullable=False, index=True, unique=False
     )
+    updated_on = sa.Column(sa.TIMESTAMP, server_default=sa.func.now())
 
 
 class TripUpdateEvents(SqlBase):  # pylint: disable=too-few-public-methods
@@ -32,7 +33,7 @@ class TripUpdateEvents(SqlBase):  # pylint: disable=too-few-public-methods
 
     __tablename__ = "eventsTripUpdates"
 
-    pk_id = sa.Column(sa.Integer, primary_key=True, index=True)
+    pk_id = sa.Column(sa.Integer, primary_key=True)
     is_moving = sa.Column(sa.Boolean)
     stop_sequence = sa.Column(sa.SmallInteger, nullable=True)
     stop_id = sa.Column(sa.String(60), nullable=True)
@@ -45,6 +46,7 @@ class TripUpdateEvents(SqlBase):  # pylint: disable=too-few-public-methods
     hash = sa.Column(
         sa.LargeBinary(16), nullable=False, index=True, unique=False
     )
+    updated_on = sa.Column(sa.TIMESTAMP, server_default=sa.func.now())
 
 
 class MetadataLog(SqlBase):  # pylint: disable=too-few-public-methods
@@ -52,7 +54,7 @@ class MetadataLog(SqlBase):  # pylint: disable=too-few-public-methods
 
     __tablename__ = "metadataLog"
 
-    pk_id = sa.Column(sa.Integer, primary_key=True, index=True)
+    pk_id = sa.Column(sa.Integer, primary_key=True)
     processed = sa.Column(sa.Boolean, default=sa.false())
     path = sa.Column(sa.String(256), nullable=False, unique=True)
     created_on = sa.Column(
@@ -178,6 +180,7 @@ class FullTripEvents(SqlBase):  # pylint: disable=too-few-public-methods
         nullable=True,
         index=True,
     )
+    updated_on = sa.Column(sa.TIMESTAMP, server_default=sa.func.now())
 
 
 class TempFullTripEvents(SqlBase):  # pylint: disable=too-few-public-methods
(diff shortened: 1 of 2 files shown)


performance_manager/lib/postgres_schema.py
```diff
@@ -11,7 +11,7 @@ class VehiclePositionEvents(SqlBase):  # pylint: disable=too-few-public-methods
 
     __tablename__ = "eventsVehiclePositions"
 
-    pk_id = sa.Column(sa.Integer, primary_key=True, index=True)
+    pk_id = sa.Column(sa.Integer, primary_key=True)
```
rymarczy, line 14: Don't think this `index=True` was actually doing anything helpful

performance_manager/lib/postgres_schema.py
```diff
@@ -25,14 +25,15 @@ class VehiclePositionEvents(SqlBase):  # pylint: disable=too-few-public-methods
     hash = sa.Column(
         sa.LargeBinary(16), nullable=False, index=True, unique=False
     )
+    updated_on = sa.Column(sa.TIMESTAMP, server_default=sa.func.now())
```
rymarczy, line 28: add `update_on` column to schema, default value handles setting value on INSERT, trigger sets value on UPDATE

performance_manager/lib/postgres_schema.py
```diff
@@ -45,14 +46,15 @@ class TripUpdateEvents(SqlBase):  # pylint: disable=too-few-public-methods
...
 
 
 class MetadataLog(SqlBase):  # pylint: disable=too-few-public-methods
     """Table for keeping track of parquet files in S3"""
 
     __tablename__ = "metadataLog"
 
-    pk_id = sa.Column(sa.Integer, primary_key=True, index=True)
+    pk_id = sa.Column(sa.Integer, primary_key=True)
```
rymarczy, line 57: not helpful

performance_manager/lib/postgres_utils.py
```diff
@@ -16,6 +16,69 @@
...
+    update_column_name = "updated_on"
+    trigger_function_statement = sa.DDL(
+        f"CREATE OR REPLACE FUNCTION update_modified_columns() "
+        f"RETURNS TRIGGER AS $$ "
+        f"BEGIN "
+        f"NEW.{update_column_name} = CURRENT_TIMESTAMP; "
+        f"RETURN NEW; "
+        f"END; $$ LANGUAGE plpgsql"
+    )
+    connection.execute(trigger_function_statement)
```
rymarczy, line 40: postgres requires a trigger function to perform a trigger action, this trigger function will update the `updated_on` column with CURRENT_TIMESTAMP for all records provided

performance_manager/lib/postgres_utils.py
```diff
@@ -16,6 +16,69 @@
...
+    if update_column_name in list(table.columns.keys()):
+        trigger_name = f"update_{table}_modified".lower()
+        trigger_statement = sa.DDL(
+            f"DO $$ BEGIN "
+            f"IF NOT EXISTS (SELECT 1 FROM pg_trigger WHERE tgname = '{trigger_name}') "
+            f"THEN "
+            f'    CREATE TRIGGER {trigger_name} BEFORE UPDATE ON "{table}" '
+            f"    FOR EACH ROW EXECUTE PROCEDURE update_modified_columns();"
+            f"END IF; "
+            f"END $$;"
```
rymarczy, line 50: Builds trigger creation text with table and trigger names, BEFORE UPDATE used to prevent infinite update loop

performance_manager/lib/postgres_utils.py
```diff
@@ -16,6 +16,69 @@
...
+    if update_column_name in list(table.columns.keys()):
+        trigger_name = f"update_{table}_modified".lower()
+        primary_key = str(table.primary_key.columns.items()[0][0])
+        trigger_statement = sa.DDL(
+            f"CREATE TRIGGER IF NOT EXISTS {trigger_name} "
+            f"AFTER UPDATE ON {table} FOR EACH ROW "
+            f"BEGIN "
+            f"  UPDATE {table} SET {update_column_name} = CURRENT_TIMESTAMP "
+            f"  WHERE {primary_key} = NEW.{primary_key};"
+            f"END; "
```
rymarczy, line 77: sqlite TRIGGER text requires primary key column to know what records to update

performance_manager/lib/postgres_utils.py
```diff
@@ -147,13 +210,24 @@ def __init__(
...
             # https://docs.sqlalchemy.org/en/14/dialects/sqlite.html#foreign-key-support
             @sa.event.listens_for(sa.engine.Engine, "connect")
-            def set_sqlite_pragma(dbapi_connection, _):  # type: ignore
-                cursor = dbapi_connection.cursor()
-                cursor.execute("PRAGMA foreign_keys=ON")
-                cursor.close()
+            def set_sqlite_pragma(
+                connection: sa.engine.Connection, _: Any
+            ) -> None:
+                connection.execute("PRAGMA foreign_keys=ON")
```
rymarczy, line 216: got rid of `#type: ignore`

performance_manager/lib/postgres_utils.py
```diff
@@ -147,13 +210,24 @@ def __init__(
...
+
+            sa.event.listen(
+                sa.schema.Table,
+                "after_create",
+                sqlite_create_modified_trigger,
+            )
```
rymarczy, line 222: sqlalchemy event that calls trigger creation function, will be called after any table creation event, passing in the table that was created as an argument

mzappitello: if we're connecting to a table thats already been created, does that mean we won't be applying this function?

rymarczy: Correct, it's only called during table creation

mzappitello: cool. we'll probably need to create a util that can drop the tables off of the RDS so that we can reinstate them from the ECS when we merge this.

rymarczy: I think we should look into using something like alembic https://pypi.org/project/alembic/ to handle schema migrations. I originally assumed this type of functionality was built into sqlalchemy, but it looks like they developed another tool instead.

mzappitello:
LGTM.

PR #65, opened 2022-09-22
add batching to trip update processing

rymarczy:
processing of trip update parquet files was resulting in very high memory usage. this commit adds batching to the processing of trip update parquet files to limit memory usage.

`read_parquet_chunks` function added to `s3_utils.py`. this function defaults to reading ~100,000 rows from a parquet at a time.

Asana Task: https://app.asana.com/0/1202841571879768/1203021761496491

### DIFF
--- a/performance_manager/lib/rt_trip_updates.py
+++ b/performance_manager/lib/rt_trip_updates.py
@@ -36,9 +36,13 @@ def get_tu_dataframe_chunks(
         ("start_time", "!=", "None"),
         ("vehicle_id", "!=", "None"),
     ]
-
+    # 100_000 batch size should result in ~5-6 GB of memory use per batch
+    # of trip update records
     return read_parquet_chunks(
-        to_load, columns=trip_update_columns, filters=trip_update_filters
+        to_load,
+        max_rows=100_000,
+        columns=trip_update_columns,
+        filters=trip_update_filters,
     )
 
 
@@ -103,6 +107,10 @@ def get_and_unwrap_tu_dataframe(
     predicted trip update stop events
     """
     events = pandas.Series(dtype="object")
+    # get_tu_dataframe_chunks set to pull ~100_000 trip update records
+    # per batch, this should result in ~5-6 GB of memory use per batch
+    # after batch goes through explod_stop_time_update vectorize operation,
+    # resulting Series has negligible memory use
     for batch_events in get_tu_dataframe_chunks(paths):
         # store start_date as int64
         batch_events["start_date"] = pandas.to_numeric(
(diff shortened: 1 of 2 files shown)


performance_manager/lib/rt_trip_updates.py
```diff
@@ -91,43 +93,50 @@ def explode_stop_time_update(
...
-            events.stop_time_update,
-            events.timestamp,
-            events.direction_id,
-            events.route_id,
-            events.start_date,
-            events.start_time,
-            events.vehicle_id,
+    events = pandas.Series(dtype="object")
+    for batch_events in get_tu_dataframe_chunks(paths):
+        # store start_date as int64
```
rymarczy, line 107: Add loop for processing trip update dataframe batches. 

100_000 default record limit for dataframe batching results in around 5-6GB of memory usage per batch

performance_manager/lib/s3_utils.py
```diff
@@ -5,30 +5,72 @@
...
+) -> pandas.core.frame.DataFrame:
+    """
+    read parquet file or files from s3 and return it as a pandas dataframe
+    """
     return (
-        pq.ParquetDataset(to_load, filesystem=active_fs, filters=filters)
-        .read_pandas()
-        .to_pandas()
+        _get_pyarrow_table(filename, filters)
+        .to_pandas(self_destruct=True)
```
mzappitello, line 41: self destruct deletes the pyarrow table?

performance_manager/lib/s3_utils.py
```diff
@@ -5,30 +5,72 @@
...
+    read parquet file or files from s3 IN CHUNKS
+    return chunks as pandas dataframes
+
+    chunk size attempts to be close to max_rows parameter, but may sometimes
+    overshoot because of chunk layout of pyarrow table
+    """
+    yield_dataframe = pandas.DataFrame()
+    for batch in _get_pyarrow_table(filename, filters).to_batches(
+        max_chunksize=None
+    ):
```
mzappitello, line 62: interesting to me that 
```
    return pq.ParquetDataset(
        to_load, filesystem=active_fs, filters=filters
    ).read_pandas()
```
doesn't load the whole thing.

performance_manager/lib/rt_trip_updates.py
mzappitello: might be a good comment on this for loop.

performance_manager/lib/s3_utils.py
rymarczy: I'm not totally sure if it actually loads the pyarrow table at that point or not, it might lazy load the table/schema and wait for the `'to_pandas()` method to actually download.

Even if it does download the pyarrow table on that call, the memory usage is a fraction of converting the table to a dataframe.

performance_manager/lib/s3_utils.py
rymarczy: It's supposed to, but it's marked as an experimental feature. I didn't really notice any significant memory savings from its use, but I figured it couldn't hurt.

performance_manager/lib/rt_trip_updates.py
rymarczy: will incorporate

mzappitello:
i tested this out locally and it seemed to do the trick.

PR #67, opened 2022-10-04
match gtfs static data to event records

rymarczy:
to facilitate metrics implementation, event records must be matched to gtfs static data.

this PR updates the processing of VEHICLE_POSITION and TRIP_UPDATE parquet files, to add matching of records to gtfs static schedule data.

this matching is accomplished by the addition of a column to the `eventsVehiclePositions` and `eventsTripUpdates` tables. the column is a foreign key that references a unique timestamp from the `staticFeedInfo` table.

using this foreign key, event records can be matched to their relevant gtfs static data for additional database join activity.

this PR also updates event processing by grouping event records by `parent_station` from gtfs static data, instead of `stop_id` from specific event records.

Asana Task: https://app.asana.com/0/1202939398067300/1203091943545605

### DIFF
--- a/performance_manager/lib/gtfs_utils.py
+++ b/performance_manager/lib/gtfs_utils.py
@@ -10,7 +10,7 @@ def add_event_hash_column(
     expected_hash_columns: Sequence[str] = (
         "is_moving",
         "stop_sequence",
-        "stop_id",
+        "parent_station",
         "direction_id",
         "route_id",
         "start_date",
--- a/performance_manager/lib/l1_full_trip_events.py
+++ b/performance_manager/lib/l1_full_trip_events.py
@@ -9,6 +9,7 @@ from .postgres_schema import (
     TripUpdateEvents,
     FullTripEvents,
     TempFullTripEvents,
+    StaticStops,
 )
 from .gtfs_utils import add_event_hash_column
 
@@ -20,7 +21,7 @@ def pull_and_transform(db_manager: DatabaseManager) -> pandas.DataFrame:
     # columns to hash for FullTripEvents primary key field
     expected_hash_columns = [
         "stop_sequence",
-        "stop_id",
+        "parent_station",
         "direction_id",
         "route_id",
         "start_date",
@@ -56,7 +57,13 @@ def pull_and_transform(db_manager: DatabaseManager) -> pandas.DataFrame:
         sa.select(
             VehiclePositionEvents.pk_id.label("fk_vp_moving_event"),
             VehiclePositionEvents.stop_sequence,
-            VehiclePositionEvents.stop_id,
+            sa.case(
+                (
+                    StaticStops.parent_station.is_(None),
+                    VehiclePositionEvents.stop_id,
+                ),
+                else_=StaticStops.parent_station,
+            ).label("parent_station"),
             VehiclePositionEvents.direction_id,
             VehiclePositionEvents.route_id,
             VehiclePositionEvents.start_date,
@@ -71,6 +78,15 @@ def pull_and_transform(db_manager: DatabaseManager) -> pandas.DataFrame:
         .join(
             dupe_hash_cte,
             VehiclePositionEvents.hash == dupe_hash_cte.c.hash,
+            isouter=True,
+        )
+        .join(
+            StaticStops,
+            sa.and_(
+                VehiclePositionEvents.fk_static_timestamp
+                == StaticStops.timestamp,
+                VehiclePositionEvents.stop_id == StaticStops.stop_id,
+            ),
             isouter=True,
         )
         .where(
@@ -92,7 +108,13 @@ def pull_and_transform(db_manager: DatabaseManager) -> pandas.DataFrame:
         sa.select(
             VehiclePositionEvents.pk_id.label("fk_vp_stopped_event"),
             VehiclePositionEvents.stop_sequence,
-            VehiclePositionEvents.stop_id,
+            sa.case(
+                (
+                    StaticStops.parent_station.is_(None),
+                    VehiclePositionEvents.stop_id,
+                ),
+                else_=StaticStops.parent_station,
+            ).label("parent_station"),
             VehiclePositionEvents.direction_id,
             VehiclePositionEvents.route_id,
             VehiclePositionEvents.start_date,
@@ -107,6 +129,15 @@ def pull_and_transform(db_manager: DatabaseManager) -> pandas.DataFrame:
         .join(
             dupe_hash_cte,
             VehiclePositionEvents.hash == dupe_hash_cte.c.hash,
+            isouter=True,
+        )
+        .join(
+            StaticStops,
+            sa.and_(
+                VehiclePositionEvents.fk_static_timestamp
+                == StaticStops.timestamp,
+                VehiclePositionEvents.stop_id == StaticStops.stop_id,
+            ),
             isouter=True,
         )
         .where(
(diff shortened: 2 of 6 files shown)


performance_manager/lib/gtfs_utils.py
```diff
@@ -10,7 +10,7 @@ def add_event_hash_column(
     expected_hash_columns: Sequence[str] = (
         "is_moving",
         "stop_sequence",
-        "stop_id",
+        "parent_station",
```
rymarczy, line 13: default hash on `parent_station` instead of `stop_id` 

this is because some parent stations correlate to multiple stop id's

performance_manager/lib/l1_full_trip_events.py
```diff
@@ -56,7 +57,13 @@ def pull_and_transform(db_manager: DatabaseManager) -> pandas.DataFrame:
...
             VehiclePositionEvents.pk_id.label("fk_vp_moving_event"),
             VehiclePositionEvents.stop_sequence,
-            VehiclePositionEvents.stop_id,
+            sa.case(
+                (
+                    StaticStops.parent_station.is_(None),
+                    VehiclePositionEvents.stop_id,
+                ),
+                else_=StaticStops.parent_station,
+            ).label("parent_station"),
```
rymarczy, line 66: replace `stop_id` select with case statement

not all stop id's have a parent station, so if not parent station is found, use stop_id

performance_manager/lib/l1_full_trip_events.py
```diff
@@ -163,8 +208,8 @@ def pull_and_transform(db_manager: DatabaseManager) -> pandas.DataFrame:
                 stop_vp_cte.c.stop_sequence,
             ).label("stop_sequence"),
             sa.func.coalesce(
-                move_vp_cte.c.stop_id, stop_vp_cte.c.stop_id
-            ).label("stop_id"),
+                move_vp_cte.c.parent_station, stop_vp_cte.c.parent_station
+            ).label("parent_station"),
```
rymarczy, line 212: replace `stop_id` with `parent_station`

performance_manager/lib/postgres_schema.py
```diff
@@ -25,6 +25,11 @@ class VehiclePositionEvents(SqlBase):  # pylint: disable=too-few-public-methods
     hash = sa.Column(
         sa.LargeBinary(16), nullable=False, index=True, unique=False
     )
+    fk_static_timestamp = sa.Column(
+        sa.Integer,
+        sa.ForeignKey("staticFeedInfo.timestamp"),
+        nullable=False,
+    )
```
rymarczy, line 32: new column to store static data timestamp for quick joining to gtfs static data

performance_manager/lib/postgres_schema.py
```diff
@@ -46,6 +51,11 @@ class TripUpdateEvents(SqlBase):  # pylint: disable=too-few-public-methods
...
```
rymarczy, line 58: new column to store static data timestamp for quick joining to gtfs static data

performance_manager/lib/rt_trip_updates.py
```diff
@@ -155,18 +160,115 @@ def get_and_unwrap_tu_dataframe(
     events["is_moving"] = False
     events["pk_id"] = None
 
+    return events
+
+
+def join_gtfs_static(
+    trip_updates: pandas.DataFrame, db_manager: DatabaseManager
+) -> pandas.DataFrame:
```
rymarczy, line 168: new function to get static timestamp value for joining event to gtfs static data

this function is pretty well commented explaining business logic

performance_manager/lib/rt_trip_updates.py
```diff
@@ -155,18 +160,115 @@ def get_and_unwrap_tu_dataframe(
...
+
+    adds "fk_static_timestamp" and "parent_station" columns to dataframe
+    """
+    # get unique "start_date" values from trip update dataframe
+    # with associated minimum "timestamp_start"
+    date_groups = trip_updates.groupby(by="start_date")["timestamp_start"].min()
+
+    # pylint: disable=duplicate-code
+    # TODO(ryan): the following code is duplicated in rt_vehicle_positions.py
+
```
rymarczy, line 180: might be able to combine this into a library file, business logic between vehicle positions and trip updates is close, but not exact.

performance_manager/lib/rt_vehicle_positions.py
```diff
@@ -70,6 +75,98 @@ def transform_vp_dtyes(vehicle_positions: pandas.DataFrame) -> pandas.DataFrame:
...
+
+
+def join_gtfs_static(
+    vehicle_positions: pandas.DataFrame, db_manager: DatabaseManager
+) -> pandas.DataFrame:
+    """
+    join vehicle position dataframe to gtfs static records
+
+    adds "fk_static_timestamp" and "parent_station" columns to dataframe
+    """
```
rymarczy, line 88: function is very close to trip update function, could potentially be combined in utils file

performance_manager/tests/july_17_filepaths.json
```diff
@@ -269,6 +269,15 @@
...
     },
+    {
+        "path": "mbta-ctd-dataplatform-dev-springboard/lamp/FEED_INFO/timestamp=1657216900/a56ca397ee4d4da89d22d52dfbc5406c-0.parquet"
+    },
+    {
+        "path": "mbta-ctd-dataplatform-dev-springboard/lamp/FEED_INFO/timestamp=1659130909/e931dca261054e68b9259ff23da860cf-0.parquet"
+    },
+    {
+        "path": "mbta-ctd-dataplatform-dev-springboard/lamp/FEED_INFO/timestamp=1660175804/55598556e865435a8019d329c7d6f2f1-0.parquet"
+    },
```
rymarczy, line 280: add some additional gtfs static file paths for testing

performance_manager/lib/rt_trip_updates.py
```diff
@@ -155,18 +160,115 @@ def get_and_unwrap_tu_dataframe(
...
+    # dictionary used to match minimum "timestamp_start" values to
+    # "timestamp" from StaticFeedInfo table, to be used as foreign key
+    timestamp_lookup = {}
+    for (date, min_timestamp) in date_groups.iteritems():
+        date = int(date)
+        min_timestamp = int(min_timestamp)
```
mzappitello, line 186: why the type conversion? i'm guessing they are stored as something else and `sa` needs them to be ints?

performance_manager/lib/rt_trip_updates.py
```diff
@@ -155,18 +160,115 @@ def get_and_unwrap_tu_dataframe(
...
+        # "start_date" from trip updates must be between "feed_start_date"
+        # and "feed_end_date" in StaticFeedInfo
+        # minimum "timestamp_start" must also be less than "timestamp" from
+        # StaticFeedInfo, order by "timestamp" descending and limit to 1 result
+        # this should deal with multiple static schedules with possible
+        # overlapping times of applicability
```
mzappitello, line 192: is there an ordering that we need to do to ensure the selection at 1 is the one we want?

mzappitello:
this LGTM. I had a few questions that were mostly for my own benefit.

performance_manager/lib/rt_trip_updates.py
rymarczy: These values are stored as numpy `Int64` dtypes in the dataframe, which is a dtype that sqlalchemy does not seem to be a fan of. Sqlalchemy throws an error in the select statement below without the conversion to the Python `int` dtype

performance_manager/lib/rt_trip_updates.py
rymarczy: Yes, the ordering is done in the `sa.select` statement with `.order_by(StaticFeedInfo.timestamp.desc())` 

There is a bit of a race condition here where we need to have processed the matching GTFS Static data before any RT data processing.

PR #68, opened 2022-10-07
add l2 headways processing

rymarczy:
this PR adds processing of l2 headways table to performance manager

insert and update logic is accomplished through all SQL calls

steps involved:
1. pull minimum timestamp_start value for records needing headways processing
2. load headways values (with 6 hours of overlapping records) into temporary headways table
3. update existing headways from temporary table
4. insert new headwasy from temporary table

Asana Task: https://app.asana.com/0/1202939398067300/1203123341195650

### DIFF
--- a/performance_manager/lib/gtfs_static_table.py
+++ b/performance_manager/lib/gtfs_static_table.py
@@ -232,6 +232,11 @@ def transform_data_tables(static_tables: List[StaticTableDetails]) -> None:
                     .astype("Int64")
                 )
 
+        table.data_table = table.data_table.fillna(numpy.nan).replace(
+            [numpy.nan], [None]
+        )
+        table.data_table = table.data_table.replace([""], [None])
+
 
 def insert_data_tables(
     static_tables: List[StaticTableDetails],
--- a/performance_manager/lib/l2_headways.py
+++ b/performance_manager/lib/l2_headways.py
@@ -1,0 +1,383 @@
+from typing import Optional
+import sqlalchemy as sa
+
+from .logging_utils import ProcessLogger
+from .postgres_utils import DatabaseManager
+from .postgres_schema import (
+    VehiclePositionEvents,
+    FullTripEvents,
+    TripUpdateEvents,
+    StaticStops,
+    Headways,
+    TempHeadways,
+)
+
+
+def get_min_timestamp(db_manager: DatabaseManager) -> Optional[int]:
+    """
+    find the minimum timestamp_start value for processing new headways
+    """
+    # query for trip update timestamp_start values
+    tu_timestamp_query = sa.select(
+        sa.func.min(TripUpdateEvents.timestamp_start).label("timestamp_start"),
+    ).join(
+        FullTripEvents,
+        FullTripEvents.fk_tu_stopped_event == TripUpdateEvents.pk_id,
+    )
+
+    # query for vehicle position timestamp_start values
+    vp_timestamp_query = sa.select(
+        sa.func.min(VehiclePositionEvents.timestamp_start).label(
+            "timestamp_start"
+        ),
+    ).join(
+        FullTripEvents,
+        FullTripEvents.fk_vp_stopped_event == VehiclePositionEvents.pk_id,
+    )
+
+    # get maximum updated_on timestamp from headways table
+    max_headway_update_query = sa.select(sa.func.max(Headways.updated_on))
+    # mypy error: error: "BaseCursorResult" has no attribute "fetchone"
+    max_update_dt = db_manager.execute(max_headway_update_query).fetchone()[0]  # type: ignore
+
+    # will only be None if no records exist in Headways table
+    # if records do exist, add constraing to vehicle position and trip update
+    # timestamp_start queries
+    if max_update_dt is not None:
+        tu_timestamp_query = tu_timestamp_query.where(
+            FullTripEvents.updated_on > max_update_dt
+        )
+        vp_timestamp_query = vp_timestamp_query.where(
+            FullTripEvents.updated_on > max_update_dt
+        )
+
+    # union query so all results are combined
+    min_timestamp_start_cte = tu_timestamp_query.union(vp_timestamp_query).cte(
+        name="min_timestamp"
+    )
+
+    # get minimum timestamp_start value for records needing to be added to
+    # headways table
+    min_timestamp_start_query = sa.select(
+        sa.func.min(min_timestamp_start_cte.c.timestamp_start)
+    )
+
+    # this will return a valid timestamp_start integer or None, if no records
+    # need to be added to headways table
+    # mypy error: error: "BaseCursorResult" has no attribute "fetchone"
+    min_timestamp_start = db_manager.execute(min_timestamp_start_query).fetchone()[0]  # type: ignore
+
+    # pull overlapping timestamp start values going back 6 Hours
+    # this effecitvely sets a maxium headway value of 6 hours
+    # it is assumed that headways over 6 hours are not really headways and do not
+    # need to be recorded
+    if isinstance(min_timestamp_start, int):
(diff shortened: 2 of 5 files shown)


performance_manager/lib/gtfs_static_table.py
```diff
@@ -232,6 +232,11 @@ def transform_data_tables(static_tables: List[StaticTableDetails]) -> None:
                     .astype("Int64")
                 )
 
+        table.data_table = table.data_table.fillna(numpy.nan).replace(
```
mzappitello, line 235: whats this about?

performance_manager/lib/postgres_schema.py
```diff
@@ -108,13 +108,13 @@ class StaticRoutes(SqlBase):  # pylint: disable=too-few-public-methods
     pk_id = sa.Column(sa.Integer, primary_key=True)
     route_id = sa.Column(sa.String(60), nullable=False)
     agency_id = sa.Column(sa.SmallInteger, nullable=False)
-    route_short_name = sa.Column(sa.String(60), nullable=False)
-    route_long_name = sa.Column(sa.String(150), nullable=False)
-    route_desc = sa.Column(sa.String(40), nullable=False)
+    route_short_name = sa.Column(sa.String(60), nullable=True)
```
mzappitello, line 111: why are all of these changing?

performance_manager/lib/l2_headways.py
```diff
@@ -0,0 +1,383 @@
...
+                FullTripEvents.fk_tu_stopped_event.isnot(None),
+                FullTripEvents.fk_vp_stopped_event.is_(None),
+                TripUpdateEvents.timestamp_start > min_timestamp_start,
+            )
+        )
+    )
+
+    # merge stop event results from vehcile positions and trip updates into
+    # one query all records should be unique
+    stop_events_cte = vp_query.union_all(tu_query).cte(name="all_stop_events")
```
mzappitello, line 188: in an ideal world, do we expect the length of this cte to be the same size as the vp_query and tu_query?

performance_manager/lib/l2_headways.py
```diff
@@ -0,0 +1,383 @@
...
+
+    # headways will be aggregated by the following fields:
+    # - direction_id
+    # - route_id
+    # - start_date
+    # - parent_station
+    #
+    # query for stop event records from vehcile positions that require headways
+    # processing
+    vp_query = (
```
mzappitello, line 106: why are these queries rather than CTEs?

performance_manager/lib/l2_headways.py
```diff
@@ -0,0 +1,383 @@
...
+    # merge stop event results from vehcile positions and trip updates into
+    # one query all records should be unique
+    stop_events_cte = vp_query.union_all(tu_query).cte(name="all_stop_events")
+
+    # this subquery performs lag operation on vehicle_id and timestamp_start
+    # fields, these lag operations create the fields of prev_vehicle_id and
+    # prev_timestamp_start for every record
+    # (except first record of partition groups)
+    filter_query = sa.select(
+        stop_events_cte.c.fk_vp_stopped_event,
```
mzappitello, line 195: what does the `c` stand for here?

performance_manager/lib/l2_headways.py
```diff
@@ -0,0 +1,383 @@
...
+    """
+    process_logger = ProcessLogger("process_l2_headways")
+    process_logger.log_start()
+    try:
+        temp_table_rows = load_temp_headways(db_manager)
+
+        # gate to check for no records needing update / insert
+        if temp_table_rows == 0:
+            process_logger.log_complete()
+            return
```
mzappitello, line 375: this seems like a funny way to do this, but i follow it. i would have guessed there was a bool on result for it.

performance_manager/lib/l2_headways.py
```diff
@@ -0,0 +1,383 @@
...
+    VehiclePositionEvents,
+    FullTripEvents,
+    TripUpdateEvents,
+    StaticStops,
+    Headways,
+    TempHeadways,
+)
+
+
+def get_min_timestamp(db_manager: DatabaseManager) -> Optional[int]:
```
mzappitello, line 16: this is another one of these functions?

performance_manager/lib/l2_headways.py
```diff
@@ -0,0 +1,383 @@
...
+                TempHeadways.fk_tu_stopped_event,
+                TempHeadways.fk_trip_event_hash,
+                TempHeadways.headway_seconds,
+            )
+            .join(
+                Headways,
+                Headways.fk_trip_event_hash == TempHeadways.fk_trip_event_hash,
+                isouter=True,
+            )
+            .where((Headways.fk_trip_event_hash.is_(None))),
```
mzappitello, line 352: how does this part work? i would think that the fk_trip_event_hash in temp wouldn't be in headways yet.

performance_manager/lib/gtfs_static_table.py
rymarczy: Dataframes can have NULL values represented in a few slightly different ways, but the only one compatible with writing to the RDS with sqlalchemy is the Python `None` dtype. 

So this line turns any na/NULL values into a python `None` in the dataframe:
`table.data_table = table.data_table.fillna(numpy.nan).replace(
            [numpy.nan], [None]
        )`

Then with the GTFS static data I also found an issue where values that are NULL / blank are loaded into the dataframe as an empty string "". This was breaking some queries where we filter on NULL in the RDS.

So this line replaces any empty strings with Python `None` dtypes, that will turn into NULL in the RDS:
`table.data_table = table.data_table.replace([""], [None])`

performance_manager/lib/postgres_schema.py
rymarczy: This was an offshoot of the previous comment where values I thought didn't have any NULL records, actually did because empty strings were getting loaded.

performance_manager/lib/l2_headways.py
rymarczy: Yes, in an ideal world every vehicle position stop event would have a corresponding trip update stop event. But then we also wouldn't need to use the trip update data for this purpose.

performance_manager/lib/l2_headways.py
rymarczy: CTE's are used when you'll reference the result of a subquery later in the your SQL statement. 

In this instance, these two queries are joined with a `UNION ALL`. The joined results are used as a CTE because they are used later in the query.

performance_manager/lib/l2_headways.py
rymarczy: `c` is column, so these are column names from the CTE

performance_manager/lib/l2_headways.py
rymarczy: Yeah, since we aren't doing anything with the `int` value being returned (other than checking if it's 0 or not) a bool makes more sense, I'll make that change

performance_manager/lib/l2_headways.py
rymarczy: You are correct.

This query does a JOIN between `TempHeadways` and `Headways`, specifically a LEFT JOIN where `TempHeadways` is the left table, so all records from `TempHeadways` are included in the JOIN. Then in the WHERE clause we limit the results to records where `fk_trip_event_hash` from the `Headways` table is NULL:

`.where((Headways.fk_trip_event_hash.is_(None)))`

This gives us only the records from `TempHeadways` that don't exist in `Headways`

performance_manager/lib/l2_headways.py
rymarczy: This one we actually need the `int` value being returned. Because it is used in `vp_query` and `tu_query`

mzappitello:
LGTM

PR #69, opened 2022-10-24
Lambda to ECS Ingestion Migration

mzappitello:
Asana Task: https://app.asana.com/0/0/1203049724800739/f
Asana Task: https://app.asana.com/0/0/1203049724800740/f

After numerous issues, we've decided to move our ingestion process from
lambda to an ECS service. This should make it easier to connect to our
RDS and keep our springboard filecount down, as well as model when
running everything in docker locally.

* Create dockerfile to describe the ingestion image.
* remove batch_files dot py
* Completely change ingest dot py to do both the batching and the
  ingestion, using the (newly added) `schedule` module to schedule the
  task to run five minutes past the hour.
* Simplify `batcher.py`, removing `compress_filenames`,
  `unpack_filenames`, `create_event`, `trigger_lambda` and
  `event_size_over_limit` functions.
* Add in (unimplimented) `LIGHT_RAIL` Config Type.
* Cleanup tests that no longer make sense.
* Update Docker File Structure so it makes sense with one RDS and two Docker Files.


py_gtfs_rt_ingestion/lib/batcher.py
```diff
@@ -0,0 +1,97 @@
+import logging
```
mzappitello, line 1: not sure why github didn't pickup the filepath rename on this, its largely the same, but a touch smaller as we're no longer compressing / uncompressing filepaths, creating event json strings, nor triggering ingestion lambdas as everything is being handled inside of one instance now.

mzappitello (commented):
lots going on here. i tried to keep the commit history clean, so looking at the changes one commit at a time might make the most sense.

docker-compose.yml
```diff
@@ -18,10 +18,22 @@ services:
...
       dockerfile: Dockerfile
     depends_on:
-      - local_rds 
+      - local_rds
     working_dir: /performance_manager
     volumes:
       - ~/.aws:/root/.aws:ro # map credentials to be used by boto3, read-only
+
+  performance_manager:
+    container_name: ingestion
```
rymarczy, line 30: need to rename `performance_manager`

docker-compose.yml
```diff
@@ -18,10 +18,22 @@ services:
     container_name: performance_manager
```
rymarczy, line 18: Should probably update this image to match our AWS RDS version:

`image: postgres:13.4`

py_gtfs_rt_ingestion/ingest.py
```diff
@@ -88,113 +83,87 @@ def validate_environment() -> None:
...
-        main(event, process_logger)
-        process_logger.log_complete()
-    except Exception as exception:
-        process_logger.log_failure(exception)
+
+def main() -> None:
+    """every second run jobs that are currently pending"""
+    while True:
+        schedule.run_pending()
+        time.sleep(1)
```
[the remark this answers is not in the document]
rymarczy: what a lazy library, needing to sleep  for 1 second

rymarczy:
With this new approach, it seems like we have to implement a new batching strategy. Previously we were limited by processing time and message size of sending file lists between the lambdas. Without that limitation, and using `batch_files(file_list, 1_000_000_000)` we could conceivable create a parquet file in memory consisting of hundreds of thousands gzip files, which I think would crash the ECS. 

Ideally these files would be batched by the `header``timestamp` value from inside of the gzip file. This would align the batches with the pyarrow partitioning and hopefully guarantee that the ECS is never holding more than 1 hours with of a parquet file in memory and that each partition folder on S3 only has one file. 

Might want to discuss this over a quick all today.

py_gtfs_rt_ingestion/lib/ingest.py
```diff
@@ -0,0 +1,85 @@
...
+
+    for file in files:
+        try:
+            config_type = ConfigType.from_filename(file)
+            converter = converters[config_type]
+            save_file = converter.add_file(file)
+
+            if save_file:
+                yield converter
+                converter.reset()
```
rymarczy, line 77: Does this workflow rely on the s3 paths being provided by `file_list_from_s3` being provided in chronological order?

py_gtfs_rt_ingestion/lib/ingest.py
```diff
@@ -0,0 +1,85 @@
...
+
+        except ConfigTypeFromFilenameException as config_exception:
+            logging.warning(config_exception)
+            converters[ConfigType.ERROR].add_file(file)
+            continue
+
+    for _, converter in converters.items():
+        yield converter
```
rymarczy, line 85: could use `.values()` instead of `.items()` since we are ignoring the key here

py_gtfs_rt_ingestion/dev_test_setup.py
```diff
@@ -185,9 +185,9 @@ def copy_obj(prefix: str, num_to_copy: int) -> int:
         src_uri_root,
     )
     try:
-        for (uri, size) in file_list_from_s3(GTFS_BUCKET, prefix):
+        for uri in file_list_from_s3(GTFS_BUCKET, prefix):
             # Skip busloc TripUpdates because of S3 permission issues
-            if size > 0 and skip_uri not in uri:
```
mzappitello, line 191: no longer need size since we're batching by the hour.

py_gtfs_rt_ingestion/ingest.py
```diff
@@ -95,23 +94,17 @@ def batch_and_ingest() -> None:
...
-        files = batch.filenames
+    for converter in ingest_files(files):
         archive_files = []
         error_files = []
 
         try:
-            config_type = batch.config_type
-            converter = get_converter(config_type)
-
-            for s3_prefix, table in converter.convert(files):
```
mzappitello: this happens in the ingest files now.

py_gtfs_rt_ingestion/ingest.py
```diff
@@ -157,6 +150,7 @@ def batch_and_ingest() -> None:
 
 def main() -> None:
     """every second run jobs that are currently pending"""
+    batch_and_ingest()
```
mzappitello, line 153: kick off a batch and ingest and then schedule one to run 5 minutes past the hour every hour.

py_gtfs_rt_ingestion/lib/batcher.py
```diff
@@ -1,97 +0,0 @@
-import logging
```
mzappitello: :wave

py_gtfs_rt_ingestion/lib/convert_gtfs.py
```diff
@@ -1,24 +1,24 @@
...
+def zip_to_pyarrow(filename: str) -> List[Tuple[str, pyarrow.Table]]:
     """
     convert a schedule gtfs zip file into tables. the zip file is essentially a
     small database with each contained file (outside of feed info) acting as its
     own table. info on the gtfs scheduling standard can be found at
     http://gtfs.org/schedule/
     """
     process_logger = ProcessLogger("convert_single_gtfs", filename=filename)
-    process_logger.log_start()
+    # process_logger.log_start()
```
rymarczy, line 21: does this need to be un-commented?

py_gtfs_rt_ingestion/lib/convert_gtfs.py
```diff
@@ -53,22 +53,22 @@ class GtfsConverter(Converter):
     Converter for GTFS Schedule Data
     """
 
-    def convert(self, files: list[str]) -> list[tuple[str, pyarrow.Table]]:
-        process_logger = ProcessLogger("convert_gtfs", filecount=len(files))
-        process_logger.log_start()
+    def __init__(self, config_type: ConfigType) -> None:
```
mzappitello, line 56: new API for Converter requires a slight change here.

mzappitello: [link](https://github.com/mbta/lamp/pull/69/commits/c427093a41ccaa8dba33659a5dd37e7b3272fff9#diff-ed3e7f70fabd180d3da9a1de30941921944bfe3957f9b9f25c825158ee031055R82-R118) to api change.

py_gtfs_rt_ingestion/lib/converter.py
```diff
@@ -69,6 +75,7 @@ def is_gtfs_rt(self) -> bool:
             self.BUS_TRIP_UPDATES,
             self.BUS_VEHICLE_POSITIONS,
             self.VEHICLE_COUNT,
+            self.LIGHT_RAIL,
```
rymarczy, line 78: This might not be GTFS data, a comment or something about this not being an implemented ConfigType and needing further review may be helpful.

py_gtfs_rt_ingestion/lib/s3_utils.py
```diff
@@ -57,8 +55,7 @@ def file_list_from_s3(
         if page["KeyCount"] == 0:
             continue
         for obj in page["Contents"]:
-            uri = os.path.join("s3://", bucket_name, obj["Key"])
-            yield (uri, obj["Size"])
+            yield os.path.join("s3://", bucket_name, obj["Key"])
```
rymarczy, line 58: update function DocString since we are no longer yielding file size

rymarczy:
Only major comment is if the work flow relies on `file_list_from_s3` providing all s3 paths in chronological order. I know there is no way to sort the files you're getting from s3 with any of the boto3 file list functionality, so I think this could be a dangerous assumption. 

I believe all of the paths we are dealing with have a timestamp in the filename (created by delta), so we could perform our own sort, after the `file_list_from_s3` operation based on filename, and then that would include the assumption that the data in the files is the same chronologically as the filename being created by delta. I'm not totally sure how solid that assumption is.

py_gtfs_rt_ingestion/lib/ingest.py
mzappitello: yeah. if it wasn't, we'd have some real issues if there was a backlog of data, since we'd have to load all of the files into memory at the same time before a write.

PR #70, opened 2022-11-01
Deployment debugging

mzappitello:
A handful of small deployment issues are fixed here. This PR should be rebased before merging.

### DIFF
--- a/.github/workflows/ingestion-deploy-dev.yaml
+++ b/.github/workflows/ingestion-deploy-dev.yaml
@@ -14,7 +14,9 @@ jobs:
     uses: mbta/workflows/.github/workflows/deploy-ecs.yml@main
     with:
       app-name: lamp-ingestion
-      environment: 'dev'
+      environment: dev
+      dockerfile-path: ./py_gtfs_rt_ingestion/
+      cluster: lamp
     secrets:
       aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
       aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
--- a/.github/workflows/ingestion-deploy-dev.yaml
+++ b/.github/workflows/ingestion-deploy-dev.yaml
@@ -1,4 +1,4 @@
-name: Deploy ECS to Dev
+name: Deploy Ingestion to Dev
 
 on:
   workflow_run:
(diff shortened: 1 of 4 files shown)


.github/workflows/ingestion-deploy-dev.yaml
```diff
@@ -1,4 +1,4 @@
-name: Deploy ECS to Dev
+name: Deploy Ingestion to Dev
```
mzappitello, line 1: better name. this is what we see on the github actions tab so it needs to be good.

.github/workflows/ingestion-deploy-dev.yaml
```diff
@@ -14,7 +14,9 @@ jobs:
     uses: mbta/workflows/.github/workflows/deploy-ecs.yml@main
     with:
       app-name: lamp-ingestion
-      environment: 'dev'
+      environment: dev
```
mzappitello, line 17: no need for quote chars.

.github/workflows/ingestion-deploy-dev.yaml
```diff
@@ -14,7 +14,9 @@ jobs:
...
+      dockerfile-path: ./py_gtfs_rt_ingestion/
```
mzappitello, line 18: define the dockerfile path. the default was `.`, so this needed to change when we moved it into the subdirs.

.github/workflows/ingestion-deploy-dev.yaml
```diff
@@ -14,7 +14,9 @@ jobs:
...
+      cluster: lamp
```
mzappitello, line 19: both apps are deployed to the lamp cluster. without this it will assume the service name and cluster are the same string.

.github/workflows/performance-manager-deploy-dev.yaml
```diff
@@ -14,7 +14,9 @@ jobs:
     uses: mbta/workflows/.github/workflows/deploy-ecs.yml@main
     with:
       app-name: lamp-performance-manager
-      environment: 'dev'
+      environment: dev
+      dockerfile-path: ./performance_manager/
+      cluster: lamp
```
mzappitello, line 19: same three changes.

py_gtfs_rt_ingestion/pyproject.toml
```diff
@@ -20,7 +20,6 @@ pytest-cov = "^3.0.0"
 black = "^22.3.0"
 pylint = "^2.15.5"
 mypy = "^0.961"
-awscli = "^1.25.8"
```
mzappitello, line 23: we don't use this dependency and it was causing a bad poetry lock file to be generated, so i just removed it.

rymarczy:
LGTM

PR #73, opened 2022-11-03
Add pool_pre_ping to SQLAlchemy Engine

mzappitello:
The pool pre ping will try to ping the db before any other interactions, and reconnect if there is a problem, hopefully avoiding some connection errors from stale connections.


rymarczy:
LGTM

PR #74, opened 2022-11-07
Improve speed of S3 file move operations

rymarczy:
This PR is meant to speed up s3 file move operations in the ECS instance. 

Performance improvement is achieved with 2 main changes:
1. Switch from Threading to Multi-Processing
  -- Threading was previously required because lambda instances did not support multi-processing. Multi-processing processes are slightly more expensive to create, but more performant during run-time than threads. 
2. Use shared boto session object for each process. 
  -- Creating boto Session objects for every file move operation is very expensive for the number of file moves we are completing. Using the multi-processing Pool `initializer` keyword argument, we are able to create a Session at process initialization and use it for all move operations done during the life of the process. 

In my testing this update showed a minimum of a 10x improvement in s3 file moving speed. 

```
** Current Copy **
ingestion    | WARNING:root:1500 of 1500 files copied
ingestion    | WARNING:root:took 221.86 seconds, 0.147908 seconds per file

** Multi-Processing Copy **
ingestion    | WARNING:root:1500 of 1500 files copied
ingestion    | WARNING:root:took 17.16 seconds, 0.011437 seconds per file

ingestion    | WARNING:root:copy speed up: 12.93 x
```

Asana Task: https://app.asana.com/0/1203185253920158/1203312862152172

### DIFF
--- a/py_gtfs_rt_ingestion/lib/s3_utils.py
+++ b/py_gtfs_rt_ingestion/lib/s3_utils.py
@@ -1,8 +1,8 @@
-import json
 import os
 import pathlib
 
-from concurrent.futures import ThreadPoolExecutor
+from multiprocessing.pool import Pool
+from multiprocessing import current_process
 from io import BytesIO
 from typing import IO, List, Tuple, cast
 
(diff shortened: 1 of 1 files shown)


py_gtfs_rt_ingestion/lib/s3_utils.py
```diff
@@ -85,65 +85,47 @@ def strip_timestamp(fileobject: str) -> str:
...
-                  function
-
-    No return value
-    """
-    lambda_client = boto3.client("lambda")
-    lambda_client.invoke(
-        FunctionName=function_arn,
-        InvocationType="Event",
-        Payload=json.dumps(event),
-    )
```
rymarczy, line 103: don't need this anymore

py_gtfs_rt_ingestion/lib/s3_utils.py
```diff
@@ -85,65 +85,47 @@ def strip_timestamp(fileobject: str) -> str:
...
         filename=filename,
-        destination=destination,
+        to_bucket=to_bucket,
     )
     process_logger.log_start()
 
     try:
-        session = boto3.session.Session()
-        s3_resource = session.resource("s3")
+        s3_resource = current_process().__dict__["boto_s3_resource"]
```
rymarczy, line 106: pull `s3_resource` from process dictionary, instead of having the create one for every file move

py_gtfs_rt_ingestion/lib/s3_utils.py
```diff
@@ -171,14 +171,15 @@ def move_s3_objects(files: List[str], destination: str) -> None:
...
     )
     process_logger.log_start()
 
-    with ThreadPoolExecutor(max_workers=pool_size) as executor:
-        for filename in files:
-            executor.submit(_move_s3_object, filename, destination)
+    with Pool(pool_size, initializer=_init_process_session) as pool:
+        pool.starmap(
+            _move_s3_object, [(filename, to_bucket) for filename in files]
+        )
```
rymarczy, line 182: initialize each process with `_init_process_session` which saves a boto Session in the process dictionary object. this might not be the best way to save the Session, but it seems to work.

py_gtfs_rt_ingestion/lib/s3_utils.py
mzappitello: 👋

py_gtfs_rt_ingestion/lib/s3_utils.py
```diff
@@ -152,17 +134,35 @@ def _move_s3_object(filename: str, destination: str) -> None:
...
     """
-    Move list of S3 objects from source to destination.
+    initialization function for any process in multi processing pool needing to
+    use a boto session object
+
+    this avoids the expensive operation of creating a new session for every unti of work
+    in the pool
+
+    not totally sure this is the best way to retain the session on process initialization
+    but it seems to work
```
mzappitello, line 146: i guess this session is only going to live as long as we're moving a list of files, so this should be fine.

mzappitello:
I'm a little concerned how this will pair with each converter also being on its own thread. we are already doing that though, and these wins speak for themselves.

LGTM

PR #75, opened 2022-11-07
Ingest sort group files 2

mzappitello:
Our ingestion was taking way too long. To speed it up, put each filetypes conversion into its own thread.

### DIFF
--- a/py_gtfs_rt_ingestion/ingest.py
+++ b/py_gtfs_rt_ingestion/ingest.py
@@ -10,8 +10,6 @@ from lib import (
     ingest_files,
     file_list_from_s3,
     DEFAULT_S3_PREFIX,
-    move_s3_objects,
-    write_parquet_file,
     ProcessLogger,
 )
 
@@ -101,59 +99,7 @@ def ingest() -> None:
 
     total_filecount = 0
 
-    for converter in ingest_files(files):
-        sub_process_logger = ProcessLogger("process_converter")
-        sub_process_logger.log_start()
-
-        archive_files = []
-        error_files = []
-
-        try:
-            for s3_prefix, table in converter.get_tables():
-                write_parquet_file(
-                    table=table,
-                    filetype=s3_prefix,
-                    s3_path=os.path.join(
-                        os.environ["SPRINGBOARD_BUCKET"],
-                        DEFAULT_S3_PREFIX,
-                        s3_prefix,
-                    ),
-                    partition_cols=converter.partition_cols,
-                )
-
-            archive_files = converter.archive_files
-            error_files = converter.error_files
-
-        except Exception as exception:
-            sub_process_logger.log_failure(exception)
-            # if unable to determine config from filename, or not implemented
-            # yet, all files are marked as failed ingestion
-            archive_files = []
-            error_files = converter.archive_files + converter.error_files
-
-        finally:
-            if len(error_files) > 0:
-                move_s3_objects(
-                    error_files,
-                    os.path.join(os.environ["ERROR_BUCKET"], DEFAULT_S3_PREFIX),
-                )
-
-            if len(archive_files) > 0:
-                move_s3_objects(
-                    archive_files,
-                    os.path.join(
-                        os.environ["ARCHIVE_BUCKET"], DEFAULT_S3_PREFIX
-                    ),
-                )
-
-            sub_process_logger.add_metadata(
-                error_filecount=len(error_files),
-                archive_filecount=len(archive_files),
-            )
-
-            total_filecount += len(error_files) + len(archive_files)
-
-            sub_process_logger.log_complete()
+    ingest_files(files)
 
     process_logger.add_metadata(total_filecount=total_filecount)
     process_logger.log_complete()
--- a/py_gtfs_rt_ingestion/lib/__init__.py
+++ b/py_gtfs_rt_ingestion/lib/__init__.py
@@ -10,8 +10,6 @@ from .lambda_types import LambdaContext, LambdaDict
 from .logging_utils import ProcessLogger
 from .postgres_utils import insert_metadata, get_local_engine
 from .s3_utils import file_list_from_s3, move_s3_objects, write_parquet_file
-from .utils import load_environment
+from .utils import load_environment, group_sort_file_list, DEFAULT_S3_PREFIX
 
 __version__ = "0.1.0"
-
-DEFAULT_S3_PREFIX = "lamp"
--- a/py_gtfs_rt_ingestion/lib/convert_gtfs.py
+++ b/py_gtfs_rt_ingestion/lib/convert_gtfs.py
@@ -1,3 +1,4 @@
+import os
 import zipfile
 
 from typing import IO, List, Tuple, Union
@@ -5,9 +6,41 @@ from typing import IO, List, Tuple, Union
 import pyarrow
 from pyarrow import csv
 
-from .s3_utils import get_zip_buffer
-from .converter import Converter, ConfigType
+from .converter import Converter
 from .logging_utils import ProcessLogger
+from .s3_utils import get_zip_buffer, move_s3_objects, write_parquet_file
+from .utils import DEFAULT_S3_PREFIX
+
+
+def zip_to_pyarrow(filename: str) -> List[Tuple[str, pyarrow.Table]]:
+    """
+    convert a schedule gtfs zip file into tables. the zip file is
+    essentially a small database with each contained file (outside of feed
+    info) acting as its own table. info on the gtfs scheduling standard can
+    be found at http://gtfs.org/schedule/
+    """
+    tables = []
+    filelike_input: Union[str, IO[bytes]] = filename
+    last_modified = 0
+
+    if filename.startswith("s3://"):
+        file_to_load = str(filename).replace("s3://", "")
+        filelike_input, last_modified = get_zip_buffer(file_to_load)
+
+    with zipfile.ZipFile(filelike_input) as gtfs_zip:
+        # iterate over each "table"
+        for gtfs_filename in gtfs_zip.namelist():
+            with gtfs_zip.open(gtfs_filename) as file:
+                table = csv.read_csv(file)
+
+                # add the last modified timestamp
+                timestamp = [last_modified] * table.num_rows
+                table = table.append_column("timestamp", [timestamp])
+
+            filename_prefix = gtfs_filename.replace(".txt", "").upper()
+            tables.append((filename_prefix, table))
+
+        return tables
 
 
 class GtfsConverter(Converter):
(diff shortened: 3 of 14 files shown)


py_gtfs_rt_ingestion/ingest.py
```diff
@@ -101,59 +99,7 @@ def ingest() -> None:
...
-
-            sub_process_logger.add_metadata(
-                error_filecount=len(error_files),
-                archive_filecount=len(archive_files),
-            )
-
-            total_filecount += len(error_files) + len(archive_files)
-
-            sub_process_logger.log_complete()
+    ingest_files(files)
```
mzappitello, line 102: most of this detail is now encapsulated in either the `ingest_files` method itself or inside of the converters it creats.

py_gtfs_rt_ingestion/lib/__init__.py
```diff
@@ -10,8 +10,6 @@
 from .logging_utils import ProcessLogger
 from .postgres_utils import insert_metadata, get_local_engine
 from .s3_utils import file_list_from_s3, move_s3_objects, write_parquet_file
-from .utils import load_environment
+from .utils import load_environment, group_sort_file_list, DEFAULT_S3_PREFIX
 
 __version__ = "0.1.0"
-
-DEFAULT_S3_PREFIX = "lamp"
```
mzappitello, line 17: had to move this since its being used internally now.

py_gtfs_rt_ingestion/lib/convert_gtfs.py
```diff
@@ -1,76 +1,90 @@
...
+                table = csv.read_csv(file)
+
+                # add the last modified timestamp
+                timestamp = [last_modified] * table.num_rows
+                table = table.append_column("timestamp", [timestamp])
+
+            filename_prefix = gtfs_filename.replace(".txt", "").upper()
+            tables.append((filename_prefix, table))
+
+        return tables
```
mzappitello, line 43: no longer a member function.

py_gtfs_rt_ingestion/lib/convert_gtfs.py
```diff
@@ -1,76 +1,90 @@
...
+            move_s3_objects(
+                error_files,
+                os.path.join(os.environ["ERROR_BUCKET"], DEFAULT_S3_PREFIX),
+            )
+
+        if len(archive_files) > 0:
+            move_s3_objects(
+                archive_files,
+                os.path.join(os.environ["ARCHIVE_BUCKET"], DEFAULT_S3_PREFIX),
+            )
```
mzappitello, line 90: convert the file to tables, write them, move files.

much less boilerplate by encapsulating this.

py_gtfs_rt_ingestion/lib/convert_gtfs_rt.py
```diff
@@ -55,65 +58,69 @@ def __init__(self, config_type: ConfigType) -> None:
         now = datetime.now(tz=timezone.utc)
         self.start_of_hour = now.replace(minute=0, second=0, microsecond=0)
 
-        self.current_time: Optional[datetime] = None
-        self.table = pyarrow.table(
-            self.detail.empty_table(), schema=self.detail.export_schema
-        )
-        self.next_table: Optional[pyarrow.table] = None
```
mzappitello, line 62: we can get rid of this requirement. of the class.

py_gtfs_rt_ingestion/lib/convert_gtfs_rt.py
```diff
@@ -55,65 +58,69 @@ def __init__(self, config_type: ConfigType) -> None:
...
+    def convert(self) -> None:
+        process_logger = ProcessLogger("gtfs_rt_convert")
+        process_logger.log_start()
 
-        self.archive_files = []
-        self.error_files = []
+        for table in self.process_files():
+            self.write_and_archive(table)
+            process_logger.log_complete()
+            process_logger.log_start()
```
mzappitello, line 71: `process_files` goes through and consumes s3 files until the hour switchover, when it yields a new one.

when a new one is processed, write it and archvive files.

py_gtfs_rt_ingestion/lib/convert_gtfs_rt.py
```diff
@@ -55,65 +58,69 @@ def __init__(self, config_type: ConfigType) -> None:
...
 
-    def add_file(self, file: str) -> bool:
-        process_logger = ProcessLogger(
-            "convert_single_gtfs_rt",
-            config_type=str(self.config_type),
-            filename=file,
+    def process_files(self) -> Iterable[pyarrow.table]:
```
mzappitello, line 73: this function looks a lot like what we had before, which makes sense since each lambda was essentially a parallelized unit.

py_gtfs_rt_ingestion/lib/convert_gtfs_rt.py
```diff
@@ -186,6 +193,39 @@ def gz_to_pyarrow(self, filename: str) -> Tuple[datetime, pyarrow.table]:
...
+                    os.path.join(os.environ["ERROR_BUCKET"], DEFAULT_S3_PREFIX),
+                )
+
+            if len(self.archive_files) > 0:
+                move_s3_objects(
+                    self.archive_files,
+                    os.path.join(
+                        os.environ["ARCHIVE_BUCKET"], DEFAULT_S3_PREFIX
+                    ),
+                )
```
mzappitello, line 231: pretty straight forward. this was what we were doing in the ingest script before.

py_gtfs_rt_ingestion/lib/converter.py
```diff
@@ -89,34 +87,13 @@ class Converter(ABC):
...
 
-    @property
     @abstractmethod
-    def partition_cols(self) -> list[str]:
+    def convert(self) -> None:
         """
-        list of columns to patition out of parquet tables
+        convert files to pyarrow tables, write them to s3 as parquete, and move
+        files from incoming to archive (or error)
         """
```
mzappitello, line 99: much simplified api for convert.

py_gtfs_rt_ingestion/lib/ingest.py
```diff
@@ -18,68 +19,46 @@ class NoImplConverter(Converter):
...
-        self.error_files.append(file)
-        return len(self.error_files) > 100
-
-    def reset(self) -> None:
-        self.error_files = []
+    def convert(self) -> None:
+        move_s3_objects(
+            self.files,
+            os.path.join(os.environ["ERROR_BUCKET"], DEFAULT_S3_PREFIX),
+        )
```
mzappitello, line 26: convert is just a big move operation.

py_gtfs_rt_ingestion/lib/ingest.py
```diff
@@ -18,68 +19,46 @@ class NoImplConverter(Converter):
...
-            prefixes and files to move to error and archive buckets
+    sort the incoming file list by type and create a converter for each type.
+    each converter will ingest and convert its files in its own thread.
     """
-    converters = {t: get_converter(t) for t in ConfigType}
+    grouped_files = group_sort_file_list(files)
 
-    for file in files:
+    converters = []
+    for file_group in grouped_files.values():
```
mzappitello, line 54: go through each file group, create a converter for that group, and use their convert functions in their own threads.

py_gtfs_rt_ingestion/lib/logging_utils.py
```diff
@@ -46,7 +46,7 @@ def log_start(self) -> None:
 
     def log_complete(self) -> None:
         """log the completion of a proccess with duration"""
-        duration = self.start_time - time.time()
```
mzappitello, line 49: fix this annoying thing.

py_gtfs_rt_ingestion/lib/postgres_utils.py
```diff
@@ -88,82 +86,6 @@ def get_local_engine(echo: bool = False) -> sa.engine.Engine:
         raise exception
 
 
-def get_psql_conn() -> psycopg2.extensions.connection:
```
mzappitello, line 91: we're using a sqlalchemy connection, so 👋

py_gtfs_rt_ingestion/lib/s3_utils.py
```diff
@@ -46,25 +45,13 @@ def file_list_from_s3(bucket_name: str, file_prefix: str) -> List[str]:
...
 
-    :return list of s3 filepaths sorted by the timestamps formatted into them
+    :return list of s3 filepaths
     """
     process_logger = ProcessLogger(
         "file_list_from_s3", bucket_name=bucket_name, file_prefix=file_prefix
     )
     process_logger.log_start()
 
-    def strip_timestamp(fileobject: str) -> str:
```
mzappitello, line 56: rather than doing the sort here, do it in the utility that buckets these by filetype.

py_gtfs_rt_ingestion/tests/test_gtfs_rt_converter.py
```diff
@@ -16,19 +16,18 @@ def test_bad_conversion_local() -> None:
     be added to the error files list
     """
     # dummy config to avoid mypy errors
-    converter = GtfsRtConverter(config_type=ConfigType.RT_ALERTS)
+    converter = GtfsRtConverter(
```
mzappitello, line 19: all of this uses the `process_files` method rather than `convert` since we don't want to write to s3, nor archive files.

py_gtfs_rt_ingestion/lib/convert_gtfs_rt.py
```diff
@@ -30,8 +33,8 @@ class GtfsRtConverter(Converter):
     https_mbta_integration.mybluemix.net_vehicleCount.gz
     """
 
-    def __init__(self, config_type: ConfigType) -> None:
-        Converter.__init__(self, config_type)
+    def __init__(self, config_type: ConfigType, files: List[str]) -> None:
+        Converter.__init__(self, config_type, files)
```
rymarczy, line 37: I think it would be worth adding the pyarrow filesystem initialization to `__init__` . it's another fairly expensive object creation that we are performing for every gzip that we ingest. 

I did some profiling and got between 35% and 40% improvement in bulk file downloading speed with just one `fs` creation per thread.

rymarczy:
Few comments. Most important probably being that we need to process `gtfs_static` files before any `gtfs_rt` files to avoid race conditions in `performance_manager` loading.

py_gtfs_rt_ingestion/lib/convert_gtfs_rt.py
mzappitello: i addressed this. we initialize a local filesystem first, and at the start of processing files, change it if the first file has an s3 prefix. that filesystem is used for all reading.

mzappitello:
added a semi big change to the ingest_files system, now there is only one `NoImplConverter` which will do all the moving for files we don't currently process. the GTFS Converter is run first.

rymarczy:
LGTM, we'll need to keep an eye on the ECS resource usage and the size of `ThreadPool`

PR #79, opened 2022-11-14
Fix try/except in gtfs_rt convert

rymarczy:
Add try/except to `convert` for gtfs_rt

Add logging start in `process_files`


mzappitello:
LGTM

PR #80, opened 2022-11-18
Rate Limit Ingestion Application 

rymarczy:
Changes in this PR:
1. Rate limit ingest application.
  - Rate limit is achieved by limiting the number of files pulled from `file_list_from_s3` by default this values is set at 250,000. We see at maximum ~6,000 files per hour for gtfs-rt processing. The gtfs-rt `convert` function has also been limited to produce a maximum of 10 tables per event loop. 10 * 6,000 = 60,000 which is significantly less than the 250,000 files we're pulling from s3.
2. Unlimited retry attempts on `insert_metadata`.
  - `insert_metadata` can not and should not fail to insert records. Any failure can be attributed to RDS connection issues which should be able to resolved while insert attempts are made.
3. Minor logging changes to `performance_manager`

Asana Task: https://app.asana.com/0/1203185253920158/1203399959411145

### DIFF
--- a/performance_manager/lib/l1_full_trip_events.py
+++ b/performance_manager/lib/l1_full_trip_events.py
@@ -308,7 +308,7 @@ def update_with_new_events(
     alternate approach is used here where just hashes are selected that represent
     update records and each column is updated seperatly from insert_dataframe
     """
-    process_logger = ProcessLogger("l1_update_events")
+    process_logger = ProcessLogger("l1_load_table")
     process_logger.log_start()
 
     # get dataframe of hashes represnting records in FullTripEvent table that
@@ -319,7 +319,7 @@ def update_with_new_events(
     )
     hashes_to_update = db_manager.select_as_dataframe(hash_to_update_query)
 
-    process_logger.add_metadata(trip_events_to_update=hashes_to_update.shape[0])
+    process_logger.add_metadata(total_updated=hashes_to_update.shape[0])
 
     # gate to see if any records need updating
     if hashes_to_update.shape[0] == 0:
@@ -340,6 +340,7 @@ def update_with_new_events(
         to_update_mask
     )
     # if mask shows records need updating, execute update
+    process_logger.add_metadata(fk_vp_moving_events_updated=0)
     if vp_moving_mask.sum() > 0:
         result = db_manager.execute_with_data(
             update_query,
@@ -348,14 +349,13 @@ def update_with_new_events(
             ].rename(columns={"hash": "b_hash"}),
         )
         process_logger.add_metadata(fk_vp_moving_events_updated=result.rowcount)
-    else:
-        process_logger.add_metadata(fk_vp_moving_events_updated=0)
 
     # mask for insert dataframe and vp_stopped events
     vp_stopped_mask = (~insert_dataframe["fk_vp_stopped_event"].isna()) & (
         to_update_mask
     )
     # if mask shows records need updating, execute update
+    process_logger.add_metadata(fk_vp_stopped_events_updated=0)
     if vp_stopped_mask.sum() > 0:
         result = db_manager.execute_with_data(
             update_query,
@@ -366,14 +366,13 @@ def update_with_new_events(
         process_logger.add_metadata(
             fk_vp_stopped_events_updated=result.rowcount
         )
-    else:
-        process_logger.add_metadata(fk_vp_stopped_events_updated=0)
 
     # mask for insert dataframe and tu_stopped events
     tu_stopped_mask = (~insert_dataframe["fk_tu_stopped_event"].isna()) & (
         to_update_mask
     )
     # if mask shows records need updating, execute update
+    process_logger.add_metadata(fk_tu_stopped_events_updated=0)
     if tu_stopped_mask.sum() > 0:
         result = db_manager.execute_with_data(
             update_query,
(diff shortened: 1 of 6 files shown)


performance_manager/lib/l1_full_trip_events.py
```diff
@@ -308,7 +308,7 @@ def update_with_new_events(
     alternate approach is used here where just hashes are selected that represent
     update records and each column is updated seperatly from insert_dataframe
     """
-    process_logger = ProcessLogger("l1_update_events")
+    process_logger = ProcessLogger("l1_load_table")
```
rymarczy, line 311: Minor logging changes in `performance_manager` files. This is a work in a progres.

py_gtfs_rt_ingestion/lib/convert_gtfs_rt.py
```diff
@@ -65,6 +65,7 @@ def __init__(self, config_type: ConfigType, metadata_queue: Queue) -> None:
         self.active_fs = fs.LocalFileSystem()
 
     def convert(self) -> None:
+        max_tables_to_convert = 10
         process_logger = ProcessLogger(
```
rymarczy, line 69: maximum of 10 tables per event loop for each `gtfs-rt` converter type

py_gtfs_rt_ingestion/lib/convert_gtfs_rt.py
```diff
@@ -82,6 +83,11 @@ def convert(self) -> None:
                 # only count table if it contains data
                 if table.num_rows > 0:
                     table_count += 1
+
+                # limit number of tables produced on each event loop
+                if table_count >= max_tables_to_convert:
+                    break
+
```
rymarczy, line 90: exit, if max tables reached for event loop, do not archive or error remaining files

py_gtfs_rt_ingestion/lib/convert_gtfs_rt.py
```diff
@@ -115,12 +122,6 @@ def process_files(self) -> Iterable[pyarrow.table]:
             try:
                 timestamp, new_data = self.gz_to_pyarrow(file)
 
-                # skip files that have been generated after the start of the
-                # current hour. don't add them to archive or error so that they
-                # are picked up next go around.
-                if timestamp >= self.start_of_hour:
-                    break
-
```
rymarczy, line 123: Moved this to after our yield in the hour break. We should only be yielding on hour breaks and not ever yielding a table that only contains a partial hour

py_gtfs_rt_ingestion/lib/convert_gtfs_rt.py
```diff
@@ -153,18 +154,21 @@ def process_files(self) -> Iterable[pyarrow.table]:
...
 
             except Exception:
                 self.error_files.append(file)
 
-        process_logger.add_metadata(
-            file_count=file_count, number_of_rows=table.num_rows
-        )
+        process_logger.add_metadata(file_count=0, number_of_rows=0)
         process_logger.log_complete()
-        yield table
```
rymarczy, line 167: Remove this `yield` so we are only yielding tables on hour breaks, never in the middle of an hour

py_gtfs_rt_ingestion/lib/convert_gtfs_rt.py
```diff
@@ -153,18 +154,21 @@ def process_files(self) -> Iterable[pyarrow.table]:
...
                 table = pyarrow.concat_tables([table, new_data])
                 self.archive_files.append(file)
                 file_count += 1
 
             except Exception:
                 self.error_files.append(file)
 
-        process_logger.add_metadata(
-            file_count=file_count, number_of_rows=table.num_rows
-        )
```
rymarczy, line 165: If we hit this, the existing process will `yield` no files or rows.

py_gtfs_rt_ingestion/lib/postgres_utils.py
```diff
@@ -187,28 +182,24 @@ def _rds_writer_process(metadata_queue: Queue) -> None:
...
-                if retry_attempt == retry_count - 1:
-                    bad_insert += 1
-                    process_logger.log_failure(exception)
-                else:
-                    # wait for gremlins to disappear
-                    time.sleep(15)
+            except Exception as _:
+                retry_attempt += 1
+                # wait for gremlins to disappear
+                time.sleep(15)
```
rymarczy, line 196: Change this to unlimited retry attempts. This should never fail unless RDS is having issues, in which case we should keep the queue of metadata files to write and write them when RDS connection is restored.

py_gtfs_rt_ingestion/lib/postgres_utils.py
```diff
@@ -187,28 +182,24 @@ def _rds_writer_process(metadata_queue: Queue) -> None:
...
 
             else:
-                good_insert += 1
+                process_logger.add_metadata(retry_attempts=retry_attempt)
                 process_logger.log_complete()
                 break
 
-    process_logger.add_metadata(
-        insert_success_count=good_insert, insert_fail_count=bad_insert
-    )
```
rymarczy, line 211: Clean up this logging these values were not going to be helpful

py_gtfs_rt_ingestion/lib/s3_utils.py
```diff
@@ -64,6 +66,9 @@ def file_list_from_s3(bucket_name: str, file_prefix: str) -> List[str]:
             for obj in page["Contents"]:
                 filepaths.append(os.path.join("s3://", bucket_name, obj["Key"]))
 
+            if len(filepaths) > max_list_size:
+                break
+
```
rymarczy, line 71: break if `max_list_size` hit. This will be checked on each new page which are ~2,000 files each. So might overshoot this value by that number, but shouldn't really matter for these purposes.

mzappitello:
pretty straight forward.

LGTM

PR #81, opened 2022-11-18
Add Postgres Tests to Performance Manager

mzappitello:
* Add tests that create a DatabaseManager object and ensure we can connect to the database.
* Add pytest to Performance Manager CI
* Add a postgres service to the test step in Performance Manager CI

Asana Task: https://app.asana.com/0/1203100082488441/1203144932470000/f

### DIFF
--- a/.github/workflows/performance-manager-ci.yaml
+++ b/.github/workflows/performance-manager-ci.yaml
@@ -57,3 +57,31 @@ jobs:
         with:
           poetry-dir: performance_manager
       - run: poetry run pylint *.py lib/ tests/
+
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    needs: setup
+    env:
+      BOOTSTRAPPED: 1
+      DB_HOST: localhost
+      DB_PORT: 5432
+      DB_NAME: pm_test
+      DB_USER: postgres
+      DB_PASSWORD: postgres
+    services:
+      postgres:
+        image: postgres
+        ports:
+          - 5432:5432
+        env:
+          POSTGRES_PASSWORD: ${{env.DB_PASSWORD}}
+          POSTGRES_USER: ${{env.DB_USER}}
+          POSTGRES_DB: ${{env.DB_NAME}}
+        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/performance-manager-ci.yaml
```diff
@@ -57,3 +57,31 @@ jobs:
...
+          POSTGRES_PASSWORD: ${{env.DB_PASSWORD}}
+          POSTGRES_USER: ${{env.DB_USER}}
+          POSTGRES_DB: ${{env.DB_NAME}}
+        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ./.github/actions/poetry_deps
+        with:
+          poetry-dir: performance_manager
+      - run: poetry run pytest
```
mzappitello, line 87: add in step to run pytest. turns out we weren't doing this at all before?

in addition to running the tests, i added a postgres service that will be spun up in parallel that we can use to test against.

performance_manager/tests/test_postgres.py
```diff
@@ -0,0 +1,49 @@
...
+    """simple test stub"""
+    assert int(os.environ.get("BOOTSTRAPPED", 0)) == 1
+
+    db_manager = DatabaseManager()
+
+    paths_to_load = get_unprocessed_files(
+        "RT_VEHICLE_POSITIONS", db_manager, file_limit=6
+    )
+
+    assert len(paths_to_load) == 0
```
mzappitello, line 49: pretty small test suite here, but shows that we can connect to the database and run queries.

.github/workflows/performance-manager-ci.yaml
```diff
@@ -57,3 +57,31 @@ jobs:
...
+      BOOTSTRAPPED: 1
+      DB_HOST: localhost
+      DB_PORT: 5432
+      DB_NAME: pm_test
+      DB_USER: postgres
+      DB_PASSWORD: postgres
+    services:
+      postgres:
+        image: postgres
+        ports:
```
rymarczy, line 75: use specific postgres image to match AWS RDS

rymarczy:
Just need specific postgres image, otherwise LGTM

PR #86, opened 2022-12-06
Bug Fix: Trip Update Loading (explode_stop_time_update)

rymarczy:
This fixes a  bug caused by the apparent batching of the `numpy.vectorize` function. 

Very infrequently a `None` type would be passed into `explode_stop_time_update` as the `stop_time_update` parameter. 

Asana Task: https://app.asana.com/0/1203185331040541/1203495730837934

### DIFF
--- a/performance_manager/lib/l0_rt_trip_updates.py
+++ b/performance_manager/lib/l0_rt_trip_updates.py
@@ -54,7 +54,7 @@ def get_tu_dataframe_chunks(
 
 # pylint: disable=too-many-arguments
 def explode_stop_time_update(
-    stop_time_update: List[Dict[str, Any]],
+    stop_time_update: Optional[List[Dict[str, Any]]],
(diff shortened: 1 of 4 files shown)


performance_manager/lib/l0_rt_trip_updates.py
```diff
@@ -75,13 +75,20 @@ def explode_stop_time_update(
         "vehicle_id": vehicle_id,
     }
     return_list: List[Dict[str, Any]] = []
+
+    # fix: https://app.asana.com/0/1203185331040541/1203495730837934
+    # it appears that numpy.vectorize batches function inputs which can
+    # result in None being passed in for stop_time_update
+    if stop_time_update is None:
+        return None
```
mzappitello, line 83: this feels like its _exactly_ the type of error that mypy is supposed to catch.

PR #90, opened 2023-02-03
Events refactor

mzappitello:
Refactor our Events Table and How they are Computed

Asana Task: https://app.asana.com/0/1203655790984034/1203774744918072/f


performance_manager/lib/__init__.py
```diff
@@ -3,15 +3,10 @@
 analize the performance of the MBTA system.
 """
 
+from .l0_gtfs_rt_events import process_gtfs_rt_files
 from .l0_gtfs_static_table import process_static_tables
-from .l0_rt_trip_updates import process_trip_updates
-from .l0_rt_vehicle_positions import process_vehicle_positions
-from .l1_full_trip_events import process_full_trip_events
-from .l2_dwell_travel_times import process_dwell_travel_times
-from .l2_headways import process_headways
```
mzappitello, line 11: these processes are either gone or folded into `process_gtfs_rt_files`

performance_manager/lib/postgres_schema.py
```diff
@@ -6,63 +6,87 @@
...
 
     pk_id = sa.Column(sa.Integer, primary_key=True)
-    is_moving = sa.Column(sa.Boolean)
-    stop_sequence = sa.Column(sa.SmallInteger, nullable=True)
-    stop_id = sa.Column(sa.String(60), nullable=True)
-    timestamp_start = sa.Column(sa.Integer, nullable=False)
-    timestamp_end = sa.Column(sa.Integer, nullable=False)
+
+    # trip identifiers
     direction_id = sa.Column(sa.SmallInteger, nullable=True)
```
rymarczy, line 23: direction_id can be a bool type, now that we're planning on never having Null values in these fields

performance_manager/lib/postgres_schema.py
```diff
@@ -6,63 +6,87 @@
...
     route_id = sa.Column(sa.String(60), nullable=True)
     start_date = sa.Column(sa.Integer, nullable=True)
     start_time = sa.Column(sa.Integer, nullable=True)
     vehicle_id = sa.Column(sa.String(60), nullable=False)
-    hash = sa.Column(
+
+    # stop identifiers
+    stop_sequence = sa.Column(sa.SmallInteger, nullable=True)
+    stop_id = sa.Column(sa.String(60), nullable=True)
+    parent_station = sa.Column(sa.String(60), nullable=True)
```
rymarczy, line 32: All of these can be `nullable=False` now

performance_manager/lib/postgres_schema.py
```diff
@@ -6,63 +6,87 @@
...
+
+    # hash of trip and stop identifiers
+    trip_stop_hash = sa.Column(
         sa.LargeBinary(16), nullable=False, index=True, unique=False
     )
```
rymarczy, line 37: `unique=True`

performance_manager/lib/s3_utils.py
```diff
@@ -17,11 +17,21 @@ def _get_pyarrow_table(
...
+        else:
+            active_fs = fs.LocalFileSystem()
+            to_load = filename
     else:
-        to_load = [filename.replace("s3://", "")]
+        if filename.startswith("s3://"):
+            to_load = [filename.replace("s3://", "")]
+        else:
+            active_fs = fs.LocalFileSystem()
+            to_load = [filename]
```
rymarczy, line 34: This `fs.LocalFileSystem()` logic breaks our `--seed` flag path loads. If `fs.LocalFileSystem()` is only used for testing maybe we can handle with and ENV flag that is set just for testing??

mzappitello: i think the better method is to do a monkey patch here.

mzappitello:
replaced by https://github.com/mbta/lamp/pull/91

PR #95, opened 2023-02-23
Fix signal+multiprocessing bug

rymarczy:
Our use of the following functions, in combination, causes `SIGTERM` signals to be thrown and blocking of main event loop.

- signal.signal
- multiprocessing.Pool.map
- multiprocessing.Manager

Launching all `Converter` processes with `multiprocessing.Pool.map_async` resolves the issue of throwing `SIGTERM` signals and blocking the main event loop. 

Also added additional steps to `check_for_sigterm` function to shutdown rds_writer_process when `SIGTERM` signal is received from AWS. 

Asana Task: https://app.asana.com/0/1203655790984049/1204005780226888

### DIFF
--- a/py_gtfs_rt_ingestion/ingest.py
+++ b/py_gtfs_rt_ingestion/ingest.py
@@ -110,12 +110,12 @@ def main() -> None:
     """every second run jobs that are currently pending"""
     # start rds writer process
     # this will create only one rds engine while app is running
-    metadata_queue, _rds_process = start_rds_writer_process()
+    metadata_queue, rds_process = start_rds_writer_process()
 
     schedule.every(5).minutes.do(ingest, metadata_queue=metadata_queue)
 
     while True:
-        check_for_sigterm(metadata_queue)
+        check_for_sigterm(metadata_queue, rds_process)
         schedule.run_pending()
         time.sleep(1)
 
--- a/py_gtfs_rt_ingestion/ingest.py
+++ b/py_gtfs_rt_ingestion/ingest.py
@@ -93,7 +93,6 @@ def ingest(metadata_queue: Queue) -> None:
     filepaths to the metadata table as unprocessed, and move gtfs files to the
     archive bucket (or error bucket in the event of an error)
     """
-    check_for_sigterm()
     process_logger = ProcessLogger("ingest_all")
     process_logger.log_start()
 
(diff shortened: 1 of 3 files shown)


py_gtfs_rt_ingestion/ingest.py
```diff
@@ -111,11 +110,12 @@ def main() -> None:
...
     # this will create only one rds engine while app is running
-    metadata_queue, _rds_process = start_rds_writer_process()
+    metadata_queue, rds_process = start_rds_writer_process()
 
     schedule.every(5).minutes.do(ingest, metadata_queue=metadata_queue)
 
     while True:
+        check_for_sigterm(metadata_queue, rds_process)
         schedule.run_pending()
         time.sleep(1)
```
rymarczy, line 120: this seems like a better spot for `check_for_sigterm` function, is continuously checked whenever `ingest` is not running.

py_gtfs_rt_ingestion/lib/ecs.py
```diff
@@ -13,10 +14,13 @@ def handle_ecs_sigterm(_: int, __: Any) -> None:
...
+def check_for_sigterm(metadata_queue: Queue, rds_process: Process) -> None:
     """
     check if SIGTERM recived from ECS. If found, terminate process.
     """
     if os.environ.get("GOT_SIGTERM") is not None:
         logging.info("SIGTERM received, terminating process...")
+        # send signal to stop rds writer process and wait for exit
+        metadata_queue.put(None)
+        rds_process.join()
         sys.exit()
```
rymarczy, line 26: shutdown rds_writer_process before `sys.exit` call. cleans everything up more nicely

py_gtfs_rt_ingestion/lib/ingest.py
```diff
@@ -81,5 +81,12 @@ def ingest_files(files: List[str], metadata_queue: Queue) -> None:
...
+    #
+    # the use of signal.signal, multiprocessing.Manager and multiprocessing.Pool.map, in combination,
+    # causes inadvertant SIGTERM signals to be sent by application and main event loop to be blocked
+    #
+    # launching converter processes with map_async avoids throwing of SIGTERM signals and blocking
     with Pool(processes=len(converters)) as pool:
-        pool.map(run_converter, converters.values())
+        pool.map_async(run_converter, converters.values())
+        pool.close()
+        pool.join()
```
rymarczy, line 92: error is documented in comment. `map_async` fixes everything!

PR #96, opened 2023-02-27
Business Logic for Trips and Metrics table loading

rymarczy:
Working version of Trips table and Metrics table loading.

Lots of SQL business logic in this change. For both the Trips and Metrics table loading, all load and compare logic is done SQL server side. Update / Insert query cleanly pulls all records requiring update and insert into dataframe.

Additional comments are needed for SQL CTE's and identification of possible edge cases that are not currently being handled. 

Asana Task: https://app.asana.com/0/1203655790984044/1203879577067931
Asana Task: https://app.asana.com/0/1203655790984044/1203655790983994
Asana Task: https://app.asana.com/0/1203655790984044/1203655790983996

### DIFF
--- a/performance_manager/lib/postgres_utils.py
+++ b/performance_manager/lib/postgres_utils.py
@@ -10,9 +10,14 @@ import sqlalchemy as sa
 from sqlalchemy.orm import sessionmaker
 
 from .logging_utils import ProcessLogger
-from .postgres_schema import MetadataLog, SqlBase
 from .s3_utils import get_utc_from_partition_path
-from .postgres_schema import VehicleEventMetrics, VehicleEvents, VehicleTrips, MetadataLog
+from .postgres_schema import (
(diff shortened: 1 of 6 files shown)


performance_manager/lib/postgres_utils.py
```diff
@@ -342,6 +353,29 @@ def seed_metadata(self) -> None:
         except Exception as exception:
             process_logger.log_failure(exception)
 
+    def reset_tables(self, clear_rt: bool, clear_static: bool) -> None:
```
mzappitello, line 356: i like this utility.

rymarczy:
For this specific PR, I'm expecting comments from the OPMI QA process to alter the business logic we're using. But once the QA process is over, and our business logic is solidified, we will be adding tests to the `test_postgres.py` framework @mzappitello has created to add coverage for these new tables.

PR #97, opened 2023-03-13
Single Python Source Tree

mzappitello:
We had a handful of overlapping features between Ingestion and Performance Manager that we wanted to make drier. One option was to pop them out into a third library and then have the two of them depend on the third shared tooling. I've opted instead to create one larger python project that we can run scripts, test, and deploy from.

Asana Task: [asana_ticket_url](https://app.asana.com/0/0/1204119636952121/f)


mzappitello:
replaced by https://github.com/mbta/lamp/pull/98

PR #98, opened 2023-03-15
Single Python Package

mzappitello:
Up until now we've been developing a different python package for each
data pipeline in the lamp ecosystem. There is some shared code between
these two packages, and we want to make that dry-er before we started
adding on migration scripts. After consulting with other engineers I
decided to settle on a single package where different utilities and
pipelines have their own subdirectories.

Asana Task: https://app.asana.com/0/0/1204119636952121/f


python_src/poetry.lock
```diff
@@ -0,0 +1,1008 @@
+[[package]]
```
mzappitello, line 1: classic poetry lock file for the new package. it stays checked in to make sure everything is the same version.

python_src/src/lamp_py/postgres/postgres_utils.py
```diff
@@ -342,8 +345,15 @@ def seed_metadata(self) -> None:
...
-                HERE, "..", "tests", "july_17_filepaths.json"
+                HERE,
+                "..",
+                "..",
+                "..",
+                "tests",
+                "test_files",
+                "july_17_filepaths.json",
             )
+            seed_file = os.path.abspath(seed_file)
```
mzappitello, line 357: had to update the relative path to find the seed info here. i think my preference is to not do this as a command line argument for this since we then have to include this json file in our docker images if we want to run it in docker compose for local testing.

i have two suggestions for how we achieve this in the future (and we have [a ticket](https://app.asana.com/0/0/1204188000578034/f) to do so). 

1. write a script that would connect to our local rds spun up by docker compose and insert it
2. use volumes to add the file to the container when spawning and pass that file in as a command line argument

one seems easier imho.

python_src/src/lamp_py/startup/start_ingestion.py
```diff
@@ -8,49 +8,18 @@
...
-                logging.info("setting %s to %s", key, value)
-                os.environ[key] = value
-
-    except FileNotFoundError as fnfe:
-        logging.warning("unable to find env file %s", fnfe)
-    except Exception as exception:
-        logging.exception("error while trying to bootstrap")
-        raise exception
-
-
```
mzappitello, line 53: collapsed this into a single util file that i already commented on.

python_src/src/lamp_py/startup/start_ingestion.py
```diff
@@ -119,15 +88,20 @@ def main() -> None:
...
     signal.signal(signal.SIGTERM, handle_ecs_sigterm)
+
+    # configure the environment
     load_environment()
+    os.environ["SERVICE_NAME"] = "ingestion"
     validate_environment()
+
+    # run the main method
     main()
+
```
mzappitello, line 104: created a `start`function that is called in the poetry script. its also called when we run this as a script.

python_src/src/lamp_py/startup/start_performance_manager.py
```diff
@@ -9,48 +9,19 @@
...
-                    continue
-                key, value = line.split("=")
-                logging.info("setting %s to %s", key, value)
-                os.environ[key] = value
-
-    except FileNotFoundError as fnfe:
-        logging.warning("unable to find env file %s", fnfe)
-    except Exception as exception:
-        logging.exception("error while trying to bootstrap")
-        raise exception
```
mzappitello, line 53: collapsed this so many times.

python_src/src/lamp_py/startup/start_performance_manager.py
```diff
@@ -171,10 +142,22 @@ def iteration() -> None:
...
+
+    # configure the environment
     load_environment()
+    os.environ["SERVICE_NAME"] = "performance_manager"
     validate_environment()
-    parsed_args = parse_args(sys.argv[1:])
 
+    # run main method with parsed args
     main(parsed_args)
+
```
mzappitello, line 160: again with the same start method to call from poetry scripts.

.github/workflows/continuous_integration.yaml
```diff
@@ -1,19 +1,19 @@
-name: Performance Manager CI
+name: Continuous Integration
```
mzappitello, line 1: this just does all Continuous Integration now.

.github/workflows/continuous_integration.yaml
```diff
@@ -83,5 +83,5 @@ jobs:
       - uses: actions/checkout@v2
       - uses: ./.github/actions/poetry_deps
         with:
-          poetry-dir: performance_manager
+          poetry-dir: python_src
       - run: poetry run pytest
```
mzappitello, line 87: this test should be changed to only run the tests that we need based on what has changed.

i have made [a ticket](https://app.asana.com/0/0/1204188000578032/f) to do so.

we can create an action that looks at what files have been changed and create bools for what tests we need to run.

docker-compose.yml
```diff
@@ -12,28 +11,30 @@ services:
...
-      context: performance_manager
+      context: ./python_src
       dockerfile: Dockerfile
     depends_on:
       - local_rds
-    working_dir: /performance_manager
+    working_dir: /lamp
     volumes:
       - ~/.aws:/root/.aws:ro # map credentials to be used by boto3, read-only
+    command: [ "poetry", "run", "performance_manager" ]
```
mzappitello, line 27: docker compose now uses a single docker image from our package, and launches different pipelines using the command argument. on the terraform side, we also want to add a task command that will do the same thing.

python_src/Dockerfile
```diff
@@ -5,22 +5,24 @@ ENV PYTHONDONTWRITEBYTECODE 1
...
 RUN apt-get update
 RUN apt-get install -y libpq-dev gcc curl
 
 # Install poetry
 RUN pip install -U pip
 RUN pip install "poetry==1.1.14"
 
-COPY . /performance_manager/
-WORKDIR /performance_manager/
-RUN poetry config virtualenvs.create false
```
mzappitello: we want to create the virtual env now so that we can get `poetry run <pipeline_name>`to work correctly.

python_src/tests/aws/test_s3_utils.py
```diff
@@ -14,18 +14,18 @@
 from botocore.stub import Stubber
 from botocore.stub import ANY
 
-from lib.s3_utils import file_list_from_s3
-from lib.s3_utils import move_s3_objects
+from lamp_py.aws.s3 import file_list_from_s3_ingestion
+from lamp_py.aws.s3 import move_s3_objects
 
-TEST_FILE_DIR = os.path.join(os.path.dirname(__file__), "test_files")
+from ..test_resources import incoming_dir
```
mzappitello, line 20: use this import to get the test files dir. should probably be a fixture, but this was easier to get off the ground.

python_src/tests/ingestion/test_gtfs_rt_converter.py
mzappitello, line 1: the tests in this file were failing on main, but we weren't running them in our CI for some reason. :/

i fixed them using the strategy we discussed of using `gz_to_pyarrow` instead of `process_files`. there is more coverage we could be doing, but again, i just wanted to port over what we had rather than make big code changes.

python_src/tests/ingestion/test_gtfs_rt_converter.py
```diff
@@ -127,7 +131,7 @@ def test_vehicle_positions_file_conversion() -> None:
         "schedule_relationship": (0, "object", "UNSCHEDULED", "ADDED"),
         "start_date": (40, "object", "nan", "nan"),
         "start_time": (87, "object", "nan", "nan"),
-        "trip_id": (426, "object", "nan", "nan"),
+        "trip_id": (0, "object", "nan", "nan"),
```
mzappitello, line 134: we added trip id into what we're parsing, so we have these records now.

python_src/tests/ingestion/test_gtfs_rt_converter.py
```diff
@@ -350,6 +363,7 @@ def test_bus_vehicle_positions_file_conversion() -> None:
         "name": (844, "object", "nan", "nan"),
         "first_name": (844, "object", "nan", "nan"),
         "last_name": (844, "object", "nan", "nan"),
+        "entity_is_deleted": (0, "bool", "False", "False"),
```
mzappitello, line 366: we added this in, so now its present, but we don't find it anywhere i guess?

python_src/tests/ingestion/test_gtfs_rt_converter.py
```diff
@@ -469,7 +486,7 @@ def test_start_of_hour() -> None:
         year=2022, month=4, day=1, tzinfo=timezone.utc
     )
 
-    table = next(converter.process_files())  # type: ignore
+    table = next(converter.process_files())
```
mzappitello, line 489: mypy is actually cool with this?

python_src/tests/performance_manager/test_performance_manager.py
```diff
@@ -65,8 +66,11 @@ def set_env_vars() -> None:
     if int(os.environ.get("BOOTSTRAPPED", 0)) == 1:
         logging.warning("allready bootstrapped")
     else:
-        env_file = os.path.join(HERE, "..", "..", ".env")
+        env_file = os.path.join(
+            os.path.dirname(os.path.abspath(__file__)), "..", "..", "..", ".env"
+        )
```
mzappitello, line 71: we need this when we're running these tests locally on the command line, but _only_ there.

i think there might be a better way.

python_src/tests/test_resources.py
mzappitello, line 1: new file that handles finding testing resources. its fine but should probably be a fixture.

docker-compose.yml
rymarczy: So this PR will be blocked by a devops PR to incorporate a `command` call in terraform

.github/workflows/ingestion-deploy-dev.yaml
```diff
@@ -15,10 +15,10 @@ jobs:
     with:
       app-name: lamp-ingestion
       environment: dev
-      dockerfile-path: ./py_gtfs_rt_ingestion/
+      dockerfile-path: ./python_src/
       cluster: lamp
     secrets:
       aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
       aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
       docker-repo: ${{ secrets.INGESTION_DOCKER_URI }}
```
rymarczy, line 23: Since we're using the same Dockerfile for all applications, should probably use the same `docker-repo` for both apps.

rymarczy:
Minor comment on `file_list_from_s3` functions. Only other issues are changes that would be blocked by PR's to `devops` for infrastructure changes. 

If the infrastructure stuff is pulled out, then should be good to go right now.

PR #99, opened 2023-03-16
GitHub actions update

mzappitello:
Correct the Deploy Script dependencies and add tests on a potential action to only test what we need to test.


.github/workflows/future_continuous_integration.yaml
mzappitello, line 1: this is an outline of how we can figure out what tests we want to run on each pull request based on whats changed. it uses bash commands to parse out what directories have changed and we can us the bools it generates to determine what needs to be run.

.github/workflows/future_continuous_integration.yaml
```diff
@@ -0,0 +1,91 @@
...
+      if: env.TEST_AWS == 'true'
+      run: echo "Running AWS tests..."
+
+    - name: Test Performance Manager
+      if: env.TEST_PERFORMANCE_MANAGER == 'true'
+      run: echo "Running Performance Manager tests..."
+
+    - name: Test Postgres 
+      if: env.TEST_POSTGRES == 'true'
+      run: echo "Running Postgres tests..."
```
mzappitello, line 91: the idea is that this section will eventually have `poetry run pytest tests/aws` or whatever. we'll also standup the postgres server only for the tests that require them.

.github/workflows/ingestion-deploy-dev.yaml
```diff
@@ -2,7 +2,7 @@ name: Deploy Ingestion to Dev
 
 on:
   workflow_run:
-    workflows: ["Ingestion CI"]
+    workflows: ["Continuous Integration"]
```
mzappitello, line 5: this needs to be tied to the ci passing and the main branch.

.github/workflows/future_continuous_integration.yaml
mzappitello: output of this is here https://github.com/mbta/lamp/actions/runs/4441953860

.github/workflows/future_continuous_integration.yaml
rymarczy: I'm not totally sure what all is happening in "Check Changed Subdirectories".

It does look there are some hard coded file names and command parameters that we need to be very aware of, which changes to folder structure in the project. Might be worth adding that information as comments and also maybe some instructions on how to handle the additions of new library folders.

rymarczy:
Main changes to deployment look good. Not too worried about "future_continuous_integration" since it won't be running as our main CI right now.

mzappitello:
i moved the new CI stuff out into its own PR

https://github.com/mbta/lamp/pull/101

i also took the notes here when doing so.

.github/workflows/deploy-dev.yaml
```diff
@@ -0,0 +1,49 @@
+name: Deploy Ingestion to Dev
```
rymarczy, line 1: Should this be a more general name?

rymarczy:
A couple of comments. Fixing the Dockerfile would be the only blocked on this.

rymarczy:
LGTM

PR #100, opened 2023-03-17
Use one impl of file_list_from_s3

mzappitello:
After combining both python packages into one, we were left with a duplicate `file_list_from_s3` functionality that had slightly different signatures (one had a threshold while the other didn't and one returned and iter of string while the other returned a list of them. We now just have on impl that all pipelines share.

* rename `file_list_from_s3_ingestion` as `file_list_from_s3` and use that one everywhere.
* get rid of itr -> list conversions that were surrounding the other impl that returned an itr
* remove uneeded imports from the aws.__init__.py file.

Asana Task: https://app.asana.com/0/0/1204188000578028/f

### DIFF
--- a/python_src/src/lamp_py/aws/__init__.py
+++ b/python_src/src/lamp_py/aws/__init__.py
@@ -2,8 +2,7 @@
 
 from .ecs import check_for_sigterm, handle_ecs_sigterm
 from .s3 import (
-    file_list_from_s3_ingestion,
-    file_list_from_s3_pm,
+    file_list_from_s3,
     get_utc_from_partition_path,
     get_zip_buffer,
     move_s3_objects,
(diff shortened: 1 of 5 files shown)


python_src/src/lamp_py/aws/__init__.py
```diff
@@ -1,15 +1 @@
...
-    file_list_from_s3_pm,
-    get_utc_from_partition_path,
-    get_zip_buffer,
-    move_s3_objects,
-    read_parquet,
-    read_parquet_chunks,
-    write_parquet_file,
-)
-
-__version__ = "0.1.0"
```
rymarczy, line 15: is this `__init__.py` file not needed anymore?

PR #101, opened 2023-03-17
Add a future continuous integration workflow

mzappitello:
This workflow should look at the source files in our lamp_py directory and decide which tests need to be run. Testing it out for now so that it runs on each push, before adding making it _the way_.

Asana Task: https://app.asana.com/0/0/1204188000578032/f


mzappitello:
i am pretty sure this is failing because the action can't be accessed until its in main.

.github/actions/determine_tests/determine_tests.sh
mzappitello, line 1: i decided to do this as a bash script since you can run it in your local shell and use it, making it easier to debug.

i also added a ton of comments.

.github/actions/determine_tests/action.yaml
mzappitello, line 1: new action that determines which tests we need to run. 

i think we might want this to run on future deployments as well?

.github/workflows/future_continuous_integration.yaml
mzappitello, line 1: workflow to test this out on every push.

rymarczy:
I'm nervous about the hard-coded nature of this framework. I think it's fine for now, but a more automatic solution would be nice. 

I think it also might point out a few issues with out folder structure. For example, `import_env.py` and `logging_utils.py` not being in a folder seems like it could be a testing blind spot with this structure. And we should probably having testing coverage on this files eventually. 

Also, is there a material benefit for having the `startup` folder? Seems like those files would make more sense in their application folders.

rymarczy:
LGTM

PR #102, opened 2023-03-20
cleanup python init files in subdirs

mzappitello:
After migrating from multiple packages to a single package, the additional imports in our init files are going to lead us to bad practices, where we're not using the full paths on our imports from sibling directories. Clean those out and clean up imports that were not using the full paths.


rymarczy:
LGTM

PR #103, opened 2023-03-21
Skip directories in S3 file_list

rymarczy:
Returning directories from `file_list_from_s3` causes errors in some parts of performance_manager. 

The original version of `file_list_from_s3` had the same control logic.


mzappitello:
whats the case that this is catching on?

mzappitello:
LGTM

PR #104, opened 2023-03-22
Alembic migrations

rymarczy:
This PR implements using alembic for automatic database migrations.

The structure should work for managing multiple databases, as long as there is one application that can run/control the migrations for each database. 

An initial RDS version is included for the performance manager database in the `migrations/versions/performance_manager` folder. 

All `alembic` commands should be run using the "-n" flag to specify the database target for the migration task. 

After a schema change run the following command to create a new migration version:
`alembic -n db_name revision --autogenerate -m "revision message"`

This will create a new version file in the appropriate database folder for manual review. 

Asana Task: https://app.asana.com/0/1203655790984059/1203144932469998


python_src/src/lamp_py/runtime_utils/import_env.py
```diff
@@ -11,7 +11,7 @@ def load_environment() -> None:
             return
 
         here = os.path.dirname(os.path.abspath(__file__))
-        env_file = os.path.join(here, "..", "..", "..", ".env")
+        env_file = os.path.join(here, "..", "..", "..", "..", ".env")
```
rymarczy, line 14: Just need an extra set of ".." here because this got moved into the `runtime_utils` folder.

python_src/src/lamp_py/runtime_utils/alembic_migration.py
```diff
@@ -0,0 +1,51 @@
...
+        db_manager = DatabaseManager()
+    else:
+        raise NotImplementedError(f"Migration for {db_name} not implemented.")
+
+    # check if alembic_version table exists in rds
+    try:
+        db_manager.select_as_list("SELECT 1 FROM alembic_version")
+        version_table_exists = True
+    except Exception as _:
+        version_table_exists = False
```
rymarczy, line 25: This just checks if the "alembic_version" table exists, if it does not, we assume that that RDS is currently in the state matching the "head" revision.

We create the "alembic_version" table and "stamp" it with the existing head revision.

python_src/src/lamp_py/runtime_utils/alembic_migration.py
```diff
@@ -0,0 +1,51 @@
...
+    else:
+        # if no version table, create version table and stamp with current head
+        # this assumes the head matches current rds state
+        migration_command = [
+            "alembic",
+            "-n",
+            db_name,
+            "stamp",
+            "head",
+        ]
```
rymarczy, line 45: "stamp" alembic_version table with current head version.

python_src/src/lamp_py/migrations/versions/performance_manager/be4c10c548f0_initial_rds_schema.py
```diff
@@ -0,0 +1,222 @@
...
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = "be4c10c548f0"
+down_revision = None
+branch_labels = None
+depends_on = None
+
```
rymarczy, line 17: version file matching state of current RDS schema

python_src/src/lamp_py/migrations/env.py
```diff
@@ -0,0 +1,96 @@
+from logging.config import fileConfig
+
+from alembic import context
+
+from lamp_py.postgres.postgres_utils import get_local_engine
+from lamp_py.runtime_utils.import_env import load_environment
+
```
rymarczy, line 7: This file controls how migrations are run, based on the commands passed to the alembic CLI program.

python_src/src/lamp_py/migrations/env.py
```diff
@@ -0,0 +1,96 @@
...
+# load local environment, if running alembic locally
+load_environment()
```
rymarczy, line 9: Running alembic locally requires the correct environmental variables to be loaded. There might be a better way to handle this?

python_src/src/lamp_py/migrations/env.py
```diff
@@ -0,0 +1,96 @@
...
+
+# this is the Alembic Config object, which provides
+# access to the values within the .ini file in use.
+config = context.config
+
+# gate to make sure alembic is run using -n flag
+if config.config_ini_section == "alembic":
+    raise SyntaxError("Run alembic with -n flag to specifiy Database name.")
+
```
rymarczy, line 18: Running alembic with no "-n" flag results in `config_ini_section` defaulting to `alembic`. In this configuration, we want to avoid this behavior. If we are managing multiple db's this could run migrations for all of them.

python_src/src/lamp_py/migrations/env.py
```diff
@@ -0,0 +1,96 @@
...
+# get database name from -n flag when alembic is run from cmd line
+db_name = config.config_ini_section
+
```
rymarczy, line 21: `db_name` is equivalent to `config_ini_section` this is the valued passed into alembic with "-n" flag.

python_src/src/lamp_py/migrations/env.py
```diff
@@ -0,0 +1,96 @@
...
+#     url = config.get_main_option("sqlalchemy.url")
+#     context.configure(
+#         url=url,
+#         target_metadata=target_metadata,
+#         literal_binds=True,
+#         dialect_opts={"paramstyle": "named"},
+#     )
+
+#     with context.begin_transaction():
+#         context.run_migrations()
```
mzappitello, line 70: if we aren't going to use this, i think we should remove it.

if we are, it could use a comment for the context of where we will.

python_src/src/lamp_py/runtime_utils/alembic_migration.py
```diff
@@ -0,0 +1,51 @@
...
+
+    result = subprocess.run(
+        migration_command, capture_output=True, text=True, check=True
+    )
+    logging.info(result.stdout)
+    logging.info(result.stderr)
```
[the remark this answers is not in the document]
mzappitello: i think it would be better to just run alembic inside of our python environment rather than spinning up another process to run this. it feels like that should be possible since alembic is a python process.

mzappitello (commented):
i like the look of this. lets have a call tomorrow to go over how we want to test this on our infrastructure and to go over a few questions.

python_src/src/lamp_py/runtime_utils/process_logger.py
```diff
@@ -28,6 +28,8 @@ def __init__(self, process_name: str, **metadata: MdValues) -> None:
         create a process logger with a name and optional metadata. a start time
         and uuid will be created for timing and unique identification
         """
+        logging.getLogger().setLevel("INFO")
+
```
rymarczy, line 32: makes sure logging module is using root logger

python_src/src/lamp_py/runtime_utils/alembic_migration.py
```diff
@@ -0,0 +1,39 @@
...
+    if db_name == "performance_manager":
+        # if database is clean, call to DatabaseManager will trigger
+        # SqlBase.metadata.create_all(self.engine)
+        # creating all database tables
+        db_manager = DatabaseManager()
+    else:
+        raise NotImplementedError(f"Migration for {db_name} not implemented.")
+
+    # load alembic configuation for db_name
+    alembic_cfg = Config(alembic_cfg_file, ini_section=db_name)
```
rymarczy, line 28: `ini_section` correlates to `-n` flag in CLI

python_src/alembic.ini
```diff
@@ -0,0 +1,117 @@
...
+
+[handler_console]
+class = StreamHandler
+args = (sys.stderr,)
+level = NOTSET
+formatter = generic
+
+[formatter_generic]
+#format = %(levelname)-5.5s [%(name)s] %(message)s
+format = %(levelname)s:%(name)s:%(message)s
```
rymarczy, line 116: match logging format to our default root logger

python_src/src/lamp_py/runtime_utils/alembic_migration.py
```diff
@@ -0,0 +1,39 @@
...
+    # check if alembic_version table exists in rds
+    try:
+        db_manager.select_as_list("SELECT 1 FROM alembic_version")
+    except Exception as _:
+        # if no version table, create version table and stamp with current head
+        # this assumes the head matches current rds state
+        command.stamp(alembic_cfg, revision="head")
+
+    # normal migration command, for if version table exists
+    command.upgrade(alembic_cfg, revision="head")
```
mzappitello, line 39: do we mean to run `ùpgrade` after running `stamp` in the event that we don't have an alembic version table?

PR #105, opened 2023-03-27
Add script to seed database for local testing

mzappitello:
We've been seeding our metadata table locally when launching the performance manager pipeline. The run script would take an optional bools to seed, clear static data, and clear real time data. These booleans would be passed to the DatabaseManager during initialization to truncate the tables and add new paths to the metadata table.

Instead, remove that functionality from the DatabaseManager. A script has been written that can seed metadata from a json file and clear out existing data from command line flags.

* remove seed and clear bools from DatabaseManager.__init__
* remove methods for adding metadata from hardcoded path and clearing out tables.
* write script to handle all of this locally.

Asana Task: https://app.asana.com/0/0/1204188000578034/f


python_src/pyproject.toml
```diff
@@ -3,14 +3,15 @@ name = "lamp_py"
...
-    "Mike Zappitello <mzappitello@mbta.com>",
+  "MBTA CTD <developer@mbta.com>",
+  "Ryan Rymarczyk <rrymarczyk@mbta.com>",
+  "Mike Zappitello <mzappitello@mbta.com>",
 ]
 
 [tool.poetry.scripts]
 ingestion = 'lamp_py.ingestion.pipeline:start'
 performance_manager = 'lamp_py.performance_manager.pipeline:start'
+seed_metadata = 'lamp_py.postgres.seed_metadata:run'
```
mzappitello, line 14: add this script so you can run `poetry run seed_metadata --env <env path> --seed-file <path to seed file>`

python_src/src/lamp_py/performance_manager/pipeline.py
```diff
@@ -57,48 +57,20 @@ def validate_environment() -> None:
 def parse_args(args: List[str]) -> argparse.Namespace:
     """parse args for running this entrypoint script"""
     parser = argparse.ArgumentParser(description=DESCRIPTION)
-    parser.add_argument(
-        "--experimental",
-        action="store_true",
-        dest="experimental",
-        help="if set, use a sqllite engine for quicker development",
-    )
-
```
mzappitello, line 66: remove these from the performance manager service. we don't want that service to be in charge of seeding the metadata table so that we can do it locally for testing. this functionality is now in a script we can run.

python_src/src/lamp_py/postgres/postgres_utils.py
```diff
@@ -201,13 +195,7 @@ class DatabaseManager:
...
     """
 
-    def __init__(
-        self,
-        verbose: bool = False,
-        seed: bool = False,
-        clear_rt: bool = False,
-        clear_static: bool = False,
-    ):
+    def __init__(self, verbose: bool = False):
```
mzappitello, line 198: database manager no longer initializes by modifying the db its connecting to. that is handled by another script that uses its functionality.

python_src/src/lamp_py/postgres/postgres_utils.py
```diff
@@ -17,13 +16,8 @@
 from .postgres_schema import (
     MetadataLog,
     SqlBase,
-    VehicleEventMetrics,
-    VehicleEvents,
-    VehicleTrips,
```
mzappitello, line 22: no longer need access to these tables.

python_src/src/lamp_py/postgres/postgres_utils.py
```diff
@@ -17,13 +16,8 @@
...
 )
 
-HERE = os.path.dirname(os.path.abspath(__file__))
```
mzappitello, line 25: no longer need a path to this file. it was used to derive a relative path to the seed file.

python_src/src/lamp_py/postgres/postgres_utils.py
```diff
@@ -338,55 +321,6 @@ def add_metadata_paths(self, paths: List[str]) -> None:
...
-                    process_fail=False,
-                )
-                .where(MetadataLog.path.like("%RT_%"))
-            )
-            self.execute(reset_rt_processed)
-
-        if clear_static:
-            SqlBase.metadata.drop_all(self.engine)
-            SqlBase.metadata.create_all(self.engine)
-
```
mzappitello, line 324: this is all handled in the script now. it makes the db manager _more_ agnostic of what its connecting to. (it does still have a function to get unprocessed files, which requires some info about the metadata table.)

python_src/src/lamp_py/postgres/seed_metadata.py
mzappitello, line 1: new script. pretty straight forward. i took a lot of stuff from what was in the database manager before.

python_src/src/lamp_py/postgres/seed_metadata.py
```diff
@@ -0,0 +1,116 @@
...
+def parse_args(args: List[str]) -> argparse.Namespace:
+    """parse args for running this entrypoint script"""
+    parser = argparse.ArgumentParser(description=DESCRIPTION)
+
+    parser.add_argument(
+        "--env-file",
+        dest="env_file",
+        help="environment file containing rds connection parameters",
+        required=True,
+    )
```
mzappitello, line 36: i kind of want to make this a flag we pass into the other argparses and get rid of our code that tries to find it that we only use when running it on the command line. it feels cleaner to me.

python_src/src/lamp_py/runtime_utils/import_env.py
```diff
@@ -1,18 +1,21 @@
 import os
 import logging
+from typing import Optional
 
 
-def load_environment() -> None:
+def load_environment(env_file: Optional[str] = None) -> None:
```
mzappitello, line 6: optionally take a path to the env file. if not, derive it using relative paths.

mzappitello (commented):
pretty straight forward i think. i prefer having this code live in a local script than being something that is potentially executed on a production machine.

python_src/src/lamp_py/postgres/seed_metadata.py
```diff
@@ -0,0 +1,116 @@
...
+def run() -> None:
+    """Run The RDS Interaction Script"""
+    parsed_args = parse_args(sys.argv[1:])
+
+    load_environment(parsed_args.env_file)
+
+    db_manager = DatabaseManager(parsed_args.verbose)
+
+    if parsed_args.seed_file:
+        seed_metadata(db_manager, parsed_args.seed_file)
```
rymarczy, line 95: This needs to be after the clear table calls.

python_src/src/lamp_py/postgres/seed_metadata.py
rymarczy, line 36: I don't imagine I'll be running this often from the command line, docker hostname resolution is a litter different between platforms and my preference would be to have this run automatically in the container with docker compose. With that in mind, the `--env-file` flag would have to not be required.

python_src/src/lamp_py/postgres/seed_metadata.py
```diff
@@ -0,0 +1,116 @@
...
+        action="store_true",
+        dest="verbose",
+        help="if set, use debug logging",
+    )
+
+    parser.add_argument(
+        "--seed-file",
+        dest="seed_file",
+        help="if set, read a json file to seed the metadata table with",
+    )
```
rymarczy, line 49: Would be nice if this defaulted to to the existing json file, since we don't have more than one to seed with.

python_src/src/lamp_py/postgres/seed_metadata.py
```diff
@@ -0,0 +1,116 @@
...
+    seed metadata table for dev environment
+    """
+    logging.info("Seeding Metadata From File %s", seed_file)
+
+    try:
+        with open(seed_file, "r", encoding="utf8") as seed_json:
+            paths = json.load(seed_json)
+
+        db_manager.add_metadata_paths(paths)
+
```
rymarczy, line 79: Probably makes sense to move this `add_metadata_paths` function out of `DatabaseManager` if this is the only part of the application using that functionality, and it's not really generic to DB interaction.

rymarczy:
Few comments, several of them a driven by the fact that we use this functionality differently in our local environments. It seems like this is mainly due to Docker hostname resolution not being platform agnostic. 

I don't need to see `add_metadata_paths` moved as part of this PR, but it reasoning is similar to why `seed_metadata` and `reset_tables` were taken out of `DatabaseManager`.

docker-compose.yml
```diff
@@ -36,3 +36,20 @@ services:
...
+      - local_rds
+    working_dir: /lamp
+    volumes:
+      # map credentials to be used by boto3, read-only
+      - ~/.aws:/root/.aws:ro
+      # add in july 17 filepahs json that will be the default seed file path
+      - ./python_src/tests/test_files/july_17_filepaths.json:/july_17_filepaths.json
+    # entrypoint passes in seed file thats added as a volume. if you want to use a different filepath run
+    # docker-compose run -v /path/to/files.json:/seed.json seed_metadata --seed-file /seed.json
+    entrypoint: [ "poetry", "run", "seed_metadata", "--seed-file", "/july_17_filepaths.json"]
```
mzappitello, line 55: added this as a docker task that can be run.

`docker-compose run seed_metadata` should do it. as requested, it has a default seed file to it now.

python_src/src/lamp_py/postgres/seed_metadata.py
mzappitello: i'm handling this in the docker compose file, since i'm adding the test files into the image with a volume. i'm hoping to eventually take the tests directory out of our image.

python_src/src/lamp_py/postgres/seed_metadata.py
mzappitello: i left this in because some tests were using this as well, although i agree it would be a good idea to get rid of it from this file eventually.

PR #107, opened 2023-03-31
Fix metrics - first stop of trip

rymarczy:
The `vehicle_event_metrics` table is missing records for the first stop of all trips. 

The first stop of a trip should never have a `travel_time` value and a WHERE clause in the `update_insert_query` select statement was specifically excluding these values, to avoid metrics with negative `travel_time` values. 

Asana Task: https://app.asana.com/0/1202087614510178/1204283360272396

### DIFF
--- a/python_src/src/lamp_py/performance_manager/l1_rt_metrics.py
+++ b/python_src/src/lamp_py/performance_manager/l1_rt_metrics.py
@@ -695,33 +695,27 @@ def process_metrics_table(
         )
         .join(
             dwell_times_cte,
-            sa.and_(
-                travel_times_cte.c.trip_stop_hash
-                == dwell_times_cte.c.trip_stop_hash,
-            ),
+            travel_times_cte.c.trip_stop_hash
+            == dwell_times_cte.c.trip_stop_hash,
             full=True,
         )
         .join(
             headways_branch_cte,
-            sa.and_(
-                sa.func.coalesce(
-                    travel_times_cte.c.trip_stop_hash,
-                    dwell_times_cte.c.trip_stop_hash,
-                )
-                == headways_branch_cte.c.trip_stop_hash,
-            ),
+            sa.func.coalesce(
+                travel_times_cte.c.trip_stop_hash,
+                dwell_times_cte.c.trip_stop_hash,
+            )
+            == headways_branch_cte.c.trip_stop_hash,
             full=True,
         )
         .join(
             headways_trunk_cte,
-            sa.and_(
-                sa.func.coalesce(
-                    travel_times_cte.c.trip_stop_hash,
-                    dwell_times_cte.c.trip_stop_hash,
-                    headways_branch_cte.c.trip_stop_hash,
-                )
-                == headways_trunk_cte.c.trip_stop_hash,
-            ),
+            sa.func.coalesce(
+                travel_times_cte.c.trip_stop_hash,
+                dwell_times_cte.c.trip_stop_hash,
+                headways_branch_cte.c.trip_stop_hash,
+            )
+            == headways_trunk_cte.c.trip_stop_hash,
             full=True,
         )
     ).cte(name="all_new_metrics")
--- a/python_src/src/lamp_py/performance_manager/l1_rt_metrics.py
+++ b/python_src/src/lamp_py/performance_manager/l1_rt_metrics.py
@@ -438,9 +438,6 @@ def process_trips_table(
     process_logger.log_complete()
 
 
-# pylint: enable=R0914
-
-
 def process_metrics_table(
     db_manager: DatabaseManager,
     seed_start_date: int,
@@ -537,13 +534,14 @@ def process_metrics_table(
     )
 
     # limit dwell times caluclations to NON-NULL positivie integers
-    # would be nice if this could be done in the first CTE, but I can't 
+    # would be nice if this could be done in the first CTE, but I can't
     # get it to work with sqlalchemy
     dwell_times_cte = (
         sa.select(
             t_dwell_times_cte.c.trip_stop_hash,
             t_dwell_times_cte.c.dwell_time_seconds,
-        ).where(
+        )
+        .where(
             t_dwell_times_cte.c.dwell_time_seconds.is_not(None),
             t_dwell_times_cte.c.dwell_time_seconds > 0,
         )
@@ -604,7 +602,7 @@ def process_metrics_table(
     )
 
     # limit headways branch calculations to NON-NULL positive integers
-    # would be nice if this could be done in the first CTE, but I can't 
+    # would be nice if this could be done in the first CTE, but I can't
     # get it to work with sqlalchemy
     headways_branch_cte = (
         sa.select(
@@ -661,7 +659,7 @@ def process_metrics_table(
     )
 
     # limit headways trunk calculations to NON-NULL positive integers
-    # would be nice if this could be done in the first CTE, but I can't 
+    # would be nice if this could be done in the first CTE, but I can't
     # get it to work with sqlalchemy
     headways_trunk_cte = (
         sa.select(
@@ -831,6 +829,9 @@ def process_metrics_table(
         )
 
     process_logger.log_complete()
+
+
+# pylint: enable=R0914
 
 
 def process_trips_and_metrics(db_manager: DatabaseManager) -> None:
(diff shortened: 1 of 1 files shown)


python_src/src/lamp_py/performance_manager/l1_rt_metrics.py
```diff
@@ -469,6 +467,8 @@ def process_metrics_table(
         )
         .where(
             trips_for_metrics.c.first_stop_flag == sa.false(),
+            trips_for_metrics.c.stop_timestamp.is_not(None),
+            trips_for_metrics.c.move_timestamp.is_not(None),
```
rymarczy, line 471: Remove null results from travel_times

python_src/src/lamp_py/performance_manager/l1_rt_metrics.py
```diff
@@ -530,6 +530,21 @@ def process_metrics_table(
...
+    # get it to work with sqlalchemy
+    dwell_times_cte = (
+        sa.select(
+            t_dwell_times_cte.c.trip_stop_hash,
+            t_dwell_times_cte.c.dwell_time_seconds,
+        )
+        .where(
+            t_dwell_times_cte.c.dwell_time_seconds.is_not(None),
+            t_dwell_times_cte.c.dwell_time_seconds > 0,
+        )
```
rymarczy, line 547: Explicitly remove NULL and negative values from dwell_time calculations

python_src/src/lamp_py/performance_manager/l1_rt_metrics.py
```diff
@@ -576,6 +591,34 @@ def process_metrics_table(
...
+        sa.select(
+            t_headways_branch_cte.c.trip_stop_hash,
+            t_headways_branch_cte.c.headway_branch_seconds,
+        )
+        .where(
+            t_headways_branch_cte.c.headway_branch_seconds.is_not(None),
+            t_headways_branch_cte.c.headway_branch_seconds > 0,
+        )
+        .cte(name="headways_branch")
+    )
```
rymarczy, line 617: Explicitly remove NULL and negative values from branch headway calculations

python_src/src/lamp_py/performance_manager/l1_rt_metrics.py
```diff
@@ -610,26 +653,42 @@ def process_metrics_table(
...
+        sa.select(
+            t_headways_trunk_cte.c.trip_stop_hash,
+            t_headways_trunk_cte.c.headway_trunk_seconds,
         )
-        .cte(name="headways")
+        .where(
+            t_headways_trunk_cte.c.headway_trunk_seconds.is_not(None),
+            t_headways_trunk_cte.c.headway_trunk_seconds > 0,
+        )
+        .cte(name="headways_trunk")
```
rymarczy, line 673: Explicitly remove NULL and negative values from trunk headway calculations

python_src/src/lamp_py/performance_manager/l1_rt_metrics.py
```diff
@@ -639,34 +698,31 @@ def process_metrics_table(
...
+                == headways_branch_cte.c.trip_stop_hash,
             ),
             full=True,
         )
-        .where(
-            sa.or_(
-                sa.func.coalesce(travel_times_cte.c.travel_time_seconds, 0) > 0,
-                sa.func.coalesce(dwell_times_cte.c.dwell_time_seconds, 0) > 0,
-                sa.func.coalesce(headways_cte.c.headway_branch_seconds, 0) > 0,
-                sa.func.coalesce(headways_cte.c.headway_trunk_seconds, 0) > 0,
```
rymarczy, line 668: Remove calculation limitations from `all_new_metrics` CTE, just join result sets from `travel_times`, `dwell_times`, and `headways`

rymarczy:
This PR makes some the the CTE statements more verbose and then makes the join queries less complicated. 

There are likely some opportunities for future optimization, but our current priority is to provide the most correct business logic.

python_src/src/lamp_py/performance_manager/l1_rt_metrics.py
```diff
@@ -639,34 +698,31 @@ def process_metrics_table(
...
                     dwell_times_cte.c.trip_stop_hash,
                 )
-                == headways_cte.c.trip_stop_hash,
-                sa.or_(
-                    sa.func.coalesce(headways_cte.c.headway_branch_seconds, 0)
-                    > 0,
-                    sa.func.coalesce(headways_cte.c.headway_trunk_seconds, 0)
-                    > 0,
-                ),
+                == headways_branch_cte.c.trip_stop_hash,
```
mzappitello, line 711: why is this in a `sa.and_` block. it looks like a single boolean to me?

python_src/src/lamp_py/performance_manager/l1_rt_metrics.py
```diff
@@ -639,34 +698,31 @@ def process_metrics_table(
...
         )
-        .where(
-            sa.or_(
-                sa.func.coalesce(travel_times_cte.c.travel_time_seconds, 0) > 0,
-                sa.func.coalesce(dwell_times_cte.c.dwell_time_seconds, 0) > 0,
-                sa.func.coalesce(headways_cte.c.headway_branch_seconds, 0) > 0,
-                sa.func.coalesce(headways_cte.c.headway_trunk_seconds, 0) > 0,
+        .join(
+            headways_trunk_cte,
+            sa.and_(
```
mzappitello, line 717: same question about the `sa.and_`

python_src/src/lamp_py/performance_manager/l1_rt_metrics.py
mzappitello: this is the where clause that was causing the issue?

rymarczy: No, this one wasn't causing the specific issue we're fixing. But this join logic becomes unnecessary if the input sets don't have NULL or negative values.

python_src/src/lamp_py/performance_manager/l1_rt_metrics.py
rymarczy: Yes, these and_ statements are unnecessary, I'll get rid of them.

python_src/src/lamp_py/performance_manager/l1_rt_metrics.py
rymarczy: Will remove, same as other one.

python_src/src/lamp_py/performance_manager/l1_rt_metrics.py
```diff
@@ -707,7 +757,6 @@ def process_metrics_table(
             == VehicleEventMetrics.trip_stop_hash,
             isouter=True,
         )
-        .where(all_new_metrics.c.travel_time_seconds > 0)
```
mzappitello, line 710: so with this where clause, we were dropping out rows that had a `Null` travel time because this evaluated to False, causing us to loose headway info?

rymarczy: Yes, it would drop the entire metrics records, if the travel time was negative or null.

mzappitello:
i tested this out locally using our seeding dataset.

i found all of the vehicle event metrics that had a single direction for a route_id / parent station pairing and found three such pairs.

1. Green-B place-lake
2. Green-D place-spmnl
3. Mattapan place-matt

then i looked into the data around these in the vehicle events table and found that the Green-B and Mattapan routes don't have any data for the other direction at that station, so its a flaw in the data, not our logic.

the Green-D vehicle events had three events going in the opposite direction from the ones in the vehicle event metrics table, but for all three, they were the only events associated with that trip, which we discard. 

so now i'm not seeing the behavior, and the logic as explained in the comments checks out.

LGTM

please squash all of the commits into a nice one before merging.

mzappitello:
I realized this check wasn't going to be sufficient so i ran some other ones.

I have a query that gets all of the unique route / station / direction combos in the events table and subtracts out the combos that appear in the metrics table, leaving the combinations that are only in the events table. a second query looks at these combinations and checks to see if they have unique trip hashes (i.e. the event is associated with just a single stop)

before this branch, we had 21 such stops
Screen Shot 2023-04-04 at 10 22 55 PM

these errors are primarily happening at the ends of each line, the place that these code changes are going to fix.

running these queries after building with the changes, we see nearly all of them drop away.
Screen Shot 2023-04-04 at 10 35 54 PM

of the two remaining combinations, the Green-D one seems to be single stop trips that we drop when computing our metrics table. the other is a Green-C line having an issue at Kenmore, where the Green Line branches into the B, C, and D branches. i haven't dug into that to figure out whats going on, but i think its safe to say its a different issue than the one solved here.

still LGTM.

PR #109, opened 2023-04-05
Remove BUS events from GTFS Static Processing

rymarczy:
Currently, performance manager processes ALL GTFS static events, loading them into the RDS. Most of these events are BUS related and can be removed from the performance manager pipeline.

This PR modifies the RDS loading pipeline to remove all BUS events for GTFS Static processing.

Asana Task: https://app.asana.com/0/1189492770004753/1204339091106883

### DIFF
--- a/python_src/src/lamp_py/performance_manager/l0_gtfs_static_table.py
+++ b/python_src/src/lamp_py/performance_manager/l0_gtfs_static_table.py
@@ -1,7 +1,7 @@
 import os
 import pathlib
 from dataclasses import dataclass
-from typing import List, Optional
+from typing import List, Optional, Dict
 
 import numpy
 import pandas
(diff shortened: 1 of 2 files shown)


python_src/src/lamp_py/performance_manager/l0_gtfs_static_table.py
```diff
@@ -240,14 +240,57 @@ def transform_data_tables(static_tables: List[StaticTableDetails]) -> None:
...
+        "gtfs.remove_bus_records",
+        stop_times_start_row_count=static_tables["stop_times"].data_table.shape[
+            0
+        ],
+    )
+    process_logger.log_start()
+
+    # remove bus routes (route_type == 3) from routes table
+    routes = static_tables["routes"].data_table
+    routes = routes[(routes["route_type"] != 3)]
```
mzappitello, line 257: so we're keeping in the commuter rail?

mzappitello (approved):
LGTM.

do we want to deploy this at the same time as the RT files to avoid having RT files be unable to lookup static data?

rymarczy:
Yes, will wait to deploy the RT fix first to avoid any conflicts.

PR #112, opened 2023-04-10
Add in API Elixir Application

mzappitello:
This application was created using the Phoenix framework. It currently is configured to connect to a local running database we create via docker compose. The app is able to connect to the database and run queries on the metadata log table. An endpoint was created to access information from those queries and return the data as a json array. There is the beginning of a testing framework that connects to the db and checks that the seeded metadata is there.

Lots more to do but you gotta start somewhere.

Asana Task: https://app.asana.com/0/1189492770004753/1204283360272405/f


api/assets/js/app.js
mzappitello, line 1: i'm not sure how much of this boiler plate UI we should keep around. i'm currently just making requests via a PostMan like client in VS Code.

api/config/dev.exs
```diff
@@ -0,0 +1,66 @@
+import Config
+
+# Configure your database
+config :api, Api.Repo,
+  username: "postgres",
+  password: "postgres",
+  hostname: "localhost",
+  database: "performance_manager",
+  port: "5432",
```
mzappitello, line 9: this is all the default config minus this bit for connectiong to our local database created by the docker-compose file at root.

it would be nice if this could be created directly from the .env file also at root.

api/config/test.exs
```diff
@@ -0,0 +1,28 @@
...
+#
+# The MIX_TEST_PARTITION environment variable can be used
+# to provide built-in test partitioning in CI environment.
+# Run `mix help test` for more information.
+config :api, Api.Repo,
+  username: "postgres",
+  password: "postgres",
+  hostname: "localhost",
+  database: "performance_manager",
+  port: "5432",
```
mzappitello, line 13: connect to the DB created by docker-compose for testing.

the assumption is that we're running some scripts using the py package first and then accessing the db it modifies to run tests.

potentially all of this gets orchestrated via docker-compose at some point.

api/lib/api/performance_manager.ex
```diff
@@ -0,0 +1,40 @@
...
+  ## Examples
+
+      iex> get_metadata_log!(123)
+      %MetadataLog{}
+
+      iex> get_metadata_log!(456)
+      ** (Ecto.NoResultsError)
+
+  """
+  def get_metadata_log!(pk_id), do: Repo.get!(MetadataLog, pk_id)
```
mzappitello, line 38: only functions here right now are to get all of the metadata log table or a just a single row from it.

my understanding is that this section gets filled out quiet a bit more once we bring in the other tables. potentially, this is the place where more complex join queries are also stored?

api/lib/api/performance_manager/metadata_log.ex
mzappitello, line 1: currently this is just the metadata log table. i'm not sure if it makes more sense to describe the entire schema in this one file or to have a file per table. in the python project we do the former.

api/lib/api_web/router.ex
```diff
@@ -0,0 +1,28 @@
+defmodule ApiWeb.Router do
+  use ApiWeb, :router
+
+  pipeline :api do
+    plug :accepts, ["json"]
+  end
+
+  scope "/api", ApiWeb do
+    pipe_through :api
+    get "/metadata_log", MetadataLogController, :index
```
mzappitello, line 10: add to the API to get the metadata log information

api/test/api/performance_manager_test.exs
```diff
@@ -0,0 +1,23 @@
...
+    end
+
+    test "get_metadata_log!/1 returns the metadata_log with given id" do
+      first_entry = PerformanceManager.get_metadata_log!(1)
+
+      assert first_entry.processed == false
+
+      assert first_entry.path ==
+               "mbta-ctd-dataplatform-dev-springboard/lamp/RT_VEHICLE_POSITIONS/year=2022/month=7/day=20/hour=0/6e247daf3469436aa574f93fa4cb7c48-0.parquet"
+    end
```
mzappitello, line 21: test assumes blank db with seeded metadata using the `docker-compose run seed_metadata`

mzappitello (commented):
This is largely the out of the box pheonix app with a few pieces for connecting to a local database, grabbing rows out of one of its tables, and echoing them back via api request.

I feel like a lot of this is boilerplate that we could discard, but I'm uncertain about which pieces we would want to keep around since we'll be expanding on them later.

Any guidance would be appreciated.

api/config/test.exs
```diff
@@ -0,0 +1,28 @@
...
+# Configure your database
+#
+# The MIX_TEST_PARTITION environment variable can be used
+# to provide built-in test partitioning in CI environment.
+# Run `mix help test` for more information.
+config :api, Api.Repo,
+  username: "postgres",
+  password: "postgres",
+  hostname: "localhost",
+  database: "performance_manager",
```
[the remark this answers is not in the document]
mzappitello: i guess i'm still a little unclear on how the testing against a db is going to work in practice since its going to be seeded from another source. i would imagine we would keep the db name the same rather than have the process that generates it use a different one?

mzappitello:
i think i would prefer to leave the env stuff as it is for the initial commit and then adjust it while completing the env var [task](https://app.asana.com/0/1189492770004753/1204390911340914/f)

PR #114, opened 2023-04-13
ci: include CI for the Elixir app

paulswartz:


.github/workflows/elixir.yaml
```diff
@@ -0,0 +1,114 @@
...
+      - 'api/**'
+      - '.github/workflows/elixir.yaml'
+  pull_request:
+    paths:
+      - 'api/**'
+      - '.github/workflows/elixir.yaml'
+
+concurrency:
+  group: ci-${{ github.ref }}
+  cancel-in-progress: true
```
mzappitello, line 16: what does this block in the workflow do?

mzappitello (commented):
One question for my own benefit.

I would also appreciate changing the Asana workflow to not break if you're missing a key. I think skipping it entirely if it can't find the key is fine there.

.github/workflows/elixir.yaml
paulswartz: It limits the workflow to one concurrent run per branch. If an update to the branch is pushed while CI is running, the in-progress CI run is cancelled.

PR #115, opened 2023-04-13
Inline Trips & Metrics Event Processing

rymarczy:
Our Trips and Metrics processing was added after our revised RT events processing pipeline was implemented in our development environment. For that reason, the original Trip and Metrics processing implementation worked by processing batches of records so it could catch up to processing current RT records. 

Now that events process and Trips/Metrics processing is in sync, it is possible to move the processing of Trips & Metrics processing in-line with RT events. 

In this updated flow, new Trips records will be loaded based on the RT events produced in `l0_gtfs_rt_events.py`. This paves the way for removing all Trips related information from the `VehicleEvents` schema and reducing our RDS storage size. 

Metrics processing is also updated to only process the `start_date` values existing in the RT events dataframe. The records still must be processed in a batch format, since current metrics rely on comparing values to RT events from outside of the current dataframe window. 

Asana Task: https://app.asana.com/0/1204389777951216/1204119636952108

PR #117, opened 2023-04-19
Fix Creation of `fk_static_timestamp` in vehicle_trips table

rymarczy:
PR #115 added the `fk_static_timstamp` column to `vehicle_trips` as a non-nullable column. However, since data already exists in the `vehicle_trips` table on the DEV RDS, creating the column automatically creates all NULL records.

This PR updates the alembic update logic to add `fk_static_timestamp` as a nullable column, populate the column from data in the vehicle_events table, and then return `fk_static_timestamp` to a non-nullable column.


mzappitello (approved):
this looks good to me.

PR #118, opened 2023-04-20
ECS Logging Fix

rymarczy:
This PR adds more verbose logging to our AWS ECS sigterm handling utils. 

Also includes fix to `_rds_writer_process` logging for overwrite of `process_logger` variable name. 

Asana Task: https://app.asana.com/0/1204389777951216/1204442411295589


python_src/src/lamp_py/postgres/postgres_utils.py
mzappitello, line 1: can we make the process logger outside of the while loop its own thing?

each metadata insert can be a process we log and the entire _rds_writer_process can be its own as well. that way the uuid's will match up a little cleaner.

mzappitello (changes requested):
one small change and then LGTM.

mzappitello (approved):

PR #120, opened 2023-04-25
RDS Triggers for `version_pub_date` and Static Trips trunk/branch designations

rymarczy:
Feat: Add `version_pub_date` column to `static_feed_info` table

LAMP will begin using new GTFS static schedule versions on the day they are issued. 

The `feed_version` field of all GTFS static `feed_info` tables is supposed to contain an ISO UTC timestamp representing the issued datetime of the schedule. This datetime will be extracted (with an RDS trigger function) from `feed_version`, converted to the Eastern timezone service date, and inserted as an integer of `YYYYMMDD` representing the service date the schedule version should be active. 

Feat: Add branch & trunk route id processing to the `static_trips` table

In preparation for the per-caculation of scheduled metrics, `branch_route_id` and `trunk_route_id` columns are being added to the `static_trips` table. These columns will be populated with an RDS trigger function on INSERT. 

Currently only the Green and Red Line routes will have `branch_route_id` values, with all other routes being `null`. 

Green line `route_id`'s will having their alpha designation dropped (-B, -C, -D, -E) for a `trunk_route_id` value of `Green` 

Red Line trips containing stops at Savin Hill, Fields Corner, Shawmut or Ashmont will be tagged as `Red-Ashmont` for `branch_route_id`. Trips containing stops at North Quincy, Wollaston, Quincy Center, Quincy Adams or Braintree will be tagged as `Red-Braintree` for `branch_route_id`. Otherwise, `branch_route_id` will be null and `trunk_route_id` will always be `Red`. 

Asana Task: https://app.asana.com/0/1204389777951216/1204449669814305
Asana Task: https://app.asana.com/0/1204389777951216/1204449669814306

### DIFF
--- a/python_src/src/lamp_py/migrations/versions/performance_manager/43153d536c2a_feed_info_init_date_trigger.py
+++ b/python_src/src/lamp_py/migrations/versions/performance_manager/43153d536c2a_feed_info_init_date_trigger.py
@@ -1,0 +1,67 @@
+"""feed_info init_date trigger
+
+Revision ID: 43153d536c2a
+Revises: 98aa70293578
+Create Date: 2023-04-25 06:53:09.672206
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = "43153d536c2a"
+down_revision = "98aa70293578"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "static_feed_info",
+        sa.Column("version_pub_date", sa.Integer(), nullable=True),
+    )
+
+    op.create_index(
+        op.f("ix_static_feed_info_version_pub_date"),
+        "static_feed_info",
+        ["version_pub_date"],
+        unique=False,
+    )
+
+    create_feed_info_insert_function = """
+        CREATE OR REPLACE FUNCTION insert_feed_info() RETURNS TRIGGER AS $$ 
+        BEGIN 
+            IF NEW.feed_version ~ '\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\+\d{2}:\d{2}' THEN
+                NEW.version_pub_date := replace((substring(NEW.feed_version from '\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\+\d{2}:\d{2}')::timestamptz at time zone 'US/Eastern')::date::text,'-','')::integer;
+            ELSE
+                NEW.version_pub_date := NEW.feed_start_date;
+            END IF;
+            RETURN NEW;
+        END;
+        $$ LANGUAGE plpgsql;
+    """
+    op.execute(create_feed_info_insert_function)
+
+    create_trigger = """
+        CREATE TRIGGER insert_into_feed_info BEFORE INSERT ON static_feed_info 
+        FOR EACH ROW EXECUTE PROCEDURE insert_feed_info();
+    """
+    op.execute(create_trigger)
+
+
+def downgrade() -> None:
+    drop_trigger = (
+        "DROP TRIGGER IF EXISTS insert_into_feed_info ON static_feed_info;"
+    )
+    op.execute(drop_trigger)
+
+    drop_function = "DROP function IF EXISTS public.insert_feed_info();"
+    op.execute(drop_function)
+
+    op.drop_index(
+        op.f("ix_static_feed_info_version_pub_date"),
+        table_name="static_feed_info",
+    )
+
+    op.drop_column("static_feed_info", "version_pub_date")
--- a/python_src/src/lamp_py/migrations/versions/performance_manager/43153d536c2a_feed_info_init_date_trigger.py
+++ b/python_src/src/lamp_py/migrations/versions/performance_manager/43153d536c2a_feed_info_init_date_trigger.py
@@ -1,4 +1,4 @@
-"""feed_info init_date trigger
+"""feed_info init_date triggerfeed_active_date
 
 Revision ID: 43153d536c2a
 Revises: 98aa70293578
@@ -19,13 +19,29 @@ depends_on = None
 def upgrade() -> None:
     op.add_column(
         "static_feed_info",
-        sa.Column("version_pub_date", sa.Integer(), nullable=True),
+        sa.Column("feed_active_date", sa.Integer(), nullable=True),
+    )
+
+    update_feed_active_date_query = """
+        UPDATE 
+            static_feed_info 
+        SET 
+            feed_active_date = replace((substring(feed_version from '\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\+\d{2}:\d{2}')::timestamptz at time zone 'US/Eastern')::date::text,'-','')::integer
+        ;
+    """
+    op.execute(update_feed_active_date_query)
+
+    op.alter_column(
+        "static_feed_info",
+        "feed_active_date",
+        existing_type=sa.Integer(),
+        nullable=False,
     )
 
     op.create_index(
-        op.f("ix_static_feed_info_version_pub_date"),
+        op.f("ix_static_feed_info_feed_active_date"),
         "static_feed_info",
-        ["version_pub_date"],
+        ["feed_active_date"],
         unique=False,
     )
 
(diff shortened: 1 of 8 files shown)


python_src/src/lamp_py/performance_manager/l0_gtfs_static_load.py
rymarczy, line 1: renamed this file in prep having a gtfs static load and modification pipeline

python_src/src/lamp_py/performance_manager/l0_gtfs_static_load.py
```diff
@@ -190,10 +190,10 @@ def get_table_objects() -> Dict[str, StaticTableDetails]:
 
     return {
         "feed_info": feed_info,
-        "trips": trips,
         "routes": routes,
         "stops": stops,
         "stop_times": stop_times,
+        "trips": trips,
```
rymarczy, line 196: The new trigger for the `trips` table requires it to be loaded after the `stop_times` table.

python_src/src/lamp_py/postgres/postgres_schema.py
```diff
@@ -144,6 +144,7 @@ class StaticFeedInfo(SqlBase):  # pylint: disable=too-few-public-methods
     feed_start_date = sa.Column(sa.Integer, nullable=False)
     feed_end_date = sa.Column(sa.Integer, nullable=False)
     feed_version = sa.Column(sa.String(75), nullable=False, unique=True)
+    version_pub_date = sa.Column(sa.Integer, nullable=True, index=True)
```
rymarczy, line 147: Open to name suggestions on this column. It's kind of a modified `feed_start_date`

python_src/src/lamp_py/migrations/versions/performance_manager/43153d536c2a_feed_info_init_date_trigger.py
```diff
@@ -0,0 +1,67 @@
...
+
+    create_feed_info_insert_function = """
+        CREATE OR REPLACE FUNCTION insert_feed_info() RETURNS TRIGGER AS $$ 
+        BEGIN 
+            IF NEW.feed_version ~ '\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\+\d{2}:\d{2}' THEN
+                NEW.version_pub_date := replace((substring(NEW.feed_version from '\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\+\d{2}:\d{2}')::timestamptz at time zone 'US/Eastern')::date::text,'-','')::integer;
+            ELSE
+                NEW.version_pub_date := NEW.feed_start_date;
+            END IF;
+            RETURN NEW;
```
rymarczy, line 40: This function checks for a the regex match first, to make sure the `substring` function will not fail.

If there is no regex match, then it just inserts `feed_start_date` as `version_pub_date`

python_src/src/lamp_py/migrations/versions/performance_manager/7f4bbec1b669_static_trips_branch_and_trunk_route_id_.py
```diff
@@ -0,0 +1,124 @@
...
+            FROM 
+                static_stop_times sst
+            WHERE 
+                sst.trip_id = p_trip_id
+                AND sst."timestamp" = p_fk_ts
+                AND sst.stop_id IN ('70097', '70098', '70099', '70100', '70101', '70102', '70103', '70104', '70105')
+            ;
+ 
+            RETURN is_braintree;
+        END;
```
rymarczy, line 46: This function abstracts out some of the business logic required in the `insert_static_trips_branch_trunk` function. 

Will return a true of false, depending on if any of the `stop_id`s for stations after JFK, going to braintree, exist in the trip.

python_src/src/lamp_py/migrations/versions/performance_manager/7f4bbec1b669_static_trips_branch_and_trunk_route_id_.py
```diff
@@ -0,0 +1,124 @@
...
+                is_ashmont
+            FROM 
+                static_stop_times sst
+            WHERE 
+                sst.trip_id = p_trip_id
+                AND sst."timestamp" = p_fk_ts
+                AND sst.stop_id IN ('70087', '70088', '70089', '70090', '70091', '70092', '70093', '70094')
+            ;
+ 
+            RETURN is_ashmont;
```
rymarczy, line 68: Same logic as the braintree function, just checking for stops after JFK going to Ashmont

rymarczy:
I thought we could try putting some of this logic into the database using PostgreSQL functions and triggers. Could also be done server side in the ECS, but thought I'd give this a whirl.

python_src/src/lamp_py/postgres/postgres_schema.py
mzappitello: `feed_activation_date`? its the date that this schedule is going to start being used, right?

also, do we want it to be nullable? aren't we going to be using this column to decide which schedule we want to use?

python_src/src/lamp_py/migrations/versions/performance_manager/43153d536c2a_feed_info_init_date_trigger.py
```diff
@@ -0,0 +1,67 @@
...
+        END;
+        $$ LANGUAGE plpgsql;
+    """
+    op.execute(create_feed_info_insert_function)
+
+    create_trigger = """
+        CREATE TRIGGER insert_into_feed_info BEFORE INSERT ON static_feed_info 
+        FOR EACH ROW EXECUTE PROCEDURE insert_feed_info();
+    """
```
mzappitello, line 49: so this is running the above function each time we insert into our feed info table?

if we have another schema version where we need to run a trigger on this table, does this one stay in place or does it get modified?

rymarczy: Yes, the function should modify the row being inserted right before insert.

This trigger stays in place unless we manually change it with an alembic update. If we update the schema in a way that would break this trigger functionality, then the RDS would start throwing errors on every insert into the table.

python_src/src/lamp_py/postgres/postgres_schema.py
rymarczy: Incorporated, changed to `feed_active_date` and made non-nullable, along with a migration call to populate existing records in dev RDS.

python_src/src/lamp_py/migrations/versions/performance_manager/7f4bbec1b669_static_trips_branch_and_trunk_route_id_.py
```diff
@@ -0,0 +1,124 @@
...
+                    NEW.trunk_route_id := NEW.route_id;
+                ELSE
+                    NEW.branch_route_id := null;
+                    NEW.trunk_route_id := NEW.route_id;
+                END IF;
+            ELSE
+                NEW.branch_route_id := null;
+                NEW.trunk_route_id := NEW.route_id;
+            END IF;
+            RETURN NEW;
```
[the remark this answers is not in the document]
mzappitello: those names seem fine to me. thats what pam and heather keep referring to them as.

python_src/src/lamp_py/performance_manager/l0_gtfs_static_load.py
mzappitello: i think we should have a comment here to that effect.

python_src/src/lamp_py/postgres/postgres_schema.py
mzappitello, line 1: do we want to add these to the API side as well or is that something we want to do on an "as needed" basis?

mzappitello (commented):
i had a few other small things to go over before merging.

mzappitello:
> I thought we could try putting some of this logic into the database using PostgreSQL functions and triggers. Could also be done server side in the ECS, but thought I'd give this a whirl.

my take is that i'd rather have this configuration stuff  here rather than server side since it'll allow us to deploy the same thing without going in manually. i do have some concerns, but we can work those out.

python_src/src/lamp_py/performance_manager/l0_gtfs_static_load.py
rymarczy: Added

python_src/src/lamp_py/migrations/versions/performance_manager/7f4bbec1b669_static_trips_branch_and_trunk_route_id_.py
```diff
@@ -0,0 +1,124 @@
...
+                IF red_is_braintree_branch(NEW.trip_id, NEW."timestamp") THEN
+                    NEW.branch_route_id := 'Red-Braintree';
+                    NEW.trunk_route_id := NEW.route_id;
+                ELSEIF red_is_ashmont_branch(NEW.trip_id, NEW."timestamp") THEN
+                    NEW.branch_route_id := 'Red-Ashmont';
+                    NEW.trunk_route_id := NEW.route_id;
+                ELSE
+                    NEW.branch_route_id := null;
+                    NEW.trunk_route_id := NEW.route_id;
+                END IF;
```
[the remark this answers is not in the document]
rymarczy: Incorporated

python_src/src/lamp_py/postgres/postgres_schema.py
rymarczy: Incorporated

rymarczy:
> i had a few other small things to go over before merging.

Should all be incorporated now.

mzappitello (approved):
Looks good to me. Squash and Merge away.

PR #121, opened 2023-04-25
Task: Create Dockerfile for API

mzappitello:
Create Dockerfile for API that can be used to deploy our application and to run locally in a container next to the performance manager database.

* edit config/dev.exs to use environment variables for database connection
* add api service to docker-compose
* add mix env variable to .env file

Asana Task: https://app.asana.com/0/1204389777951216/1204304526686816/f

### DIFF
--- a/.env
+++ b/.env
@@ -13,6 +13,3 @@ SPRINGBOARD_BUCKET=mbta-ctd-dataplatform-dev-springboard
 ARCHIVE_BUCKET=mbta-ctd-dataplatform-dev-archive
 ERROR_BUCKET=mbta-ctd-dataplatform-dev-error
 INCOMING_BUCKET=mbta-ctd-dataplatform-dev-incoming
-
-# api env stuff
-MIX_ENV=dev
--- a/.env
+++ b/.env
@@ -13,3 +13,6 @@ SPRINGBOARD_BUCKET=mbta-ctd-dataplatform-dev-springboard
 ARCHIVE_BUCKET=mbta-ctd-dataplatform-dev-archive
 ERROR_BUCKET=mbta-ctd-dataplatform-dev-error
 INCOMING_BUCKET=mbta-ctd-dataplatform-dev-incoming
+
+# api env stuff
+MIX_ENV=dev
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -1,12 +1,23 @@
-FROM elixir:1.14.4
+# elixir version and erlang version are defined in .tool_versions
+ARG ELIXIR_VERSION=1.14.4
+ARG ERLANG_VERSION=25.3
+
+# this is the latest as of 27 april 2023
+ARG ALPINE_VERSION=3.17.2
+
+# use this image to install dependencies and build the application
+FROM hexpm/elixir:${ELIXIR_VERSION}-erlang-${ERLANG_VERSION}-alpine-${ALPINE_VERSION} as builder
 
 # create a directory and add the project files to it
 WORKDIR /root
 
-ADD assets assets
-ADD config config
-ADD lib lib
-ADD mix.* /root/
+# define this as a prod environment
+ENV MIX_ENV=prod
+
+COPY assets assets
+COPY config config
+COPY lib lib
+COPY mix.* /root/
 
 #  install hex, rebar, and dependencies
 RUN mix local.hex --force
@@ -14,7 +25,27 @@ RUN mix local.rebar --force
 RUN mix deps.get --only prod
 
 # build the application
-RUN mix do compile
+RUN mix release 
+
+# use this image to run the application
+FROM alpine:${ALPINE_VERSION}
+
+# add dumb init for entrypoint
+RUN apk upgrade --no-cache --update
+RUN apk add --no-cache --update dumb-init libstdc++ ncurses-libs libgcc libssl1.1
+
+# create a directory to run in and a directory for the compiled application
+RUN mkdir /work /api
+RUN adduser -D api && chown api /api && chown api /work
+COPY --from=builder /root/_build/prod/rel/api /api
+
+ENV MIX_ENV=prod
+
+# create a user to own the work dir, switch to that user and dir
+USER api
+WORKDIR /work
+
+ENTRYPOINT ["/usr/bin/dumb-init", "--"]
 
 # run the application
-CMD ["mix", "phx.server"]
+CMD ["/api/bin/api", "start"]
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -1,0 +1,20 @@
+FROM elixir:1.14.4
+
+# create a directory and add the project files to it
+WORKDIR /root
+
+ADD assets assets
+ADD config config
+ADD lib lib
+ADD mix.* /root/
+
+#  install hex, rebar, and dependencies
+RUN mix local.hex --force
+RUN mix local.rebar --force
+RUN mix deps.get --only prod
+
+# build the application
+RUN mix do compile
+
+# run the application
+CMD ["mix", "phx.server"]
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -8,21 +8,28 @@ ARG ALPINE_VERSION=3.17.2
 # use this image to install dependencies and build the application
 FROM hexpm/elixir:${ELIXIR_VERSION}-erlang-${ERLANG_VERSION}-alpine-${ALPINE_VERSION} as builder
 
+RUN mix local.hex --force
+RUN mix local.rebar --force
+
 # create a directory and add the project files to it
 WORKDIR /root
 
 # define this as a prod environment
 ENV MIX_ENV=prod
 
-COPY assets assets
-COPY config config
-COPY lib lib
+# copy config and mix files over
+RUN mkdir config
+COPY config/config.exs config/config.exs
+COPY config/prod.exs config/prod.exs
 COPY mix.* /root/
 
-#  install hex, rebar, and dependencies
-RUN mix local.hex --force
-RUN mix local.rebar --force
+# get dependndencies
 RUN mix deps.get --only prod
+
+# copy code files over
+COPY assets assets
+COPY config/runtime.exs config
+COPY lib lib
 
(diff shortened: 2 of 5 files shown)


.env
```diff
@@ -13,3 +13,6 @@ SPRINGBOARD_BUCKET=mbta-ctd-dataplatform-dev-springboard
 ARCHIVE_BUCKET=mbta-ctd-dataplatform-dev-archive
 ERROR_BUCKET=mbta-ctd-dataplatform-dev-error
 INCOMING_BUCKET=mbta-ctd-dataplatform-dev-incoming
+
+# api env stuff
+MIX_ENV=dev
```
mzappitello, line 18: if we're running locally, i think we want the dev environment.

api/Dockerfile
mzappitello, line 1: this is much simpler than the dockerfiles in the other apps, but i'm not sure that we have a need for their bells and whistles yet.

this uses an elixir image, moves our code into it, adds dependencies, compiles, and starts up the server.

api/config/dev.exs
```diff
@@ -2,11 +2,11 @@ import Config
...
-  username: "postgres",
-  password: "postgres",
-  hostname: "localhost",
-  database: "performance_manager",
-  port: "5432",
+  username: System.get_env("DB_USER"),
+  password: System.get_env("DB_PASSWORD"),
+  hostname: System.get_env("DB_HOST"),
+  database: System.get_env("DB_NAME"),
+  port: System.get_env("DB_PORT"),
```
mzappitello, line 9: this change implies we'll only be able to run dev in a docker container until we figure out how we want to [handle env vars](https://app.asana.com/0/1189492770004753/1204390911340914/f) in the shell.

i needed to bring hostname over in order for the images to be able to find the locally run database.

api/config/dev.exs
```diff
@@ -20,7 +20,7 @@ config :api, Api.Repo,
 config :api, ApiWeb.Endpoint,
   # Binding to loopback ipv4 address prevents access from other machines.
   # Change to `ip: {0, 0, 0, 0}` to allow access from other machines.
-  http: [ip: {127, 0, 0, 1}, port: 4000],
+  http: [ip: {0, 0, 0, 0}, port: 4000],
```
mzappitello, line 23: want to be able to access the endpoints from local into the container.

docker-compose.yml
```diff
@@ -7,8 +7,8 @@ services:
     image: postgres:14.4
     env_file: .env
     environment:
-      POSTGRES_DB: ${DB_NAME}
-      POSTGRES_PASSWORD: ${DB_PASSWORD}
+      POSTGRES_DB: $${DB_NAME}
+      POSTGRES_PASSWORD: $${DB_PASSWORD}
```
mzappitello, line 11: was getting warning about not using $$

docker-compose.yml
```diff
@@ -34,9 +34,22 @@ services:
...
+    # network_mode: "host"
+    env_file: .env
+    build:
+      context: ./api
+    ports:
+      # Mapping the port to make the Phoenix app accessible outside of the container.
+      - "4000:4000"
+    depends_on:
+      - local_rds
+
```
mzappitello, line 52: add the server to docker compose.

docker-compose.yml
rymarczy: It seems like this is causing my `password authentication failed` error, changing back to single dollar sign fixes for me.

rymarczy:
Just one issue with the `$$` in the `local_rds` environment section

.env
paulswartz: This is the default, so you don't need to include it in the environment variables.

api/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+ADD lib lib
+ADD mix.* /root/
+
+#  install hex, rebar, and dependencies
+RUN mix local.hex --force
+RUN mix local.rebar --force
+RUN mix deps.get --only prod
+
+# build the application
+RUN mix do compile
```
paulswartz, line 17: nitpick: you don't need `do` if you're only running one item
```suggestion
RUN mix compile
```

api/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM elixir:1.14.4
```
paulswartz, line 1: suggestion: we generally use the containers from `hexpm/elixir`, which allow us to specify the Elixir, Erlang, and Linux versions we're using.

api/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+# create a directory and add the project files to it
+WORKDIR /root
+
```
paulswartz, line 5: suggestion: somewhere in here (at least before `mix phx.server`) you should set `MIX_ENV=prod`
```suggestion
ENV MIX_ENV=prod
```

api/config/dev.exs
paulswartz: This isn't the environment where that happens. You want to update the `prod` environment, which is where all deployments run (even if it's in the staging or other environment according to the MBTA).

api/Dockerfile
mzappitello: so the mix env should always be prod inside of a docker container? in that scheme, we only ever use mix dev when launching from the command line?

from [this](https://github.com/mbta/lamp/pull/121#discussion_r1178201722) comment it sounds like there is a doc on this?

api/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+#  install hex, rebar, and dependencies
+RUN mix local.hex --force
+RUN mix local.rebar --force
+RUN mix deps.get --only prod
+
+# build the application
+RUN mix release 
+
+# use this image to run the application
+FROM alpine:${ALPINE_VERSION}
```
mzappitello, line 31: i switched to a multi stage build as suggested in the new app checklist.

after looking at some of the other projects, this seems to be the pattern they are using.

api/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+
+# add dumb init for entrypoint
+RUN apk upgrade --no-cache --update
+RUN apk add --no-cache --update dumb-init libstdc++ ncurses-libs libgcc libssl1.1
+
+# create a directory to run in and a directory for the compiled application
+RUN mkdir /work /api
+RUN adduser -D api && chown api /api && chown api /work
+COPY --from=builder /root/_build/prod/rel/api /api
```
mzappitello, line 40: adding a new user to limit vulnerabilities, also as outlined in new app checklist.

api/config/runtime.exs
```diff
@@ -21,77 +21,94 @@ if System.get_env("PHX_SERVER") do
...
+  if System.get_env("BOOTSTRAPPED") do
+    config :api, Api.Repo,
+      username: System.get_env("DB_USER"),
+      password: System.get_env("DB_PASSWORD"),
+      hostname: System.get_env("DB_HOST"),
+      database: System.get_env("DB_NAME"),
+      port: System.get_env("DB_PORT"),
+      stacktrace: true,
+      show_sensitive_data_on_connection_error: true,
+      pool_size: 10
```
mzappitello, line 34: i added this if block that will be triggered when we are running locally via docker compose, switching on a bootstrapped var in the env file. it might be better to add another var there that says we're local?

if this block isn't hit, we'll do the other runtime config stuff that should describe how our endpoint and rds connection are configured. i know that @paulswartz had suggested putting this in `prod.exs`, but that file is evaluated at build time rather than runtime, meaning it wouldn't have access to these env vars?

api/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+
+COPY assets assets
+COPY config config
+COPY lib lib
+COPY mix.* /root/
+
+#  install hex, rebar, and dependencies
+RUN mix local.hex --force
+RUN mix local.rebar --force
+RUN mix deps.get --only prod
```
paulswartz, line 25: nitpick: from an ordering perspective, this isn't the most efficient. since you're also running this container for local development, it'll make the container rebuild more frequently that you need.

- local.hex / local.rebar: these can happen before you add any files
- deps.get: should happen after mix.*, config/config.exs and config/prod.exs are added
- release: after the rest of the files (include config/runtime.exs) are added

api/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+# add dumb init for entrypoint
+RUN apk upgrade --no-cache --update
+RUN apk add --no-cache --update dumb-init libstdc++ ncurses-libs libgcc libssl1.1
+
+# create a directory to run in and a directory for the compiled application
+RUN mkdir /work /api
+RUN adduser -D api && chown api /api && chown api /work
+COPY --from=builder /root/_build/prod/rel/api /api
+
+ENV MIX_ENV=prod
```
paulswartz, line 42: suggestion: you don't need this (it's already baked into the release). but you might need `PHX_SERVER=true`

api/config/runtime.exs
```diff
@@ -21,77 +21,94 @@ if System.get_env("PHX_SERVER") do
...
 
-  maybe_ipv6 = if System.get_env("ECTO_IPV6") in ~w(true 1), do: [:inet6], else: []
+    config :api, ApiWeb.Endpoint,
+      http: [ip: {0, 0, 0, 0}, port: 4000],
+      cache_static_manifest: "priv/static/cache_manifest.json",
+      server: true
+  else
```
paulswartz, line 39: it's a bit hard to read this: is some of the configuration the same between the two cases?

mzappitello: i'm not really sure what needs to go in the runtime portion when we're on aws. i've noticed that other repos don't even have this runtime.exs file.

i thought about deleting it altogether, but didn't have a good justification for doing so.

paulswartz: You definitely can't delete it (since you need to tell the application where to access the database). I took a look in my editor and this looks okay: we can clean it up later if we need to.

api/Dockerfile
mzappitello: i think i put that in the runtime configuration.

paulswartz (approved):

PR #123, opened 2023-04-28
Chore: Reorder py Dockerfile for faster buildtimes

mzappitello:
By reordering when files are copied and dependencies are installed, we can improve build times when we are making when we only modify source files and not dependencies.

### DIFF
--- a/python_src/Dockerfile
+++ b/python_src/Dockerfile
@@ -13,9 +13,15 @@ RUN apt-get install -y libpq-dev gcc curl
 RUN pip install -U pip
 RUN pip install "poetry==1.1.14"
 
-# Copy src dir and pyproject file and create a virtualenv with all needed dependenciies using poetry
-COPY . /lamp/
+# copy poetry and pyproject files and install dependencies
 WORKDIR /lamp/
+COPY poetry.lock poetry.lock
+COPY pyproject.toml pyproject.toml
+RUN poetry install --no-dev --no-interaction --no-ansi -v
+
+# Copy src directory to run against and build lamp py
+COPY src src
+COPY alembic.ini alembic.ini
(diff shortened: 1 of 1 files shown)


python_src/Dockerfile
```diff
@@ -13,11 +13,18 @@ RUN apt-get install -y libpq-dev gcc curl
...
+RUN mkdir lamp/
+
+# copy poetry and pyproject files and install dependencies
 WORKDIR /lamp/
+COPY poetry.lock poetry.lock
+COPY pyproject.toml pyproject.toml
 RUN poetry install --no-dev --no-interaction --no-ansi -v
 
+# Copy src directory to run against
+COPY src src
```
rymarczy, line 26: Just need to run `poetry install` again to install the `lamp_py` module

```suggestion
COPY src src
RUN poetry install --no-dev --no-interaction --no-ansi -v
```

python_src/Dockerfile
```diff
@@ -13,11 +13,18 @@ RUN apt-get install -y libpq-dev gcc curl
 RUN pip install -U pip
 RUN pip install "poetry==1.1.14"
 
-# Copy src dir and pyproject file and create a virtualenv with all needed dependenciies using poetry
-COPY . /lamp/
+# Create a directory for project to live in
+RUN mkdir lamp/
+
+# copy poetry and pyproject files and install dependencies
 WORKDIR /lamp/
```
rymarczy, line 20: nitpick: I think `WORKDIR` creates a directory, if one doesn't exist, so the `RUN mkdir lamp/` is technically redundant. 

If we want to stick with `RUN mkdir lamp/`, it should probably be an absolute path `/lamp/` instead of relative,  in case our pwd isn't root.

rymarczy (approved):
looks good

PR #124, opened 2023-04-29
Feat: update `add_fk_static_timestamp_column` with new match logic

rymarczy:
After discussions with OPMI, new rules to match GTFS-RT events to GTFS schedules are as follows:

Matching between a GTFS-RT events and a GTFS-static schedule version will be done on a whole `start_date` / "service_date" basis. Each distinct `start_date` / "service_date" of GTFS-RT data will be matched to only one GTFS-static version. 

When ingesting live GTFS Schedule files, the `feed_active_date` will be extracted from the ISO timestamp found in the `feed_version` field of the FEED_INFO table. This will be the first date that a GTFS-static version can be applied to GTFS-RT events. 

GTFS-RT `start_date` must be between `feed_start_date` and `feed_end_date` in FEED_INFO, `start_date` must also be less than or equal to `feed_active_date`.  The most recent schedule that meets these criteria will be matched

When processing archived GTFS-static schedules, these rules change because `feed_start_date` and `feed_end_date` are retroactively modified in archived schedules so that only one GTFS-static version is applicable per calendar day. 

If the initial matching fails, we will drop the requirements of matching on `feed_active_date` and just match `start_date` by `feed_start_date` and `feed_end_date`. 

Asana Task: https://app.asana.com/0/1204389777951216/1204500786018314

### DIFF
--- a/python_src/src/lamp_py/performance_manager/gtfs_utils.py
+++ b/python_src/src/lamp_py/performance_manager/gtfs_utils.py
@@ -67,7 +67,6 @@ def start_time_to_seconds(
 
 def add_fk_static_timestamp_column(
     events_dataframe: pandas.DataFrame,
-    timestamp_column: str,
     db_manager: DatabaseManager,
 ) -> pandas.DataFrame:
     """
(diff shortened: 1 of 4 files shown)


python_src/src/lamp_py/performance_manager/gtfs_utils.py
```diff
@@ -79,58 +78,67 @@ def add_fk_static_timestamp_column(
...
+                StaticFeedInfo.feed_start_date <= date,
+                StaticFeedInfo.feed_end_date >= date,
+                StaticFeedInfo.feed_active_date <= date,
+            )
+            .order_by(
+                StaticFeedInfo.feed_active_date.desc(),
+                StaticFeedInfo.created_on.desc(),
+            )
+            .limit(1)
+        )
```
rymarczy, line 106: This initial query will match to GTFS-static schedules we are processing in real time with the `feed_active_date` field from FEED_INFO

python_src/src/lamp_py/performance_manager/gtfs_utils.py
```diff
@@ -79,58 +78,67 @@ def add_fk_static_timestamp_column(
...
+            .where(
+                StaticFeedInfo.feed_start_date <= date,
+                StaticFeedInfo.feed_end_date >= date,
+            )
+            .order_by(
+                StaticFeedInfo.feed_start_date.desc(),
+                StaticFeedInfo.created_on.desc(),
             )
-            .order_by(StaticFeedInfo.timestamp.desc())
             .limit(1)
```
rymarczy, line 121: This archive query will get us matches to archived GTFS-static schedules with modified `feed_start_date` and `feed_end_date` fields.

python_src/src/lamp_py/performance_manager/gtfs_utils.py
```diff
@@ -67,7 +67,6 @@ def start_time_to_seconds(
 
 def add_fk_static_timestamp_column(
     events_dataframe: pandas.DataFrame,
-    timestamp_column: str,
```
rymarczy, line 70: We're not basing GTFS-static matching on a RT-timestamp column anymore, it's only based on `start_date`

mzappitello:
could a schedule be added that would require we retroactively update our fk_static_timestamp "key"? it seems like if one is uploaded where the feed activation date is later than the date in some of our trips, we could want to change that?

also, now that i see it and we're doing this adding twice, it makes me wonder if we shouldn't be calling this on the vehicle events data frames after we join the trip updates and vehicle positions dataframes. thats maybe outside of scope here though?

rymarczy:
The discussions we had with OPMI resulted in the decision to not retroactively update our fk_static_timestamp keys. When processing live GTFS Schedule files, we will only start applying them on the day there were received, based on the `active` date included in the `feed_version` field. 

Currently, we can not call this function later, after joining the dataframes, because joining to static data is required to add the parent_station field (for event grouping) and dropping bus events from the dataframes. Calling this later would result in much larger dataframes moving through the rest of the processing pipeline.

mzappitello:
is there a link we can post as a comment that documents why we don't want to retroactively reassign schedules? if not, i think we'd want to have a comment to that effect in the code. this feels like the kind of decision that is tough to retroactively recall when refactoring in a year or two.

other than that, LGTM.

mzappitello:
lgtm

PR #126, opened 2023-05-02
Chore: remove env loading from python_src

mzappitello:
We should be able to use direnv to handle this for us, even inside of a poetry shell. The following steps worked for me, and should work in linux as well.
* Install direnv to your machine. (i'm on version 2.32.2)
* Update the config toml file to load dot env files.
  * https://stackoverflow.com/a/71635629
* Run `direnv allow .` in the root dir
* Run `poetry install` once more in python_src to install the dependencies into the new virtualenv.

Asana Task: https://app.asana.com/0/1189492770004753/1204390911340914/f


rymarczy:
Should we be using `asdf` to install `direnv` in a more controlled manner? 

I doubt there would be breaking changes but my system is on version 2.25.2 for `direnv` and `asdf` seems like a good way to make sure everyone using the project is one the same version.

rymarczy:
This seems to be working for me. I followed the directions here https://github.com/asdf-community/asdf-direnv instead of the stackoverflow link you posted. 

I ran `asdf direnv local` in our root project directory and that set `direnv` to allow in the folder and created  a `.envrc` file in our root project folder with just the following inside:
`use asdf`

rymarczy (approved):

PR #128, opened 2023-05-04
CHORE: Remove Ingestion from Docker Compose

mzappitello:
As Ryan said, its kind of like having a loaded gun sitting around since it is pointing right at our live aws buckets.


rymarczy:
LGTM

PR #132, opened 2023-05-10
feat: Add static directions table to trips

rymarczy:
At the request of OPMI, this PR adds `directions` table data from the GTFS static schedule to the `vehicle_trips` table.

The directions data is added by creating a `static_diretions` table in the RDS, and then joining the `direction` and `direction_destination` fields to newly processed trips with the `update_directions` function in `l1_rt_trips.py`

Asana Task: https://app.asana.com/0/1204517240145658/1203464608464154


python_src/src/lamp_py/migrations/versions/performance_manager/001_be4c10c548f0_initial_rds_schema.py
rymarczy, line 1: renamed all these to make them look a lot nicer in the folder browser

python_src/src/lamp_py/performance_manager/l0_gtfs_static_load.py
```diff
@@ -202,6 +221,7 @@ def get_table_objects() -> Dict[str, StaticTableDetails]:
         "trips": trips,
         "calendar": calendar,
         "calendar_dates": calendar_dates,
+        "directions": directions,
```
mzappitello, line 224: we don't care about the order that directions are loaded in do we?

mzappitello (commented):
two comments. one on git practices and one is a question.

python_src/src/lamp_py/performance_manager/l0_gtfs_static_load.py
rymarczy: We do not, the table is not used until RT trips records are loaded.

mzappitello:
🍰

PR #133, opened 2023-05-11
chore: rename migration files with migration order

rymarczy:
The migration files show up in the migration folder in random order based on the migration hash string.

This change pre-pends the migration files with their migration order to more easily manage the files in the folder structure.


mzappitello:
:cake

PR #137, opened 2023-05-17
chore: drop trips columns from `vehicle_events` table

rymarczy:
Our `vehicle_events` table still contains columns that are duplicated in the `vehicle_trips` table. This change removes the following columns from `vehicle_events`:

- service_date
- route_id
- vehicle_id
- start_time
- direction_id

Asana Task: https://app.asana.com/0/1204614792357231/1204616772301249

### DIFF
--- a/python_src/src/lamp_py/migrations/versions/performance_manager/014_18d962b3df59_remove_trip_fields_from_events.py
+++ b/python_src/src/lamp_py/migrations/versions/performance_manager/014_18d962b3df59_remove_trip_fields_from_events.py
@@ -1,0 +1,63 @@
+"""remove trip fields from events
+
+Revision ID: 18d962b3df59
+Revises: 97f5fb821b8a
+Create Date: 2023-05-17 13:06:46.832362
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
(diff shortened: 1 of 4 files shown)


python_src/src/lamp_py/migrations/versions/performance_manager/014_18d962b3df59_remove_trip_fields_from_events.py
```diff
@@ -0,0 +1,63 @@
...
+            autoincrement=False,
+            nullable=True,
+        ),
+    )
+    op.add_column(
+        "vehicle_events",
+        sa.Column(
+            "service_date", sa.INTEGER(), autoincrement=False, nullable=True
+        ),
+    )
```
rymarczy, line 63: On downgrade for these columns `nullable` need to be `True` to account for adding a new column result in all null values for the added columns.

mzappitello (commented):
wont this line cause problems, if we're trying to add values to columns that don't exist?

https://github.com/mbta/lamp/blob/b682e82acb9d45cbddfb28b5dfd15bd47782c63a/python_src/src/lamp_py/performance_manager/l0_gtfs_rt_events.py#L310-L332

mzappitello (approved):
🍰

PR #138, opened 2023-05-19
change `timestamp` to `static_version_key`

rymarczy:
Our use of `timestamp` as a foreign key column that represents GTFS Static Schedule versions is very confusing.

This change renames the `timestamp` columns to `static_version_key` across the performance manager application. 

Asana Task: https://app.asana.com/0/1204614792357231/1204616772301260

### DIFF
--- a/python_src/src/lamp_py/migrations/versions/performance_manager/015_4fe83fd4091d_timestamp_to_static_version_key.py
+++ b/python_src/src/lamp_py/migrations/versions/performance_manager/015_4fe83fd4091d_timestamp_to_static_version_key.py
@@ -1,0 +1,797 @@
+"""timestamp to static_version_key
+
+Revision ID: 4fe83fd4091d
+Revises: 18d962b3df59
+Create Date: 2023-05-19 09:52:16.177008
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = "4fe83fd4091d"
+down_revision = "18d962b3df59"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.drop_constraint(
+        "vehicle_events_fk_static_timestamp_fkey",
+        "vehicle_events",
+        type_="foreignkey",
+    )
+    op.drop_constraint(
+        "vehicle_trips_fk_static_timestamp_fkey",
+        "vehicle_trips",
+        type_="foreignkey",
+    )
+    op.drop_constraint(
+        "static_feed_info_timestamp_key", "static_feed_info", type_="unique"
+    )
+
+    op.drop_column("vehicle_events", "fk_static_timestamp")
+
+    op.alter_column(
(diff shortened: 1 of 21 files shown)


python_src/src/lamp_py/postgres/postgres_schema.py
```diff
@@ -39,13 +39,6 @@ class VehicleEvents(SqlBase):  # pylint: disable=too-few-public-methods
     vp_stop_timestamp = sa.Column(sa.Integer, nullable=True)
     tu_stop_timestamp = sa.Column(sa.Integer, nullable=True)
 
-    # forign key to static schedule expected values
-    fk_static_timestamp = sa.Column(
-        sa.Integer,
-        sa.ForeignKey("static_feed_info.timestamp"),
-        nullable=False,
-    )
-
```
mzappitello, line 48: i guess this is just associated with the trip now, rather than any of the specific events?

python_src/src/lamp_py/migrations/versions/performance_manager/015_4fe83fd4091d_timestamp_to_static_version_key.py
```diff
@@ -0,0 +1,797 @@
...
+        "fk_static_timestamp",
+        new_column_name="static_version_key",
+    )
+    op.create_foreign_key(
+        "vehicle_trips_static_version_key_fkey",
+        "vehicle_trips",
+        "static_feed_info",
+        ["static_version_key"],
+        ["static_version_key"],
+    )
```
mzappitello, line 56: this adds the constraint that we dropped here?

https://github.com/mbta/lamp/pull/138/files#diff-86cee9a01a079edb3a0547026e377f156eb76d2326da00bafb68600ea495d49aR25-R29

mzappitello (approved):
one question on the migration file but this all looks good to me.

python_src/src/lamp_py/postgres/postgres_schema.py
rymarczy: Yes, our new logic that matches static schedules to realtime events makes it so all events for each service date should match to the same static schedule. So there's no way that different stops in the same trip would be matched to a different static schedule. Tracking this in the trips table should be totally sufficient. 

I really should of taken this out with the other PR that dropped the trip info from the events table, but I missed it.

python_src/src/lamp_py/migrations/versions/performance_manager/015_4fe83fd4091d_timestamp_to_static_version_key.py
rymarczy: Technically, this is modifying the foreign key constraint on the `vehicle_trips` table, changing the reference from the `timestamp` column to `static_version_key`. But there's no way to modify a foreign key constraint, so you just have to drop it and add a new one. 

This is not currently represented in the api schema, as a foreign key, but I don't think it's really needed. The foreign key modeling is more for schema migration purposes than anything, and the api application code won't be doing any migration actions.

PR #139, opened 2023-05-23
All Headways as Departure to Departure 

rymarczy:
OPMI has requested that all headways be calculated as departure to departure for every stop. 

Previously, the first stop of every trip was departure to departure and all other stops were calculated as arrival to arrival. 

Asana Task: https://app.asana.com/0/1204614792357231/1204659932565026


mzappitello:
LGTM. Pretty much what I expected.

PR #143, opened 2023-06-08
Add logging checking files are moved correctly

mzappitello:
We've been seeing errors in the performance manager where a static schedule has been ingested twice. Logs from the ingestion pipeline also show instances where a schedule has been processed twice. The current theory is that our s3 move files function is not correctly deleting the incoming static schedule the first time, and its getting processed twice.

To confirm this, `Converter.convert` has been modified to return a list of all the files it ingested. In the main method, we now take that list and intersect it with any files that are still in the incoming bucket. Any files that haven't been properly moved will be logged.

Additional changes in this commit:
* some comment improvements i made while i was working out how everything in the gtfs_rt threading work
* process logger in the `verify_environment` function that logs all the env vars we're using.

Asana Task: https://app.asana.com/0/1189492770004753/1203403213474707/f


python_src/src/lamp_py/ingestion/convert_gtfs.py
```diff
@@ -52,7 +52,7 @@ class GtfsConverter(Converter):
     Converter for GTFS Schedule Data
     """
 
-    def convert(self) -> None:
+    def convert(self) -> List[str]:
```
mzappitello, line 55: all the converters now return a list of all the files they moved.

python_src/src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -141,7 +144,7 @@ def process_files(self) -> Iterable[pyarrow.table]:
         yield_threshold = max(10, max_workers * 3)
 
         process_logger = ProcessLogger(
-            "create_parquet_table",
+            "create_pyarrow_tables",
```
mzappitello, line 147: renamed this to make it clearer in the logs. we also have "parquet table creator" process that its easy to get confused with.

python_src/src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -152,7 +155,9 @@ def process_files(self) -> Iterable[pyarrow.table]:
             for result_dt, result_filename, result_table in pool.map(
                 self.gz_to_pyarrow, self.files
             ):
-                # handle error in gz_to_pyarrow processing
+                # errors in gtfs_rt conversions are handled in the gz_to_pyarrow
+                # function. if one is encountered, the datetime will be none. log
+                # the error and move on to the next file.
```
mzappitello, line 160: i improved a few comments in here as well.

i was coming back to this file with fresh eyes and i put comments on the sections where it wasn't quickly obvious why things were the way they are.

python_src/src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -291,6 +326,8 @@ def gz_to_pyarrow(
                 for key, value in record.items():
                     table[key].append(value)
 
+        except FileNotFoundError as _:
+            return (None, filename, None)
```
mzappitello, line 330: this is one of the errors we can hit and we don't need to init the thread when we do.

python_src/src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -330,12 +367,14 @@ def move_s3_files(self) -> None:
...
         """
         if len(self.error_files) > 0:
+            self.moved_files += self.error_files
             self.error_files = move_s3_objects(
                 self.error_files,
                 os.path.join(os.environ["ERROR_BUCKET"], DEFAULT_S3_PREFIX),
             )
 
         if len(self.archive_files) > 0:
+            self.moved_files += self.archive_files
```
mzappitello, line 377: add the error and archive files to the moved files. then this list is returned after conversion.

python_src/src/lamp_py/ingestion/ingest.py
```diff
@@ -72,7 +73,7 @@ def ingest_files(files: List[str], metadata_queue: Queue) -> None:
     # GTFS Static Schedule must be processed first for performance manager to
     # work as expected
     if ConfigType.SCHEDULE in converters:
-        converters[ConfigType.SCHEDULE].convert()
+        moved_static_files = converters[ConfigType.SCHEDULE].convert()
```
mzappitello, line 76: get the static files that were moved. they'll be added to the moved realtime files.

python_src/src/lamp_py/ingestion/ingest.py
```diff
@@ -83,10 +84,12 @@ def ingest_files(files: List[str], metadata_queue: Queue) -> None:
...
     # launching converter processes with map_async avoids throwing of SIGTERM signals and blocking
     with Pool(processes=len(converters)) as pool:
-        pool.map_async(run_converter, converters.values())
-        pool.close()
-        pool.join()
+        result = pool.map_async(run_converter, converters.values())
+        moved_realtime_files = result.get()
+
+    # join all of the lists of processed files into a single list and return
+    return sum(moved_realtime_files, moved_static_files)
```
mzappitello, line 95: return all the moved files as a single list.

python_src/src/lamp_py/ingestion/pipeline.py
```diff
@@ -39,23 +42,32 @@ def validate_environment() -> None:
...
 
     if missing_required:
-        raise EnvironmentError(
+        exception = EnvironmentError(
             f"Missing required environment variables {missing_required}"
         )
+        process_logger.log_failure(exception)
+        raise exception
+    
+    process_logger.log_complete()
```
mzappitello, line 67: add a process logger to log the env vars we're launching with.

python_src/src/lamp_py/ingestion/pipeline.py
```diff
@@ -71,23 +83,67 @@ def ingest(metadata_queue: Queue) -> None:
...
+
+    return processed_files
+
+
+def wait_for_deletion(filenames: List[str]) -> None:
+    """
+    wait for all of the files in the filenames list to be removed from the
+    incoming bucket. log how many files are still
+    """
```
mzappitello, line 97: new function that intersects files that should have been moved with the files still in the incoming bucket.

it waits until the intersection is empty or a timeout (which is a function of the number of files) is reached.

it logs files that aren't moved correctly.

python_src/src/lamp_py/ingestion/pipeline.py
```diff
@@ -71,23 +83,67 @@ def ingest(metadata_queue: Queue) -> None:
...
+            )
+        )
+
+        files_to_delete &= existing_files
+        if len(files_to_delete) == 0:
+            break
+
+        # increment attempts and wait at most 30 seconds to check again
+        attempts += 1
+        time.sleep(min(30, int(len(files_to_delete) / 100)))
```
rymarczy, line 119: This random sleep feels like were crossing over into dictating behavior instead of observing it.

python_src/src/lamp_py/ingestion/pipeline.py
```diff
@@ -71,23 +83,67 @@ def ingest(metadata_queue: Queue) -> None:
...
-    schedule.every(5).minutes.do(ingest, metadata_queue=metadata_queue)
-
+    # run the event loop every five minutes
     while True:
         check_for_sigterm(metadata_queue, rds_process)
-        schedule.run_pending()
-        time.sleep(1)
+        processed_files = ingest(metadata_queue=metadata_queue)
+        wait_for_deletion(processed_files)
+        time.sleep(60 * 5)
```
rymarczy, line 146: probably need to add another `check_for_sigterm` call in here before the sleep, with this new scheduling

rymarczy:
We're doing quite a bit of extra work to pass these file lists in and out of the processes. But really, all of the files that are passed into the `ingest_files` function should be moved by the time the function has completed. 

You could just return your `files` list from that you get from `file_list_from_s3` in the `ingest` function and pass that to your `wait_for_deletion` function.

rymarczy:
We also still need some kind of logging from inside of the `_move_s3_object` function. If this logging shows us that we're not moving some objects, we still have no idea why they aren't being moved.

mzappitello:
i added the logging statement to `_move_s3_object`.

we do need to return the files that have been processed since some of the files that are in the bucket will be skipped if they are inside of the current hour.

rymarczy (approved):

rymarczy:
LGTM

PR #145, opened 2023-06-21
Write Feed Info Schedule File Last

mzappitello:
There is a race condition between Ingestion and Performance Manager when writing the parquet files from a gtfs static schedule. The Performance Manager understands there to be a new schedule to read when the Feed Info file is written to the metadata log. In processing this schedule, it will build other URLs for other files in the schedule. If the Feed Info file is written before these other files, it could trigger processing of files that have yet to be written.

To fix this, we want to write the Feed Info table last, so that the Performance Manager will only attempt to read a static schedule when all of the files associated with it have been written.

* change some of the structure of the GtfsConverter class to skip writing the feed info file while looping through zip contents and then write the feed info file
* update tests to acommadate this, including monkey patching the write parquet and move s3 objects functions.

Asana Task: https://app.asana.com/0/1204817744674813/1204868004140902/f


python_src/src/lamp_py/ingestion/convert_gtfs.py
```diff
@@ -97,3 +52,70 @@ def convert(self) -> List[str]:
...
+            table=table,
+            config_type=s3_prefix,
+            s3_path=os.path.join(
+                os.environ["SPRINGBOARD_BUCKET"],
+                DEFAULT_S3_PREFIX,
+                s3_prefix,
+            ),
+            partition_cols=["timestamp"],
+            visitor_func=self.send_metadata,
+        )
```
mzappitello, line 121: i could have left this nested in the above function, but it looked kind of clunky. happy to switch back if you prefer.

the one question i have is around the visitor function. should we only add feed info to the metadata table and avoid writing the other static schedule files there? is anything reading them later?

python_src/tests/ingestion/test_gtfs_converter.py
```diff
@@ -300,17 +303,92 @@ def test_schedule_conversion() -> None:
...
+        return []
+
+    monkeypatch.setattr(
+        "lamp_py.ingestion.convert_gtfs.move_s3_objects", mock_move_s3_objects
+    )
+
+    yield
+
+
+def test_schedule_conversion() -> None:
```
mzappitello, line 359: i think this reads pretty straight forwardly. i added comments where i think that helps.

python_src/src/lamp_py/ingestion/convert_gtfs.py
```diff
@@ -97,3 +52,70 @@ def convert(self) -> List[str]:
...
+        essentially a small database with each contained file (outside of feed
+        info) acting as its own table. info on the gtfs scheduling standard can
+        be found at http://gtfs.org/schedule/
+        """
+        # s3 objects are read directly from s3 as a stream of bytes
+        filelike_input: Union[str, IO[bytes]] = filename
+
+        # the last modified date is pulled from s3 and we use it as our version
+        # key to link all of the table data together
+        last_modified = 0
```
rymarczy, line 68: I think we should take this opportunity to stop using the last modified date for zip file itself as the `timestamp` to link table data together. 

`ZipInfo.date_time` provides the last modified date of the files in the archive. Using this value from the `feed_info.text` file inside of the archive seems like a good option, that won't be altered by moving the archive file itself around S3.

python_src/src/lamp_py/ingestion/convert_gtfs.py
```diff
@@ -97,3 +52,70 @@ def convert(self) -> List[str]:
...
+                # ingest and write the feed info table last.
+                if gtfs_filename == "feed_info.txt":
+                    continue
+                self.create_table(gtfs_zip, gtfs_filename, last_modified)
+
+            self.create_table(gtfs_zip, "feed_info.txt", last_modified)
+
+    def create_table(
+        self, gtfs_zip: zipfile.ZipFile, table_filename: str, version_key: int
+    ) -> pyarrow.Table:
```
rymarczy, line 90: Looks like we're not returning anything from this function.

python_src/src/lamp_py/ingestion/convert_gtfs.py
rymarczy: We are ignoring all of the other tables being in the metadata table, except `feed_info`, so yeah I think we could drop the visitor function for the non `feed_info` tables.

rymarczy (changes requested):
A few comments based on our lessons learned over the last few months.

python_src/src/lamp_py/ingestion/convert_gtfs.py
rymarczy: I think the only other viable option would be to extract the datetime from the `feed_version` field inside of the `feed_info.txt` file.

rymarczy (changes requested):
Just need to fix the return value of `get_zip_buffer`

rymarczy (approved):

rymarczy:
LGTM

PR #147, opened 2023-06-27
Add missing initialization in ingest_files

mzappitello:
the `ingest_files` method is missing an initialization for the `processed_static_files` and `processed_realtime_files` lists. If we don't enter certain gates, they'll remain uninitialized and we throw an error when summing them at the end of the function.


python_src/src/lamp_py/ingestion/ingest.py
mzappitello, line 1: pretty straight forward.

i tested this on the staging environment (since we use dev) and it worked as expected.

rymarczy:
LGTM

PR #148, opened 2023-06-28
fix: 017 Migrate Order of Operations

rymarczy:
Alembic migration 017_37d97f420d54 has caused our RDS to lock up for 24 hours.

Re-ordering the migration steps should allow the migration to actually complete in a reasonable time.

Previously we were adding two new columns, creating an index for those columns and then doing a large UPDATE-FROM command to populate the columns. 

This change moves the index creation to after the UPDATE-FROM operation completes.


mzappitello:
LGTM

PR #149, opened 2023-06-29
Add pool.close and pool.join statements to ingest

mzappitello:
The pool as currently written causes SIGTERMS to be thrown on child processes if a certain race condition is lost, which can happen when using the default ubuntu subprocess starting method. To fix this, add in a `pool.close` and `pool.join` call inside of the pools with block.


rymarczy:
LGTM

PR #153, opened 2023-07-14
reversion: Remove Delete Verification from Ingestion

rymarczy:
Functionality to verify S3 delete/move operations was previously added to ingestion because of duplicate processing errors we were seeing with GTFS files. 

These errors have since been traced back to how Delta populates our `incoming` S3 buckets with two processes running in parallel creating a race condition with file writing to S3. 

Removing the delete/move verification from ingestion because we've never gotten a log indicating that the delete/move didn't work and it adds unnecessary overhead. 

Asana Task: https://app.asana.com/0/1204921490312295/1205063998912225

### DIFF
--- a/python_src/src/lamp_py/ingestion/ingest.py
+++ b/python_src/src/lamp_py/ingestion/ingest.py
@@ -18,12 +18,11 @@ class NoImplConverter(Converter):
     files.
     """
 
-    def convert(self) -> List[str]:
+    def convert(self) -> None:
         move_s3_objects(
             self.files,
             os.path.join(os.environ["ERROR_BUCKET"], DEFAULT_S3_PREFIX),
         )
-        return self.files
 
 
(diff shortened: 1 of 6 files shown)


python_src/src/lamp_py/ingestion/ingest.py
```diff
@@ -85,25 +83,17 @@ def ingest_files(files: List[str], metadata_queue: Queue) -> List[str]:
     # The remaining converters can be run in parallel
     #
     # Using signal.signal to detect ECS termination and multiprocessing.Manager
-    # to manage the metadata queue along with multiprocessing.Poo.map causes
+    # to manage the metadata queue along with multiprocessing.Pool.map causes
```
mzappitello, line 86: 👀

mzappitello (approved):

PR #155, opened 2023-07-26
feat: Flat Schema Migration for Prod Deployment

rymarczy:
Summary of changes:

- Provide flattened alembic migration file for RDS deployment to prod us alembic `db_name` of `performance_manager_prod`
- Add/Edit RDS indexes to allow for performant application function
- Re-name Red line trunk / branch route-id designations to `Red-A` and `Red-B`
- change names of fields in `opmi_all_rt_fields_joined` VIEW per OPMI requests
- add `stop_departure_timestamp` field to `opmi_all_rt_fields_joined' VIEW

This change updates  alembic to use the `ALEMBIC_DB_NAME` environmental variable to select which migration files are used. This will require our prod AWS devops configuration to have this variable set. 

Asana Task: https://app.asana.com/0/1204931901750655/1205131619941048
Asana Task: https://app.asana.com/0/1204931901750655/1204999066398944
Asana Task: https://app.asana.com/0/1204931901750655/1205067159884234
Asana Task: https://app.asana.com/0/1204931901750655/1204999066398936


.env
```diff
@@ -7,6 +7,7 @@ DB_PORT=5432
 DB_NAME=performance_manager
 DB_USER=postgres
 DB_PASSWORD=postgres
+ALEMBIC_DB_NAME=performance_manager_prod
 
 # s3 locations
```
rymarczy, line 12: new env var for alembic to use for migration

python_src/src/lamp_py/performance_manager/l0_gtfs_rt_events.py
```diff
@@ -91,6 +91,11 @@ def combine_events(
     )
     process_logger.log_start()
 
+    if vp_events.shape[0] == 0 and tu_events.shape[0] == 0:
+        process_logger.add_metadata(total_event_count=0)
+        process_logger.log_complete()
+        return pandas.DataFrame()
```
rymarczy, line 97: handle the situation where vehicle events and/or trip updates together produce no usable rows from parquet files

python_src/src/lamp_py/performance_manager/l0_gtfs_static_mod.py
```diff
@@ -6,7 +6,6 @@
     StaticStopTimes,
     StaticStops,
     StaticTrips,
-    TempStaticHeadwaysGen,
 )
```
rymarczy, line 9: Drop this TempStaticHeadwaysGen table, improved table indexes make this unnecessary.

python_src/src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -15,46 +15,54 @@
...
+    """
+    Update `previous_trip_stop_pk_id` and `next_trip_stop_pk_id` fields in `vehicle_events` table
+    """
+    temp_trips = (
+        sa.select(
+            TempEventCompare.pm_trip_id,
+        )
+        .distinct()
+        .subquery()
+    )
```
mzappitello, line 28: we use this so that we're only running the `prev_next_trip_stop_sub` subquery on trips that have been updated with this round of gtfs_rt data?

python_src/src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -320,8 +336,8 @@ def update_directions(db_manager: DatabaseManager) -> None:
         )
         .select_from(VehicleTrips)
         .join(
-            TempEventCompare,
-            TempEventCompare.pm_trip_id == VehicleTrips.pm_trip_id,
+            temp_trips,
+            temp_trips.c.pm_trip_id == VehicleTrips.pm_trip_id,
```
mzappitello, line 340: was there an index update that made this subquery better than using the TempEventCompare table directly?

python_src/src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -373,7 +389,6 @@ def backup_rt_static_trip_match(
     # static trip are needed.
     first_stop_static_sub = (
         sa.select(
-            static_trips_sub.c.static_version_key,
```
mzappitello, line 376: guessing this is redundant since all these subqueries already have the static version key baked in?

python_src/src/lamp_py/migrations/versions/performance_manager_prod/001_5d9a7ee21ae5_initial_prod_schema.py
```diff
@@ -0,0 +1,966 @@
...
+                END IF;
+                NEW.trunk_route_id := NEW.route_id;
+            ELSE
+                NEW.branch_route_id := null;
+                NEW.trunk_route_id := NEW.route_id;
+            END IF;
+            RETURN NEW;
+        END;
+        $$ LANGUAGE plpgsql;
+    """
```
mzappitello, line 476: does it make sense to have this command as a stored as a string and insert it into these commands since its largely the same between the static and rt versions of this?

might not be worth it, but its always a code smell to see duplication.

python_src/src/lamp_py/migrations/versions/performance_manager_prod/001_5d9a7ee21ae5_initial_prod_schema.py
```diff
@@ -0,0 +1,966 @@
...
+    op.execute(create_get_rt_branch_trunk_id)
+
+    create_trigger = """
+        CREATE TRIGGER rt_trips_update_branch_trunk BEFORE UPDATE ON vehicle_trips 
+        FOR EACH ROW EXECUTE PROCEDURE update_rt_branch_trunk_id();
+    """
+    op.execute(create_trigger)
+
+    create_service_id_by_date_and_route = """
```
mzappitello, line 568: all of these are just direct reproductions?

mzappitello (commented):
I have a few questions, but looks good.

python_src/src/lamp_py/performance_manager/l0_gtfs_static_mod.py
rymarczy: It seems to be at least equally performative in my testing.

python_src/src/lamp_py/postgres/postgres_schema.py
```diff
@@ -43,14 +45,6 @@ class VehicleEvents(SqlBase):  # pylint: disable=too-few-public-methods
...
     updated_on = sa.Column(sa.TIMESTAMP, server_default=sa.func.now())
 
-    __mapper_args__ = {
-        "primary_key": [
-            service_date,
-            pm_trip_id,
-            parent_station,
-        ]
-    }
-
```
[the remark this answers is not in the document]
rymarczy: Yeah, we need the `pm_event_id` for the next/previous trip id fields, so might as well make it the primary key

python_src/src/lamp_py/postgres/postgres_schema.py
```diff
@@ -228,7 +234,19 @@ class StaticTrips(SqlBase):  # pylint: disable=too-few-public-methods
...
+    "ix_static_trips_composite_2",
+    StaticTrips.static_version_key,
+    StaticTrips.branch_route_id,
+)
+
+sa.Index(
+    "ix_static_trips_composite_3",
+    StaticTrips.static_version_key,
+    StaticTrips.trunk_route_id,
 )
```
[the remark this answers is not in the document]
rymarczy: Yes, these should be helping that query.

python_src/src/lamp_py/postgres/postgres_schema.py
```diff
@@ -247,7 +265,15 @@ class StaticRoutes(SqlBase):  # pylint: disable=too-few-public-methods
...
-    static_version_key = sa.Column(sa.Integer, nullable=False, index=True)
+    static_version_key = sa.Column(sa.Integer, nullable=False)
+
+
+sa.Index(
+    "ix_static_routes_composite_1",
+    StaticRoutes.static_version_key,
+    StaticRoutes.route_id,
+    StaticRoutes.route_type,
+)
```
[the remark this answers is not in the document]
rymarczy: We should have at least one index that starts with `static_version_key` on all static tables since that field is going to be used as a join key or where clause statement for just about any query involving them.

python_src/src/lamp_py/performance_manager/l1_rt_trips.py
rymarczy: Yes, limits to trips in the `TempEventCompare` table.

python_src/src/lamp_py/performance_manager/l1_rt_trips.py
rymarczy: I actually think this can be re-written. We don't need to join to the `VehicleTrips` table in this subquery. I'll update this.

python_src/src/lamp_py/performance_manager/l1_rt_trips.py
rymarczy: Yup

python_src/src/lamp_py/migrations/versions/performance_manager_prod/001_5d9a7ee21ae5_initial_prod_schema.py
rymarczy: They are direct reproductions of the most recent definitions from the old migration files.

python_src/src/lamp_py/migrations/versions/performance_manager_prod/001_5d9a7ee21ae5_initial_prod_schema.py
rymarczy: The rt vs static functions are slightly different enough that they couldn't be stored as the same string. The whole inserting giant strings with alembic into the RDS in general feels pretty messy, but I'm not sure if a better solution exists. 

It could be worth extracting sections of the command, like the list of stop_ids and storing those as a string, to be inserted. 

Might be worth having a quick call to discuss options.

python_src/tests/test_files/pipeline_flat_out.csv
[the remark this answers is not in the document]
rymarczy: correct

python_src/src/lamp_py/postgres/postgres_schema.py
```diff
@@ -192,6 +191,13 @@ class MetadataLog(SqlBase):  # pylint: disable=too-few-public-methods
     )
 
 
+sa.Index(
+    "ix_metadata_log_not_processed",
+    MetadataLog.path,
+    postgresql_where=(MetadataLog.processed == sa.false()),
+)
```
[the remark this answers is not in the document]
mzappitello: cool. thats a nice feature.

python_src/src/lamp_py/performance_manager/l0_gtfs_rt_events.py
```diff
@@ -500,7 +494,22 @@ def process_gtfs_rt_files(db_manager: DatabaseManager) -> None:
...
+                events = vp_events
+                events["tu_stop_timestamp"] = None
+            elif vp_events.shape[0] == 0:
+                # no vp events found
+                events = tu_events
+                events["vp_move_timestamp"] = None
+                events["vp_stop_timestamp"] = None
+            else:
+                # no events found
+                events = pandas.DataFrame()
```
rymarczy, line 512: I saw a few edge cases in our seed file processing that I think makes this work flow make more sense. We were assuming that if we have a parquet file for a specific hour we're processing, we would get events records from that file, but I don't think that is always the case.

python_src/src/lamp_py/performance_manager/l0_gtfs_rt_events.py
```diff
@@ -500,7 +494,22 @@ def process_gtfs_rt_files(db_manager: DatabaseManager) -> None:
...
+            elif tu_events.shape[0] == 0:
+                # no tu events found
+                events = vp_events
+                events["tu_stop_timestamp"] = None
+            elif vp_events.shape[0] == 0:
+                # no vp events found
+                events = tu_events
+                events["vp_move_timestamp"] = None
+                events["vp_stop_timestamp"] = None
```
rymarczy, line 512: I saw a few edge cases in our seed file processing that I think makes this work flow make more sense. We were assuming that if we have a parquet file for a specific hour we're processing, we would get events records from that file, but I don't think that is always the case.

mzappitello (approved):
LGTM 🍰

PR #156, opened 2023-07-31
Update GTFS_RT ingestion pipeline for older files

mzappitello:
GTFS_RT files generated from Sept 2019 through March 2020 have a different naming scheme and different extensions. Both changes break the way our ingestion pipeline currently works. Patch these issues and update our tests to check them.

* `ConfigType.from_filename` updated to handle new trip updates and vehicle position file names. (NOTE, I put all the other files from these dates through this function, and they were a ok.
* `GtfsRtConverter.pyarrow_from_gz` tries to choose the compression algo from the file extension, but tries to force a gz algo if that fails.

Asana Task: https://app.asana.com/0/1204931901750655/1205130684703916/f


rymarczy:
LGTM

PR #158, opened 2023-08-02
feat: Add GTFS ROUTE_PATTERNS table to schema and processing

rymarczy:
Per request from OPMI, `static_route_patterns` table added to RDS schema.

Additionally, `canonical_stop_sequence` added to `vehicle_events` table to represent an expected `stop_sequence` that will not change for an entire schedule rating. 

Asana Task: https://app.asana.com/0/1204931901750660/1205119273606571


python_src/src/lamp_py/migrations/versions/performance_manager_prod/001_5d9a7ee21ae5_initial_prod_schema.py
```diff
@@ -242,6 +242,25 @@ def upgrade() -> None:
         unique=False,
     )
 
+    op.create_table(
```
mzappitello, line 245: should probably create a similar migration in the dev dir, right?

python_src/src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -364,6 +367,110 @@ def update_directions(db_manager: DatabaseManager) -> None:
...
+    select_update_params = sa.select(
+        TempEventCompare.service_date,
+        TempEventCompare.static_version_key,
+    ).distinct()
+
+    for result in db_manager.select_as_list(select_update_params):
+        service_date = result["service_date"]
+        version_key = result["static_version_key"]
+
+        static_sub = (
```
mzappitello, line 383: I'm confused why we need the `StaticRoutePatterns` here to get the `stop_sequence` out of the `StaticStopTimes` since it seems like we're getting `StaticStopTimes.stop_sequence` by matching on `trip_id`, which is ultimately coming from the `StaticTrips` table?

mzappitello (commented):
I have a question on generating the canonical stop sequence from the route patterns and a comment on migrations, but the general shape all looks correct.

python_src/src/lamp_py/migrations/versions/performance_manager_prod/001_5d9a7ee21ae5_initial_prod_schema.py
rymarczy: I don't think we can. I'm not exactly sure how we'd deploy to dev at this point, since the schema changes combined with application logic to reflect those schema changes, has a lot of breaking changes with the existing dev schema.

python_src/src/lamp_py/performance_manager/l1_rt_trips.py
rymarczy: `StaticRoutePatterns` contains a list of trips to be considered "typical" for each route `route_pattern_typicality=1`. The end result being `representative_trip_id` that is joined to `StaticTrips` to get our `branch_route_id/trunk_route_id` designations that we do ourselves. 

These "typical" trips are joined to `StaticStopTimes` to get `stop_id/stop_sequence` values for each stop in each "typical" trip. 

Finally, the `stop_id` values pulled from `StaticStopTimes` are joined to `StaticStops` to get `parent_station` values we can join to our `vehicle_events` table. 

It's pretty convoluted, but it's the most direct path that I was able to identify.

mzappitello (approved):
Thanks for the walk through.

LGTM 🍰

PR #157, opened 2023-08-01
Add function for generating flat pyarrow table

mzappitello:
For some external users, we're going to provide flat parquet files via a data portal that can be used to run analysis on the data we've synthesized in the rail performance manager. Create a function that generates the pyarrow table with the appropriate data types for a given service date. This function will be run on a daily cadence, with the resulting table being written to a parquet file in an accessible s3 bucket.

Asana Task: https://app.asana.com/0/1204931901750655/1204907226653447/f

PR #159, opened 2023-08-08
Schedule flat file write 2

mzappitello:
Create method to write a flat file from the database for a given service date. It joins the Vehicle Events, Vehicle Positions, and Static Stops tables into a format third parties can use to read performance data.

Add flat file writing to performance manager event loop after new data is generated from gtfs realtime files. Write flat files for service dates from newly processed data.

Asana Task: https://app.asana.com/0/1204931901750655/1204907226653447/f
Asana Task: https://app.asana.com/0/1204931901750655/1204907226653449/f

### DIFF
--- a/python_src/src/lamp_py/performance_manager/flat_file.py
+++ b/python_src/src/lamp_py/performance_manager/flat_file.py
@@ -1,0 +1,162 @@
+import sqlalchemy as sa
+import pyarrow
+
+from lamp_py.performance_manager.gtfs_utils import (
+    static_version_key_from_service_date,
+)
+from lamp_py.postgres.postgres_schema import (
+    VehicleEvents,
+    VehicleTrips,
+    StaticStopTimes,
+    StaticStops,
+)
+from lamp_py.postgres.postgres_utils import DatabaseManager
+
+
+def generate_daily_table(
+    db_manager: DatabaseManager, service_date: int
+) -> pyarrow.Table:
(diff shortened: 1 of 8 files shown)


mzappitello:
This Replaces https://github.com/mbta/lamp/pull/157

python_src/src/lamp_py/performance_manager/flat_file.py
```diff
@@ -0,0 +1,192 @@
...
+)
+from lamp_py.postgres.postgres_utils import DatabaseManager
+
+
+def write_flat_files(
+    db_manager: DatabaseManager, dates: List[datetime.datetime]
+) -> None:
+    """write flat files to s3 for datetimes"""
+    dates_to_write = {get_service_date_from_timestamp(d) for d in dates}
+
```
rymarczy, line 27: I think there is a chance that the service_date pulled from this function won't match the service_date's processed during the event loop. But I'm not totally sure on the edge cases that could cause issues here. 

Might be worth talking through the different variations of possible service_dates's returned compared to service_date's processed based on timezone and time of day variations.

rymarczy (changes requested):
Looks pretty good, a few comments regarding the README documentation and a more general question regarding timestamps vs service_date for the writing of files in the event loop.

rymarczy (approved):
LGTM

PR #162, opened 2023-08-29
Write Flat Files to Public Archvie Bucket

mzappitello:
A new env var is used for the s3 bucket that we'll write our public flat files to. Write the flat files iff that env ver exists.

Asana Task: https://app.asana.com/0/1204931901750660/1204907226653452/f


rymarczy (approved):
LGTM

PR #163, opened 2023-08-29
Add logic for missing Start Times for Trips

mzappitello:
Asana Task: (https://app.asana.com/0/1204931901750665/1205259099679738/f)

Some older datasets do not provide start times for vehicle positions and
trip updates. For those datasets, we need to pull start times from the
static schedule (for planned trips) or compute it from the data we do
have (added trips which are, by definition, not in the static schedule).

* Add a function to `l1_rt_trips.py` that adds the start time to the
  TempVehicleTrips table.
* Call that function when running `proccess_trips`.


python_src/src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -217,6 +217,83 @@ def update_static_version_key(db_manager: DatabaseManager) -> None:
...
+    Add in start times to trips that do not have one. For scheduled trips,
+    look these up in the static schedule by finding the earliest Stop Time for
+    that trip id. Added trips cannot be found in the static schedule, so use
+    the earliest vehicle position information we have about them.
+    """
+    missing_start_times = (
+        sa.select(TempEventCompare.pm_trip_id)
+        .distinct()
+        .where(TempEventCompare.start_time.is_(None), TempEventCompare.new_trip)
+        .subquery("missing_start_times_sub")
```
rymarczy, line 231: I don't think we actually use this `TempEventCompare.new_trip` column in the pipeline. We should probably drop from the schema.

rymarczy, line 231: I think we'd want to join this to `VehicleTrips` and get records where `start_time` is null for both. 

```suggestion
        sa.select(TempEventCompare.pm_trip_id)
        .distinct()
        .join(
            VehicleTrips,
            VehicleTrips.pm_trip_id == TempEventCompare.pm_trip_id,
        )
        .where(
            TempEventCompare.start_time.is_(None), 
            VehicleTrips.start_time.is_(None),
        )
        .subquery("missing_start_times_sub")
```

python_src/src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -217,6 +217,83 @@ def update_static_version_key(db_manager: DatabaseManager) -> None:
...
+        .subquery("scheduled_start_times")
+    )
+
+    static_start_times_update_query = (
+        sa.update(VehicleTrips.__table__)
+        .where(
+            VehicleTrips.pm_trip_id == scheduled_start_times.c.pm_trip_id,
+        )
+        .values(
+            start_time=scheduled_start_times.c.static_start_time,
```
rymarczy, line 264: This `static_start_time` value will be in the format of String(HH:MM:SS) after midnight. But all of our other `start_time` values are stored as Integers of seconds after midnight. 

I think we could just a postures function convert these values on the fly.

python_src/src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -217,6 +217,83 @@ def update_static_version_key(db_manager: DatabaseManager) -> None:
...
+        sa.select(
+            TempEventCompare.pm_trip_id,
+            sa.func.min(TempEventCompare.vp_move_timestamp).label(
+                "rt_start_time"
+            ),
+        )
+        .where(TempEventCompare.start_time.is_(None), TempEventCompare.new_trip)
+        .group_by(
+            TempEventCompare.pm_trip_id,
+        )
```
rymarczy, line 280: This needs to be a `group_by` from the VehicleEvents table, because each trip could have previous events already in the VehcileEvents table that you would need to pull a min(vp_move_timstamp) from.

python_src/src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -217,6 +217,83 @@ def update_static_version_key(db_manager: DatabaseManager) -> None:
...
+    )
+
+    rt_start_times_update_query = (
+        sa.update(VehicleTrips.__table__)
+        .where(
+            VehicleTrips.pm_trip_id == unscheduled_start_times.c.pm_trip_id,
+        )
+        .values(
+            start_time=unscheduled_start_times.c.rt_start_time,
+        )
```
rymarczy, line 291: This `unscheduled_start_times.c.rt_start_time` value will be an ISO timstamp, but `start_time` should be an integer of seconds after midnight. We should be able to get this value with a native Postgres functions in combination with the `start_date` field.

python_src/src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -631,6 +708,7 @@ def process_trips(db_manager: DatabaseManager) -> None:
 
     """
     update_static_version_key(db_manager)
+    update_start_times(db_manager)
```
rymarczy, line 711: We should call this function after `update_static_trip_id_guess_exact`  since we are using the `VehicleTrips.trip_id` field in these queries.

rymarczy (changes requested):
I think we'll need to make use of some fancy built in postgres functions to do value conversions for this `start_time` field. I can try to sketch out some ideas for that, if you'd like.

python_src/src/lamp_py/performance_manager/gtfs_utils.py
```diff
@@ -16,9 +16,16 @@
 from lamp_py.aws.s3 import get_datetime_from_partition_path
 
 
+# boston tzinfo to be used with datetimes that require DST considerations.
+#
+# NOTE: do not use this as an argument for datetime.deatime constructors.
+# instead use BOSTON_TZ with a naive datetime. (https://pypi.org/project/pytz/)
+BOSTON_TZ = pytz.timezone("EST5EDT")
```
mzappitello, line 23: made a local var here for the boston timezone since we use it a few times.

left a note on the constructor issue.

python_src/src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -217,6 +220,100 @@ def update_static_version_key(db_manager: DatabaseManager) -> None:
...
+
+        start_times_update_query = (
+            sa.update(VehicleTrips.__table__)
+            .where(VehicleTrips.pm_trip_id == bindparam("b_pm_trip_id"))
+            .values(start_time=bindparam("b_start_time"))
+        )
+
+        db_manager.execute_with_data(
+            start_times_update_query, unscheduled_start_times
+        )
```
mzappitello, line 314: use a function to do the conversion on timestamps from travel times.

python_src/tests/performance_manager/test_performance_manager.py
```diff
@@ -524,6 +523,71 @@ def test_vp_and_tu(
     check_logs(caplog)
 
 
+def test_missing_start_time(
```
mzappitello, line 526: new test against a csv that has no start times.

python_src/tests/test_files/SPRINGBOARD/RT_VEHICLE_POSITIONS/year=2023/month=5/day=8/hour=12/04fae375300c45e2a0f70b8fbefa8f57-0.parquet
mzappitello, line 1: i took some start times out of this file to make sure it matched up with what we already had.

python_src/tests/test_resources.py
```diff
@@ -25,7 +25,11 @@ def csv_to_vp_parquet(csv_filepath: str, parquet_filepath: str) -> None:
             "start_time": pyarrow.string(),
             "vehicle_id": pyarrow.string(),
             "vehicle_consist": pyarrow.string(),
-        }
+        },
+        # in our ingestion, if a key is missing, the value written to the
+        # parquet file is null. mimic this behavior by making empty strings
+        # null instead of ''.
+        strings_can_be_null=True,
```
mzappitello, line 32: self explanatory. caught this in the new test.

python_src/src/lamp_py/performance_manager/gtfs_utils.py
```diff
@@ -28,16 +35,32 @@ def start_time_to_seconds(
...
+    month = int(service_date_string[4:6])
+    day = int(service_date_string[6:8])
+    start_of_service_day = BOSTON_TZ.localize(
+        datetime.datetime(year=year, month=month, day=day)
+    )
+
+    start_time = datetime.datetime.fromtimestamp(start_timestamp, tz=BOSTON_TZ)
+
+    time_after_midnight = start_time - start_of_service_day
+    return int(time_after_midnight.total_seconds())
```
rymarczy, line 53: nitpick: I think we could avoid some timezone tom foolery here, but I might be wrong. Since `start_timestamp` is a UTC timestamp, once we have the `service_date_string` we could get a UTC timestamp for that day at midnight and then our return value should just be the difference of the two UTC timestamps. 

`
return int(start_timestamp - service_date_utc_timestamp)
`

rymarczy (approved):
One nitpick/suggestion, but otherwise LGTM.

PR #164, opened 2023-09-07
Create a migration config for dev, staging, & prod

mzappitello:
We want to be able to specify migrations on an environmental basis. Create a versions folder for staging (thats currently a copy of prod) and move the generic performance_manager version folder to performance_manager_dev.

Additionally, add the alembic_db_name variable to the environment validation function in the performance manager pipeline and remove the fallback value, ensuring that it must be specified by the container running it.


python_src/alembic.ini
mzappitello, line 1: move `performance_manager` to `performance_manager_dev` and add `performance_manger_staging`

python_src/pyproject.toml
```diff
@@ -77,5 +77,6 @@ max-line-length = 80
 min-similarity-lines = 10
 # ignore session maker as it gives pylint fits
 # https://github.com/PyCQA/pylint/issues/7090
-ignored-classes = ['sqlalchemy.orm.session.sessionmaker' ]
-ignore = ["migrations"]
+ignored-classes = ['sqlalchemy.orm.session.sessionmaker']
+# ignore the migrations directory. its going to have duplication and _that is ok_.
+ignore-paths = ["^src/lamp_py/migrations/.*$"]
```
mzappitello, line 82: i don't think the `ignore` was doing anything before since the migrations dir didn't have an `__init__.py` file in it. once i added that (to get mypy to work properly), it started vomiting up issues. `ignore-paths` seems to work though.

python_src/src/lamp_py/ingestion/pipeline.py
```diff
@@ -38,7 +38,6 @@ def validate_environment() -> None:
         "DB_NAME",
         "DB_PORT",
         "DB_USER",
-        "INCOMING_BUCKET",
```
mzappitello, line 41: this was a duplicate i removed.

python_src/src/lamp_py/migrations/__init__.py
mzappitello, line 1: added to get mypy working.

python_src/src/lamp_py/performance_manager/pipeline.py
```diff
@@ -35,6 +35,8 @@ def validate_environment() -> None:
         "DB_NAME",
         "DB_PORT",
         "DB_USER",
+        "SERVICE_NAME",
+        "ALEMBIC_DB_NAME",
```
mzappitello, line 39: add these to validation.

python_src/src/lamp_py/performance_manager/pipeline.py
```diff
@@ -119,10 +121,9 @@ def start() -> None:
     # configure the environment
     os.environ["SERVICE_NAME"] = "performance_manager"
     validate_environment()
-    db_name = os.getenv("ALEMBIC_DB_NAME", "performance_manager")
 
     # run rds migrations
-    alembic_upgrade_to_head(db_name=db_name)
+    alembic_upgrade_to_head(db_name=os.getenv("ALEMBIC_DB_NAME"))
```
mzappitello, line 126: no more fallback, and pass the value directly into the alembic upgrade fall.

mzappitello (commented):
a few comments to clarify what i did and why.

all of the files in the migrations directory were either moved without change or are direct copies from prod to staging.

python_src/src/lamp_py/performance_manager/pipeline.py
rymarczy, line 126: I think `os.getenv` returns `None` if the env var is not found, and `alembic_upgrade_to_head` takes a string, so maybe have a dummy default to return if the `ALEMBIC_DB_NAME` env var is not found

```suggestion
    alembic_upgrade_to_head(db_name=os.getenv("ALEMBIC_DB_NAME", "db_not_found"))
```

rymarczy (changes requested):
The way we have alembic currently configured, any new `db_name` values also have to be added to the `get_alembic_config` function of `alembic_migration.py` and the `db_details` dictionary in `python_src/src/lamp_py/migrations/env.py`. 

I don't remember the specific reasons I configured everything like that originally, but I was probably assuming we wouldn't add these db_names frequently, but I'm up for any alternative architectures if you have ideas. 

The `ALEMBIC_DB_NAME` env var is also specified in our `ci_python.yaml` github action configuration. I'm not sure if that needs to be updated for if we want to run CI on a different migration, depending on the PR???

python_src/src/lamp_py/performance_manager/pipeline.py
mzappitello: this should be ok because we will throw for a missing env var before we get here.

rymarczy (approved):
LGTM

PR #165, opened 2023-09-11
feat: Suppress first stop record in `opmi_all_rt_fields_joined`

rymarczy:
Ops Analytics requested that the first stop of every trip from the `ompi_all_rt_fields_joined` VIEW be dropped as the information is not relevant to their metrics. This change has been incorporated by added the WHERE clause:

`
     ve.previous_trip_stop_pm_event_id is not NULL
`

to the VIEW

While looking into this ticket I also noticed an error with joining the `vehicle_events` table to itself to generate previous stop information for each record. The JOIN condition was providing next stop information, editing the ON statement has fixed this bug. 

`canonical_stop_sequence` was also added to the view as it was a previous request from Ops Analytics, that was implemented but the VIEW was not updated. 

Asana Task: https://app.asana.com/0/1204931901750670/1205316545733163


mzappitello:
LGTM

PR #171, opened 2023-10-06
feat: add `sync_stop_sequence` field to performance manager

rymarczy:
Add `sync_stop_sequence` field to performance manager based on OPMI request.

`sync_stop_sequence` is a generated stop_sequence value that is meant to be consistent across parent stations for different branch-routes on the same trunk-route.

Asana Task: https://app.asana.com/0/1205634094440864/1205670479715896

### DIFF
--- a/python_src/src/lamp_py/migrations/versions/performance_manager_prod/sql_strings/strings_001.py
+++ b/python_src/src/lamp_py/migrations/versions/performance_manager_prod/sql_strings/strings_001.py
@@ -387,6 +387,8 @@ view_opmi_all_rt_fields_joined = """
             , ve.stop_sequence
             , ve.canonical_stop_sequence
             , prev_ve.canonical_stop_sequence as previous_canonical_stop_sequence
+            , ve.sync_stop_sequence
+            , prev_ve.sync_stop_sequence as previous_sync_stop_sequence
             , ve.stop_id
             , prev_ve.stop_id as previous_stop_id
             , ve.parent_station
--- a/python_src/src/lamp_py/performance_manager/l1_rt_trips.py
+++ b/python_src/src/lamp_py/performance_manager/l1_rt_trips.py
@@ -469,107 +469,275 @@ def update_directions(db_manager: DatabaseManager) -> None:
     process_logger.log_complete()
 
 
-def update_canonical_stop_sequence(db_manager: DatabaseManager) -> None:
+def update_stop_sequence(db_manager: DatabaseManager) -> None:
     """
-    Update canonical_stop_sequence from static_route_patterns
+    Update canonical_stop_sequence  and sync_stop_sequence from static_route_patterns
     """
-    select_update_params = sa.select(
-        TempEventCompare.service_date,
-        TempEventCompare.static_version_key,
-    ).distinct()
+    static_canon = (
+        sa.select(
+            StaticRoutePatterns.direction_id,
+            StaticTrips.trunk_route_id,
+            sa.func.coalesce(
+                StaticTrips.branch_route_id, StaticTrips.trunk_route_id
+            ).label("route_id"),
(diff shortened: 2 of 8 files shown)


python_src/src/lamp_py/migrations/versions/performance_manager_prod/sql_strings/strings_001.py
```diff
@@ -387,6 +387,8 @@
             , ve.stop_sequence
             , ve.canonical_stop_sequence
             , prev_ve.canonical_stop_sequence as previous_canonical_stop_sequence
+            , ve.sync_stop_sequence
+            , prev_ve.sync_stop_sequence as previous_sync_stop_sequence
```
mzappitello, line 391: glad that this has to be included as well.

python_src/src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -469,107 +469,275 @@ def update_directions(db_manager: DatabaseManager) -> None:
...
+                sa.func.array(
+                    sa.select(TempEventCompare.static_version_key)
+                    .distinct()
+                    .scalar_subquery()
+                )
+            ),
+            StaticRoutePatterns.route_pattern_typicality == 1,
+        )
+        .subquery("static_canon")
+    )
```
mzappitello, line 526: this is roughly the same as the `static_sub` that was here before, but using multiple static version keys and adding in the `trunk_route_id` field?

python_src/src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -469,107 +469,275 @@ def update_directions(db_manager: DatabaseManager) -> None:
...
-            )
+    update_rt_canon = (
+        sa.update(VehicleEvents.__table__)
+        .values(
+            canonical_stop_sequence=rt_canon.c.stop_sequence,
         )
+        .where(
+            VehicleEvents.pm_event_id == rt_canon.c.pm_event_id,
+        )
+    )
```
mzappitello, line 573: so we're all the same as we were before here, right?

python_src/src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -469,107 +469,275 @@ def update_directions(db_manager: DatabaseManager) -> None:
...
+        .join(
+            sync_adjust_vals,
+            sa.and_(
+                sync_adjust_vals.c.direction_id == static_canon.c.direction_id,
+                sync_adjust_vals.c.trunk_route_id
+                == static_canon.c.trunk_route_id,
+            ),
+        )
+        .subquery(("sync_values"))
+    )
```
mzappitello, line 689: wow.

mzappitello (commented):
a question for clarification, but i think i get it.

its truly wild to have learned how to read an algorithm like this written in a pythonic SQL syntax.

python_src/src/lamp_py/performance_manager/l1_rt_trips.py
rymarczy: Correct, I did some profiling on the query and using this:
```
      StaticRoutePatterns.static_version_key
      == sa.func.any(
          sa.func.array(
              sa.select(TempEventCompare.static_version_key)
              .distinct()
              .scalar_subquery()
          )
      ),
```

Hits the index for `static_version_key`, instead of calling each `static_version_key` separately with a `where =`

python_src/src/lamp_py/performance_manager/l1_rt_trips.py
rymarczy: Yup, nothing changed here.

mzappitello:
LGTM 🍰

PR #174, opened 2023-10-11
FIX: Use Trip Updates to help compute start time

mzappitello:
When processing trips, we need to compute a start time for added trips. We've coded doing that based on the minimal vp move and vp stop timestamps for a given trip. However, this flow triggers a failure if there are only Trip Update records for this event. In those cases, use the minimal tu stop timestamp as well.

### DIFF
--- a/python_src/src/lamp_py/ingestion/convert_gtfs_rt.py
+++ b/python_src/src/lamp_py/ingestion/convert_gtfs_rt.py
@@ -85,7 +85,7 @@ class GtfsRtConverter(Converter):
         self.archive_files: List[str] = []
 
     def convert(self) -> None:
-        max_tables_to_convert = 12
+        max_tables_to_convert = 6
         process_logger = ProcessLogger(
             "parquet_table_creator",
             table_type="gtfs-rt",
--- a/python_src/src/lamp_py/ingestion/pipeline.py
+++ b/python_src/src/lamp_py/ingestion/pipeline.py
@@ -60,7 +60,7 @@ def main() -> None:
         check_for_sigterm(metadata_queue, rds_process)
         ingest(metadata_queue=metadata_queue)
         check_for_sigterm(metadata_queue, rds_process)
-        time.sleep(60 * 5)
+        time.sleep(5)
 
 
 def start() -> None:
--- a/python_src/src/lamp_py/performance_manager/gtfs_utils.py
+++ b/python_src/src/lamp_py/performance_manager/gtfs_utils.py
@@ -31,8 +31,14 @@ def start_time_to_seconds(
     """
     if time is None:
         return time
-    (hour, minute, second) = time.split(":")
-    return int(hour) * 3600 + int(minute) * 60 + int(second)
+
+    try:
+        (hour, minute, second) = time.split(":")
+        return int(hour) * 3600 + int(minute) * 60 + int(second)
+    except ValueError:
+        # some older files have the start time already formatted as seconds
+        # after midnight. in those cases, convert to an int and pass through.
+        return int(time)
 
 
 def start_timestamp_to_seconds(start_timestamp: int) -> int:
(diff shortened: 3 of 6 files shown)


mzappitello:
i tested this locally using parquet files from s3 and it worked as expected.

python_src/src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -85,7 +85,7 @@ def __init__(self, config_type: ConfigType, metadata_queue: Queue) -> None:
         self.archive_files: List[str] = []
 
     def convert(self) -> None:
-        max_tables_to_convert = 12
+        max_tables_to_convert = 6
```
mzappitello, line 88: i dropped this so that each iteration happened faster. happy to revert if you prefer.

python_src/src/lamp_py/ingestion/pipeline.py
```diff
@@ -60,7 +60,7 @@ def main() -> None:
         check_for_sigterm(metadata_queue, rds_process)
         ingest(metadata_queue=metadata_queue)
         check_for_sigterm(metadata_queue, rds_process)
-        time.sleep(60 * 5)
+        time.sleep(5)
```
mzappitello, line 63: no need to sleep for 5 minutes between event loops.

would also be down to eliminate this all together.

python_src/src/lamp_py/performance_manager/gtfs_utils.py
```diff
@@ -31,8 +31,14 @@ def start_time_to_seconds(
...
-    (hour, minute, second) = time.split(":")
-    return int(hour) * 3600 + int(minute) * 60 + int(second)
+
+    try:
+        (hour, minute, second) = time.split(":")
+        return int(hour) * 3600 + int(minute) * 60 + int(second)
+    except ValueError:
+        # some older files have the start time already formatted as seconds
+        # after midnight. in those cases, convert to an int and pass through.
+        return int(time)
```
mzappitello, line 41: just convert if we can't split the time. this should still surface other errors.

python_src/src/lamp_py/performance_manager/l0_rt_trip_updates.py
```diff
@@ -183,7 +183,7 @@ def process_tu_files(
     Generate a dataframe of Vehicle Events from gtfs_rt trip updates parquet files.
     """
     process_logger = ProcessLogger(
-        "process_trip_updates", file_count=len(paths)
+        "process_trip_updates", file_count=len(paths), paths=paths
```
mzappitello, line 186: added for debugging help.

python_src/src/lamp_py/performance_manager/l0_rt_vehicle_positions.py
```diff
@@ -218,7 +218,7 @@ def process_vp_files(
     Generate a dataframe of Vehicle Events from gtfs_rt vehicle position parquet files.
     """
     process_logger = ProcessLogger(
-        "process_vehicle_positions", file_count=len(paths)
+        "process_vehicle_positions", file_count=len(paths), paths=paths
```
mzappitello, line 221: added for debugging help

python_src/src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -290,6 +290,7 @@ def update_start_times(db_manager: DatabaseManager) -> None:
             sa.func.coalesce(
                 sa.func.min(VehicleEvents.vp_move_timestamp),
                 sa.func.min(VehicleEvents.vp_stop_timestamp),
+                sa.func.min(VehicleEvents.tu_stop_timestamp),
```
mzappitello, line 293: needed when events only have trip updates values.

mzappitello (commented):
pretty straight forward. i'm down for other approaches, but this feels good enough to get to the next phase of the project.

outside of the errors patched here, i haven't seen any problems on staging and have confirmed that locally.

rymarczy (approved):
LGTM

PR #177, opened 2023-10-26
Change rail-pm task count from github action

mzappitello:
We sometimes want to change the task count for the rail performance manager as well. Add an input that lets us specify the application we're adjusting the task count on.

Use the choice type to create drop-downs on the github page.


.github/workflows/change_ingestion_task_count.yaml
mzappitello, line 1: moved this since its not just ingestion.

.github/workflows/change_task_count.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      environment:
+        type: choice
+        description: What environment to change the task count for
+        options:
+          - dev
+          - staging
```
mzappitello, line 11: use a choice rather than having us enter the string. it should create a drop down on the github action ui.

.github/workflows/change_task_count.yaml
```diff
@@ -0,0 +1,35 @@
...
+      new_count:
+        description: 1 to turn on 0 to turn off
+        required: true
+      application_name:
+        type: choice
+        description: What application to adjust the task count for
+        options:
+          - ingestion
+          - rail-performance-manager
```
mzappitello, line 20: let us select what application we're adjusting the task count of.

.github/workflows/change_task_count.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+jobs:
+  set_count:
+    if: |
+      ( github.event.inputs.new_count == 0 || github.event.inputs.new_count == 1)
```
mzappitello, line 25: choice only seems to work with ints (at least my linter yelled at me) so i left these as a manual entry. keeping this if gate to prevent us from setting the task count to 11 or something.

.github/workflows/change_task_count.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: run command
+        run: aws ecs update-service --cluster lamp --service lamp-${{ github.event.inputs.application_name }}-${{ github.event.inputs.environment }} --desired-count ${{ github.event.inputs.new_count }}
```
mzappitello, line 31: added the application name to the update service call.

rymarczy (approved):
LGTM

PR #179, opened 2023-10-27
FIX: Tableau Server Import Exception 

rymarczy:
`tableau_server` function call being in function signature was causing Exception at import time. Move function calls out of parameter declaration. 

Also move HyperJob creation into function to avoid creation of jobs at import.


mzappitello:
LGTM 🍰

PR #180, opened 2023-11-08
Run vacuum analyze on static tables

mzappitello:
Some of our subquery execution times have blown up and the assumption is that they will be more performative if we run `VACUUM (ANALYZE)` on static tables after a static schedule is processed.

* add a new function to run vacuum analyze on a table to the DatabaseManager class.
* use that function on static tables after new data is loaded.
* use that function on vehicle events and vehicle trips tables after processing a batch of gtfs_rt data.


rymarczy (approved):
LGTM :cake:

PR #183, opened 2023-11-13
FIX: Stream DB Results to Parquet Files

rymarczy:
The current Parquet -> Tableau pipeline process is flawed in the amount of memory required to create parquet files from DB SELECT queries. This change is meant to result in a fixed amount of memory usage, no matter what the number of results are returned from a DB query, when creating a parquet file. 

This fixed memory usage is achieved by utilizing the `yield_per` method of the SQLAlchemy `Result` object, as well as the `RecordBatch` object of the pyarrow library. 

In testing, memory usage for the creation of a parquet file from the `static_stop_times` table maxes out at approximately 5-6GB. 

If memory usage needs to be further limited, the `write_to_parquet` function of `DatabaseManager` offers a `batch_size` parameter to limit the number for records flowing into a parquet file per partition. 

Asana Task: https://app.asana.com/0/1205827492903547/1205940053804614

### DIFF
--- a/python_src/src/lamp_py/postgres/postgres_utils.py
+++ b/python_src/src/lamp_py/postgres/postgres_utils.py
@@ -283,7 +283,7 @@ class DatabaseManager:
         write_path: str,
         schema: pyarrow.schema,
         batch_size: int = 1024 * 1024,
-    ) -> str:
+    ) -> None:
         """
         stream db query results to parquet file in batches
 
@@ -300,8 +300,6 @@ class DatabaseManager:
         :param write_path: local file path for resulting parquet file
         :param schema: schema of parquet file from select query
         :param batch_size: number of records to stream from db per batch
-
-        :return local path to created parquet file
         """
         with self.session.begin() as cursor:
             result = cursor.execute(select_query).yield_per(batch_size)
@@ -312,8 +310,6 @@ class DatabaseManager:
                             [row._asdict() for row in part], schema=schema
                         )
                     )
-
-        return write_path
 
     def truncate_table(
         self,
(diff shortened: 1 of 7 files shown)


python_src/src/lamp_py/postgres/postgres_utils.py
```diff
@@ -86,12 +95,14 @@ def get_local_engine(
...
+        # on mac, when running in docker locally db is accessed by "0.0.0.0" ip
         if db_host == "local_rds" and "macos" in platform.platform().lower():
             db_host = "0.0.0.0"
-        # if not running_in_docker():
-        #     db_host = "127.0.0.1"
+
+        # when running application locally in CLI for configuration
+        # and debugging, db is accessed by localhost ip
+        if not running_in_docker() and not running_in_aws():
+            db_host = "127.0.0.1"
```
rymarczy, line 105: pulled this over from `dmap_import`

python_src/src/lamp_py/postgres/postgres_utils.py
```diff
@@ -266,6 +277,44 @@ def select_as_list(
...
+            result = cursor.execute(select_query).yield_per(batch_size)
+            with pq.ParquetWriter(write_path, schema=schema) as pq_writer:
+                for part in result.partitions():
+                    pq_writer.write_batch(
+                        pyarrow.RecordBatch.from_pylist(
+                            [row._asdict() for row in part], schema=schema
+                        )
+                    )
+
+        return write_path
```
mzappitello, line 316: i don't think we need to return the write path since it was provided as part of the input.

mzappitello (changes requested):
nice. excited to see this in staging.

mzappitello (approved):
lgtm 🍰

PR #184, opened 2023-11-14
FIX: LAMP_ALL_RT_fields incorrect SQL comparison

rymarczy:
```
 max_stats["service_date"]
```
statement was returning `datetime.date` object that was being piped direction into SQL statement for comparison to `vehicle_trips.service_date` which is an `Integer` field in the format of `YYYYMMDD`.

This change converts the `datetime.date` object into string representation of `YYYYMMDD` for the SQL statement comparison.


mzappitello (commented):
🍰

PR #187, opened 2023-11-14
FIX: Update how Secrets are used in Base Deploy

mzappitello:
The secrets should be as they are in the the repo's secrets. Additionally, they need to be accessed via the `secrets` variable, not the `inputs` variable.


rymarczy (approved):
:cake:

PR #190, opened 2023-11-15
FIX: More bugs in Manual Deploy Workflow

mzappitello:
* Inputs need to be passed in from the workflow dispatch to the base workflow as booleans, so use a fromJson function to flip them.
* Rename the env-name input to environment so that it matches with the workflow dispatch name.
* Pass the environment string, not the name of the variable.

* Add prod to list of environments we can change the task count for.

### DIFF
--- a/.github/workflows/change_task_count.yaml
+++ b/.github/workflows/change_task_count.yaml
@@ -9,6 +9,7 @@ on:
         options:
           - dev
           - staging
+          - prod
       new_count:
         description: 1 to turn on 0 to turn off
         required: true
--- a/.github/workflows/deploy-base.yaml
+++ b/.github/workflows/deploy-base.yaml
@@ -8,7 +8,7 @@
 on:
   workflow_call:
     inputs:
-      env-name:
+      environment:
         description: One of 'prod', 'staging', or 'dev'
         required: true
         type: string
@@ -47,14 +47,12 @@ jobs:
     steps:
       - name: Checkout Branch
         uses: actions/checkout@v3
-
       - name: Configure AWS credentials
         id: setup-aws
         uses: aws-actions/configure-aws-credentials@v4
         with:
           role-to-assume: ${{ secrets.AWS_ROLE_ARN }}
           aws-region: us-east-1
-
       - name: Build and Push Docker Image
         id: build-push
         uses: mbta/actions/build-push-ecr@v2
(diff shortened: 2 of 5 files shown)


.github/workflows/change_task_count.yaml
```diff
@@ -9,6 +9,7 @@ on:
         options:
           - dev
           - staging
+          - prod
```
mzappitello, line 12: we can change the task count for prod now.

.github/workflows/deploy-base.yaml
```diff
@@ -8,7 +8,7 @@
 on:
   workflow_call:
     inputs:
-      env-name:
+      environment:
```
mzappitello, line 11: rename this input, with cascading changes in the deploy workflows that call `deploy-base.yaml`

.github/workflows/manual-deploy.yaml
```diff
@@ -25,13 +25,14 @@ on:
...
       group: github.event.inputs.environment
     uses: ./.github/workflows/deploy-base.yaml
     with:
-      env-name: github.event.inputs.environment
-      deploy-ingestion: github.envent.inputs.deploy-ingestion
-      deploy-rail-pm: github.envent.inputs.deploy-rail-pm
-      deploy-tableau-publisher: github.envent.inputs.deploy-tableau-publisher
+      # pass the inputs from the workflow dispatch through to the deploy base. the booleans are
+      # converted to strings, so flip them back using fromJson function
+      environment: ${{ github.event.inputs.environment }}
```
mzappitello, line 34: i was passing the string of the variable name before 😵‍💫

.github/workflows/manual-deploy.yaml
```diff
@@ -25,13 +25,14 @@ on:
...
+      deploy-ingestion: ${{ fromJson(github.event.inputs.deploy-ingestion) }}
+      deploy-rail-pm: ${{ fromJson(github.event.inputs.deploy-rail-pm) }}
+      deploy-tableau-publisher: ${{ fromJson(github.event.inputs.deploy-tableau-publisher) }}
```
mzappitello, line 37: using `fromJson` converts these from strings back to bools.

rymarczy: would you want the `fromJson` on the inside or the outside of `${{..}}` here?

rymarczy (approved):
One question, but otherwise LGTM

.github/workflows/manual-deploy.yaml
mzappitello: expressions go inside.

https://docs.github.com/en/actions/learn-github-actions/expressions#example-returning-a-json-data-type

i still find all of the syntax rules a bit hard to follow, but i'm sure they stem from some weird backend interpreter they have to use.

PR #193, opened 2023-11-27
FIX: trip_id varchar length to 512

rymarczy:
A GTFS Static Schedule was issued with a character length that exceeds 128. This caused a failure in our GTFS static schedule loading. 

This change updates the RDS migrations on our DEV server to extend the length of all trip_id fields and reset database state for the failed schedule processing.

Asana Task: https://app.asana.com/0/1205827492903547/1206030596372460


mzappitello (approved):
LGTM

PR #197, opened 2023-12-11
FEAT: Add `stop_name` and `previous_stop_name` values to LAMP_ALL_RT_fields hyper file

rymarczy:
Add `stop_name` and `previous_stop_name` values to LAMP_ALL_RT field hyper  file extract. 

These fields are joined from `static_stops` table.

### DIFF
--- a/python_src/src/lamp_py/tableau/jobs/rt_rail.py
+++ b/python_src/src/lamp_py/tableau/jobs/rt_rail.py
@@ -35,6 +35,8 @@ class HyperRtRail(HyperJob):
             "   , prev_ve.stop_id as previous_stop_id"
             "   , ve.parent_station"
             "   , prev_ve.parent_station as previous_parent_station"
+            "   , ss.stop_name as stop_name"
+            "   , prev_ss.stop_name as previous_stop_name"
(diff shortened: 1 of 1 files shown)


python_src/src/lamp_py/tableau/jobs/rt_rail.py
```diff
@@ -71,6 +73,16 @@ def __init__(self) -> None:
             "   vehicle_events prev_ve"
             " ON "
             "   ve.pm_event_id = prev_ve.next_trip_stop_pm_event_id"
+            " LEFT JOIN "
+            "   static_stops ss"
```
mzappitello, line 77: just noticing this, but is the `AS` keyword unnecessary?

mzappitello (approved):
LGTM

python_src/src/lamp_py/tableau/jobs/rt_rail.py
rymarczy: It's generally considered to be optional: https://www.postgresql.org/docs/current/queries-table-expressions.html#QUERIES-TABLE-ALIASES

PR #198, opened 2023-12-11
FIX: Update ALL_RT_fields to fix UTC timezone adjustement on seconds fields

rymarczy:
Fix for following feedback from Heather Davis on LAMP_ALL_RT_fields Hyper file extract: 

> In the all RT fields view, the stop arrival sec and previous stop departure sec fields are not calculated properly -- they seem to be in GMT rather than EST/EDT. Pre-11/5/23, they are 14,400 seconds or 4 hours too large, and post-11/5/23 they are 18,000 seconds or 5 hours too large. The datetime fields are in the correct time zone and do account for DST.

> The stop departure datetime field does not have an associated field in seconds after midnight (i.e. stop departure sec). Can you please add this field?


mzappitello (approved):
LGTM

PR #199, opened 2023-12-11
FIX: Check Project of Datasource Name

rymarczy:
When finding a Tableau datasource, based on name, datasources with the same name, in multiple projects could result in unexpected hyper file update behavior. 

`datasource_from_name` function updated to also match datasource based on expected tableau project name.

### DIFF
--- a/python_src/src/lamp_py/tableau/hyper.py
+++ b/python_src/src/lamp_py/tableau/hyper.py
@@ -40,7 +40,7 @@ class HyperJob(ABC):  # pylint: disable=R0902
         self,
         hyper_file_name: str,
         remote_parquet_path: str,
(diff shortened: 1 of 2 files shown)


python_src/src/lamp_py/tableau/hyper.py
```diff
@@ -40,7 +40,7 @@ def __init__(
         self,
         hyper_file_name: str,
         remote_parquet_path: str,
-        project_name: str = "Staging Data",
+        project_name: str = "LAMP API",
```
rymarczy, line 43: Tableau PROD Project Name for LAMP data

mzappitello (approved):
lgtm

PR #201, opened 2023-12-12
Metadata rds

mzappitello:
Create a Metadata RDS for the project that can be used by multiple applications to query the status of files stored on S3

* move postgres_schema into rpm and metadata schemas
* create methods for connecting to both of them at the same time
* update ingestion and performance manager to use both of them
* add flow for copying old metadata over from rpm to metadata

Asana Task: https://app.asana.com/0/1189492770004753/1205798784639666/f


.env
mzappitello, line 1: new env variables used to distinguish what db connection params are for.

.github/workflows/ci_python.yaml
mzappitello, line 1: two dbs are needed by the performance manager tests.

docker-compose.yml
mzappitello, line 1: two dbs in local docker compose.

python_src/alembic.ini
```diff
@@ -78,6 +78,11 @@ sqlalchemy.url = driver://user:pass@localhost/dbname
 script_location = src/lamp_py/migrations
 version_locations = src/lamp_py/migrations/versions/performance_manager_prod
 
+[metadata_prod]
+sqlalchemy.url = driver://user:pass@localhost/dbname
+script_location = src/lamp_py/migrations
+version_locations = src/lamp_py/migrations/versions/metadata_prod
+
```
mzappitello, line 85: add metadata db migrations.

python_src/pyproject.toml
```diff
@@ -12,7 +12,6 @@ authors = [
 ingestion = 'lamp_py.ingestion.pipeline:start'
 performance_manager = 'lamp_py.performance_manager.pipeline:start'
 seed_metadata = 'lamp_py.postgres.seed_metadata:run'
-snapshot = 'lamp_py.postgres.snapshot:run'
```
mzappitello, line 15: we don't use this so i removed it.

python_src/pyproject.toml
```diff
@@ -80,6 +79,6 @@ max-line-length = 80
 min-similarity-lines = 10
 # ignore session maker as it gives pylint fits
 # https://github.com/PyCQA/pylint/issues/7090
-ignored-classes = ['sqlalchemy.orm.session.sessionmaker','pyarrow.compute']
+ignored-classes = ['sqlalchemy.orm.session.sessionmaker', 'pyarrow.compute']
```
mzappitello, line 82: yaml linting

python_src/src/lamp_py/ingestion/pipeline.py
```diff
@@ -81,6 +85,11 @@ def start() -> None:
         validate_db=True,
     )
 
+    # run metadata rds migrations
+    alembic_upgrade_to_head(db_name="metadata_prod")
+
```
mzappitello, line 90: migrations for the metadata db are owned by ingestion. this should make our deployment experience better.

python_src/src/lamp_py/ingestion/pipeline.py
```diff
@@ -81,6 +85,11 @@ def start() -> None:
...
+    move_metadata()
```
mzappitello, line 91: temp function to move all the metadata over from rail performance manager. we should only have to do this once.

python_src/src/lamp_py/migrations/versions/metadata_prod/07903947aabe_initial_changes.py
mzappitello, line 1: initial migration to establish the metadata table in the database.

python_src/src/lamp_py/performance_manager/l0_gtfs_rt_events.py
```diff
@@ -425,7 +425,10 @@ def update_events_from_temp(db_manager: DatabaseManager) -> None:
     process_logger.log_complete()
 
 
-def process_gtfs_rt_files(db_manager: DatabaseManager) -> None:
+def process_gtfs_rt_files(
```
mzappitello, line 428: we need both dbs now. one to figure out what files we have to extract and one to do the compute for the tranfromations.

python_src/src/lamp_py/performance_manager/l0_gtfs_static_load.py
```diff
@@ -426,7 +426,10 @@ def insert_data_tables(
         raise error
 
 
-def process_static_tables(db_manager: DatabaseManager) -> None:
+def process_static_tables(
+    rpm_db_manager: DatabaseManager,
+    md_db_manager: DatabaseManager,
+) -> None:
```
mzappitello, line 432: we need the md db manager to figure out what the new feed info files are and the rpm db manager to load the new data into.

python_src/src/lamp_py/performance_manager/pipeline.py
```diff
@@ -46,13 +49,29 @@ def parse_args(args: List[str]) -> argparse.Namespace:
...
+    """
+    are there any records in the legacy metadata table
+    """
+    metadata = rpm_db_manager.select_as_dataframe(
+        sa.select(LegacyMetadataLog.path)
+    )
+
+    return not metadata.empty
+
+
```
mzappitello, line 62: check to see if there is legacy metadata that hasn't been copied over yet.

python_src/src/lamp_py/performance_manager/pipeline.py
```diff
@@ -66,18 +85,23 @@ def iteration() -> None:
...
 
         try:
-            process_static_tables(db_manager)
-            process_gtfs_rt_files(db_manager)
-            write_flat_files(db_manager)
-            start_parquet_updates(db_manager)
+            # temporary check to see if legacy metadata exists. if it does,
+            # raise an exception and try again on the next cycle.
+            if legacy_metadata_exists(rpm_db_manager):
+                raise EnvironmentError("Legacy Metadata Detected")
```
mzappitello, line 91: raise an exception if there is legacy metadata. this will prevent us from running any of the performance manager logic, waiting for the next loop to give it a shot.

python_src/src/lamp_py/postgres/rail_performance_manager_schema.py
```diff
@@ -3,10 +3,10 @@
 import sqlalchemy as sa
 from sqlalchemy.ext.declarative import declarative_base
 
-SqlBase: Any = declarative_base()
+RpmSqlBase: Any = declarative_base()
```
mzappitello, line 6: new base, one for each database.

python_src/src/lamp_py/postgres/rail_performance_manager_schema.py
```diff
@@ -177,7 +177,7 @@ class TempEventCompare(SqlBase):  # pylint: disable=too-few-public-methods
     )
 
 
-class MetadataLog(SqlBase):  # pylint: disable=too-few-public-methods
+class LegacyMetadataLog(RpmSqlBase):  # pylint: disable=too-few-public-methods
```
mzappitello, line 180: rename this to avoid two tables with the same name.

python_src/src/lamp_py/postgres/snapshot.py
mzappitello, line 1: we weren't using this.

python_src/tests/performance_manager/test_performance_manager.py
mzappitello, line 1: no functional changes. just creating the db managers and passing them in correctly and resetting the correct tables.

python_src/src/lamp_py/ingestion/pipeline.py
rymarczy: I think we would want to do this in the migration file for the new metadata RDS, which essentially operates as a "temp" function.

python_src/src/lamp_py/migrations/versions/metadata_prod/07903947aabe_initial_changes.py
```diff
@@ -0,0 +1,54 @@
...
+    )
+    # ### end Alembic commands ###
+
+
+def downgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_index(
+        "ix_metadata_log_not_processed",
+        table_name="metadata_log",
+        postgresql_where=sa.text("rail_pm_processed = false"),
```
rymarczy, line 51: I don't think this should be required when calling to drop the index, just the index name and table name should be sufficient.

python_src/src/lamp_py/ingestion/pipeline.py
rymarczy: This does complicate local development, since we don't have/don't really want to run a local ingestion image, standing up the local metadata rds would need a different configuration.

One option would be to allow the 'seed_metadata` image to do this, but it would require some re-working of that logic (probably some different env vars)

python_src/src/lamp_py/postgres/rail_performance_manager_schema.py
rymarczy, line 180: Should we just be dropping this table?

python_src/src/lamp_py/performance_manager/pipeline.py
```diff
@@ -46,13 +49,29 @@ def parse_args(args: List[str]) -> argparse.Namespace:
...
 
+def legacy_metadata_exists(rpm_db_manager: DatabaseManager) -> bool:
+    """
+    are there any records in the legacy metadata table
+    """
+    metadata = rpm_db_manager.select_as_dataframe(
+        sa.select(LegacyMetadataLog.path)
+    )
+
+    return not metadata.empty
```
rymarczy, line 60: I don't think this is really required. If everything is the performance manager pipeline is now configured to read from the new MetadataLog table, whether or not the LegacyMetadataLog exists shouldn't matter.

rymarczy (changes requested):
I think we need to discuss the migration process in a little more detail along with the plan for local development with the MetadataLog table residing in a different DB instance. 

Before deployment, we should also change the configuration of the metadata-rds terraform module. The current metadata log table on the dev instance  is 34 MB in size, so something like an `allocated_storage` of 2 gibibytes and `max_allocated_storage` of 5 should be sufficient. The `m5.large` instance class is also probably overkill for a DB with one table, should drop that down to something like `t3.small`

docker-compose.yml
```diff
@@ -2,16 +2,28 @@ version: '3'
...
+      POSTGRES_DB: ${RPM_DB_NAME}
+      POSTGRES_PASSWORD: ${RPM_DB_PASSWORD}
     ports:
-      - "5432:5432"
+      - "${RPM_DB_PORT}:5432"
+    command: ["postgres", "-c", "log_statement=all"]
+
+  metadata_rds:
+    container_name: metadata_rds
+    image: postgres:14.4
```
rymarczy, line 19: The terraform module has postgres 15 set for `metadata-rds` need to sync this between the two.

python_src/src/lamp_py/ingestion/pipeline.py
mzappitello: good note. lets discuss in person, but i think adding to `seed_metadata` makes sense.

python_src/src/lamp_py/ingestion/pipeline.py
mzappitello: my instinct was that it would be wrong to have a connection to the RPM database in the migration file as that migration could run when there is no longer a metadata table in that database.

python_src/src/lamp_py/performance_manager/pipeline.py
mzappitello: i was worried about some deployment error on our part and wanted to protect against that. i had planned on taking this out on a subsequent PR where everything is complete.

python_src/src/lamp_py/postgres/rail_performance_manager_schema.py
mzappitello: i think that depends on the deployment order.

we'll discuss in a call.

python_src/src/lamp_py/ingestion/pipeline.py
```diff
@@ -88,8 +85,6 @@ def start() -> None:
     # run metadata rds migrations
     alembic_upgrade_to_head(db_name="metadata_prod")
 
-    move_metadata()
```
mzappitello: do this in migration now.

python_src/src/lamp_py/migrations/versions/metadata_prod/07903947aabe_initial_changes.py
```diff
@@ -40,6 +45,52 @@ def upgrade() -> None:
         unique=False,
         postgresql_where=sa.text("rail_pm_processed = false"),
     )
+
+    # pull metadata from the rail performance manager database into the
```
mzappitello, line 49: i think this works how we would like. it uses a test based query to trya and select metadata from the rail performance manager and eats an exception iff its caused by an "undefined table" error, i.e. the metadata log table does not exist.

python_src/src/lamp_py/migrations/versions/performance_manager_staging/005_96187da84955_remove_metadata.py
```diff
@@ -0,0 +1,67 @@
+"""remove_metadata
```
mzappitello, line 1: new migration to remove metadata table from performance manager db.

python_src/src/lamp_py/performance_manager/pipeline.py
```diff
@@ -9,11 +9,8 @@
 import signal
 from typing import List
 
-import sqlalchemy as sa
```
mzappitello: reset this file back to how it was.

python_src/src/lamp_py/postgres/postgres_utils.py
```diff
@@ -176,25 +175,25 @@ class DatabaseIndex(Enum):
...
             return "MD"
         raise NotImplementedError("No environment prefix for index {self.name}")
 
     def get_args_from_env(self) -> PsqlArgs:
         """
         generate a sql argument instance for this ind
         """
-        prefix = self.get_env_prefix()
-        return PsqlArgs(prefix)
+        return PsqlArgs(self.env_prefix)
```
mzappitello, line 196: more correct way of doing this. (and fix the stupid auto typo i left)

python_src/src/lamp_py/postgres/postgres_utils.py
```diff
@@ -404,52 +403,16 @@ def vacuum_analyze(self, table: Any) -> None:
...
-    def add_metadata_paths(self, paths: List[str]) -> None:
-        """
-        add metadata filepaths to metadata table for testing
-        """
-        print(paths)
-        with self.session.begin() as session:
-            session.execute(
-                sa.insert(MetadataLog.__table__),
-                [{"path": p} for p in paths],
-            )
```
mzappitello: moved this out of db manager class since it only makes sense for one of the databases.

python_src/src/lamp_py/postgres/rail_performance_manager_schema.py
```diff
@@ -177,27 +177,6 @@ class TempEventCompare(RpmSqlBase):  # pylint: disable=too-few-public-methods
     )
 
 
-class LegacyMetadataLog(RpmSqlBase):  # pylint: disable=too-few-public-methods
```
mzappitello: 👋

python_src/src/lamp_py/performance_manager/pipeline.py
```diff
@@ -52,7 +52,12 @@ def main(args: argparse.Namespace) -> None:
     main_process_logger.log_start()
 
     # get the engine that manages sessions that read and write to the db
-    db_manager = DatabaseManager(verbose=args.verbose)
+    rpm_db_manager = DatabaseManager(
+        db_index=DatabaseIndex.RAIL_PERFORMANCE_MANAGER, verbose=args.verbose
+    )
+    md_db_manager = DatabaseManager(
+        db_index=DatabaseIndex.RAIL_PERFORMANCE_MANAGER, verbose=args.verbose
+    )
```
rymarczy, line 60:
```suggestion
    md_db_manager = DatabaseManager(
        db_index=DatabaseIndex.METADATA, verbose=args.verbose
    )
```

rymarczy (changes requested):
Most changes related to getting the local environment up and running correctly.

python_src/src/lamp_py/migrations/versions/metadata_prod/07903947aabe_initial_changes.py
```diff
@@ -40,6 +45,52 @@ def upgrade() -> None:
...
+            )
+            for row in result:
+                (path, processed, process_fail) = row
+                insert_data.append(
+                    {
+                        "path": path,
+                        "rail_pm_processed": processed,
+                        "rail_pm_process_fail": process_fail,
+                    }
+                )
```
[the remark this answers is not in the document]
rymarczy: Ah, nevermind, I see you're renaming the dictionary values here.

python_src/src/lamp_py/migrations/versions/performance_manager_staging/005_96187da84955_remove_metadata.py
```diff
@@ -0,0 +1,67 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "96187da84955"
+down_revision = "45dedc21086e"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
```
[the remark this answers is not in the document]
rymarczy: Actually, we need to check if records exist in the new metadata log table before executing the drop table command, otherwise the table could be dropped before the records are transferred to the new metadata_log table.

python_src/src/lamp_py/migrations/versions/metadata_prod/07903947aabe_initial_changes.py
```diff
@@ -40,6 +45,52 @@ def upgrade() -> None:
...
+
+        # insert data into the metadata database
+        with md_db_manager.session.begin() as session:
+            # do nothing on conflicts in file paths
+            result = session.execute(
+                psql_insert(MetadataLog.__table__)
+                .values(insert_data)
+                .on_conflict_do_nothing()
+            )
```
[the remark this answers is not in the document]
mzappitello: i'm a little concerned in a case where this migration gets run out of sync with the migration on the production manager database and we find ourselves in a situation where we can't do anything to unblock prod since we can't manually adjust the databases. ingestion should only ever have flags marked as false, so i don't think we would be at risk of missing processing of incoming data.

happy to revert this though if you think its risky.

python_src/src/lamp_py/runtime_utils/env_validation.py
```diff
@@ -8,7 +8,7 @@ def validate_environment(
     required_variables: List[str],
     private_variables: Optional[List[str]] = None,
     optional_variables: Optional[List[str]] = None,
-    validate_db: bool = False,
+    db_prefixes: Optional[List[str]] = None,
```
rymarczy, line 11: nitpick:
```suggestion
    db_prefixes: Iterable[str] = (),
```

I know this is following the convention of `private_variables` and `optional_variables` because pylint complains of a default parameter values of `[]` but `Iterable[str] = ()` could be used for all of those parameters and we could drop the `if` checks before our `for` loops and setting `private_variables=[]` at the start of the function.

python_src/src/lamp_py/migrations/versions/performance_manager_staging/005_96187da84955_remove_metadata.py
```diff
@@ -0,0 +1,96 @@
...
+            raise
+
+    modern_result = md_db_manager.select_as_list(sa.select(MetadataLog.path))
+    modern_paths = set([record["path"] for record in modern_result])
+
+    missing_paths = legacy_paths - modern_paths
+    if len(missing_paths) != 0:
+        raise Exception(
+            f"Detected {len(missing_paths)} in Legacy Metadata Table not found in Metadata Database."
+        )
```
rymarczy, line 58: non-blocking:

This failure mechanism will result in the ECS dying an continually restarting if performance manager is trying to run before the ingestion migration completes. An alternate would be this check running in a `while` loop and breaking once there are no outstanding `legacy_paths`, compared to `modern_paths`.

rymarczy (changes requested):
Almost there!

The last significant issue is the insert operation on the new metdata table. Everything else appears to be running correctly for me, locally.

python_src/src/lamp_py/migrations/versions/metadata_prod/001_07903947aabe_initial_changes.py
```diff
@@ -0,0 +1,99 @@
...
+    )
+
+    # pull metadata from the rail performance manager database into the
+    # metadata database. the table may or may not exist, so wrap this in a try
+    # except
+    try:
+        rpm_db_manager = DatabaseManager(
+            db_index=DatabaseIndex.RAIL_PERFORMANCE_MANAGER
+        )
+        md_db_manager = DatabaseManager(db_index=DatabaseIndex.METADATA)
```
rymarczy, line 55: non-blocking:
```suggestion
```

python_src/src/lamp_py/migrations/versions/metadata_staging/001_07903947aabe_initial_changes.py
```diff
@@ -0,0 +1,99 @@
...
+    )
+
+    # pull metadata from the rail performance manager database into the
+    # metadata database. the table may or may not exist, so wrap this in a try
+    # except
+    try:
+        rpm_db_manager = DatabaseManager(
+            db_index=DatabaseIndex.RAIL_PERFORMANCE_MANAGER
+        )
+        md_db_manager = DatabaseManager(db_index=DatabaseIndex.METADATA)
```
rymarczy, line 55: non-blocking:
```suggestion
```

rymarczy (changes requested):
One more change..

python_src/src/lamp_py/migrations/versions/metadata_staging/001_07903947aabe_initial_changes.py
```diff
@@ -0,0 +1,99 @@
...
+        # Raise all other sql errors
+        insert_data = []
+        if error.orig.pgcode == "42P01":
+            logging.info("No Metadata Table in Rail Performance Manager")
+        else:
+            raise
+
+    # insert data into the metadata database
+    if insert_data:
+        op.bulk_insert(MetadataLog.__table, insert_data)
```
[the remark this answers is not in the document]
mzappitello: i'm mad that `mypy` didn't catch this one.

maybe thats why we need sqlalchemy 2.0

mzappitello:
replaced by https://github.com/mbta/lamp/pull/224

PR #202, opened 2023-12-12
FIX: Set SESSION TIMEZONE to fix date to epoch conversion

rymarczy:
To calculate seconds after midnight, from epoch timestamp fields, the Postgres Session timezone must be set to `America/New_York` to correctly convert service_date values into 0 hour timestamp objects. 

This change updates the LAMP_ALL_RT_fields SQL query to set the session timezone and then reset the session timezone after the query has completed.


mzappitello:
still confusing, but LGTM. squash it all together to clean history before merging.

PR #204, opened 2023-12-19
FIX: Update service_id_by_date_and_route VIEW

rymarczy:
Ops Analytics has requested a change to the `service_id_by_date_and_route` export VIEW. 

This change extends the generation of service id's (by date and route) to the `feed_end_date` of the latest available static schedule. 

Updated calculation of canonical_stop_sequence to use row_number function along with associated migration of existing data. 

Downgrade action has been excluded from this migration because the VIEW should never be reverted.

### DIFF
--- a/python_src/src/lamp_py/migrations/versions/performance_manager_prod/003_ae6c6e4b2df5_extend_service_id_view.py
+++ b/python_src/src/lamp_py/migrations/versions/performance_manager_prod/003_ae6c6e4b2df5_extend_service_id_view.py
@@ -28,6 +28,13 @@ def upgrade() -> None:
     op.execute("DROP VIEW IF EXISTS service_id_by_date_and_route;")
     op.execute(view_service_id_by_date_and_route)
 
+    op.create_index(
+        "ix_static_trips_composite_4",
+        "static_trips",
+        ["static_version_key", "service_id"],
+        unique=False,
+    )
+
     update_stop_sequences = (
         "UPDATE vehicle_events "
         "SET canonical_stop_sequence = static_canon.stop_sequence "
@@ -67,4 +74,4 @@ def upgrade() -> None:
 
 
 def downgrade() -> None:
-    pass
+    op.drop_index("ix_static_trips_composite_4", table_name="static_trips")
--- a/python_src/src/lamp_py/migrations/versions/performance_manager_prod/003_ae6c6e4b2df5_extend_service_id_view.py
+++ b/python_src/src/lamp_py/migrations/versions/performance_manager_prod/003_ae6c6e4b2df5_extend_service_id_view.py
@@ -1,0 +1,32 @@
+"""extend service_id_by_date_and_route
+
+Revision ID: ae6c6e4b2df5
+Revises: 1b53fd278b10
+Create Date: 2023-12-17 06:56:17.330783
+
+Details
+* upgrade -> extend service_id_by_date_and_route VIEW to generate values past current date
+
+* downgrade -> Nothing
+"""
+from alembic import op
+
+from lamp_py.migrations.versions.performance_manager_prod.sql_strings.strings_003 import (
+    view_service_id_by_date_and_route,
+)
+
+
+# revision identifiers, used by Alembic.
+revision = "ae6c6e4b2df5"
+down_revision = "1b53fd278b10"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.execute("DROP VIEW IF EXISTS service_id_by_date_and_route;")
+    op.execute(view_service_id_by_date_and_route)
+
+
+def downgrade() -> None:
+    pass
--- a/python_src/src/lamp_py/migrations/versions/performance_manager_prod/003_ae6c6e4b2df5_extend_service_id_view.py
+++ b/python_src/src/lamp_py/migrations/versions/performance_manager_prod/003_ae6c6e4b2df5_extend_service_id_view.py
@@ -6,6 +6,7 @@ Create Date: 2023-12-17 06:56:17.330783
 
 Details
 * upgrade -> extend service_id_by_date_and_route VIEW to generate values past current date
+* upgrade -> update canonical_stop_sequence to use row_number function instead of direct from static schedule
 
 * downgrade -> Nothing
 """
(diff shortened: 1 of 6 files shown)


python_src/src/lamp_py/migrations/versions/performance_manager_prod/sql_strings/strings_003.py
```diff
@@ -0,0 +1,89 @@
...
+        FROM static_calendar_dates
+        WHERE static_calendar_dates.exception_type = 2) AS service_ids_exclude ON all_service_ids.service_id = service_ids_exclude.service_id
+        AND service_ids_exclude.service_date = all_service_ids.service_date
+        AND all_service_ids.static_version_key = service_ids_exclude.static_version_key
+        JOIN
+        (SELECT replace(mod_sd::date::text, '-', '')::integer AS service_date,
+                static_version_key
+        FROM
+            (SELECT feed_start_date::text::date AS start_date,
+                    lead(feed_start_date, 1, feed_end_date) OVER (ORDER BY feed_start_date)::text::date - 1 AS end_date,
```
rymarczy, line 76: This is the only line that changed in the view, the default value for the `lead` function is now `feed_end_date`, instead of the current date.

python_src/src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -511,6 +511,13 @@ def update_stop_sequence(db_manager: DatabaseManager) -> None:
     """
     Update canonical_stop_sequence  and sync_stop_sequence from static_route_patterns
     """
+    partition_by = (
+        StaticRoutePatterns.static_version_key,
+        StaticRoutePatterns.direction_id,
+        sa.func.coalesce(
+            StaticTrips.branch_route_id, StaticTrips.trunk_route_id
+        ).label("route_id"),
+    )
```
mzappitello, line 520: why doesn't this just get added to the `static_canon` select query?

python_src/src/lamp_py/migrations/versions/performance_manager_prod/sql_strings/strings_003.py
mzappitello: i don't quite follow what this change does?

mzappitello (commented):
do we have a method of canceling on staging if the migration fails? id hate to be in a situation where we just have to let it run and hope it finishes.

python_src/src/lamp_py/migrations/versions/performance_manager_prod/sql_strings/strings_003.py
```diff
@@ -0,0 +1,111 @@
...
+                generate_series(mod_fed_dates.start_date, mod_fed_dates.end_date, '1 day') AS mod_sd
+        ) AS sd_match 
+        ON 
+                all_service_ids.static_version_key = sd_match.static_version_key
+                AND all_service_ids.service_date = sd_match.service_date
+        JOIN 
+                static_trips 
+        ON 
+                all_service_ids.service_id = static_trips.service_id
+                AND all_service_ids.static_version_key = static_trips.static_version_key
```
rymarczy, line 101: This JOIN condition needed to be added to fix the issues Heather found.

python_src/src/lamp_py/performance_manager/l1_rt_trips.py
rymarczy: It was just an effort to not make the statement a nested mess

python_src/src/lamp_py/migrations/versions/performance_manager_prod/003_ae6c6e4b2df5_extend_service_id_view.py
```diff
@@ -0,0 +1,77 @@
...
+def upgrade() -> None:
+    op.execute("DROP VIEW IF EXISTS service_id_by_date_and_route;")
+    op.execute(view_service_id_by_date_and_route)
+
+    op.create_index(
+        "ix_static_trips_composite_4",
+        "static_trips",
+        ["static_version_key", "service_id"],
+        unique=False,
+    )
```
rymarczy, line 36: Updating the service_id_by_date_and_route VIEW with the extra join condition to static_trips resulted in a very extended run time to query, total event loop time went up to ~500 seconds.  

In testing, on staging, adding this index reduces query time from ~320 seconds down to ~3 secs.

rymarczy:
> do we have a method of canceling on staging if the migration fails? id hate to be in a situation where we just have to let it run and hope it finishes.

I'm not really sure how that would be possible. Alembic runs the migration, it would need to have some timeout feature that would cancel the migration.

python_src/src/lamp_py/migrations/versions/performance_manager_prod/sql_strings/strings_003.py
rymarczy: This subquery is responsible for creating a list of  all possible service_date's and their associated static_version_key, so essentially which static schedule applies to which service_date. Previously, that list creation stopped at the current date (when the query was executed) but the change makes to go until the feed_end_date of the last static schedule we have. 

If you have a lamp local_rds instance up and running (with some static data in it), the difference can be seen by comparing the results of these two queries: 
```
SELECT 
        TO_CHAR(mod_sd, 'YYYYMMDD')::integer AS service_date,
        static_version_key
FROM
(
        SELECT 
                feed_start_date::text::date AS start_date,
                lead(feed_start_date, 1, feed_end_date) OVER (ORDER BY feed_start_date)::text::date - 1 AS end_date,
                static_version_key
        FROM public.static_feed_info
) AS mod_fed_dates,
generate_series(mod_fed_dates.start_date, mod_fed_dates.end_date, '1 day') AS mod_sd
;
```

```
SELECT 
        TO_CHAR(mod_sd, 'YYYYMMDD')::integer AS service_date,
        static_version_key
FROM
(
        SELECT 
                feed_start_date::text::date AS start_date,
                lead(feed_start_date, 1, TO_CHAR(CURRENT_DATE, 'YYYYMMDD')::integer) OVER (ORDER BY feed_start_date)::text::date - 1 AS end_date,
                static_version_key
        FROM public.static_feed_info
) AS mod_fed_dates,
generate_series(mod_fed_dates.start_date, mod_fed_dates.end_date, '1 day') AS mod_sd
;
```

python_src/src/lamp_py/performance_manager/l1_rt_trips.py
mzappitello: i think its fine. i was just curious.

python_src/src/lamp_py/migrations/versions/performance_manager_prod/003_ae6c6e4b2df5_extend_service_id_view.py
mzappitello: indexes for the win.

python_src/src/lamp_py/migrations/versions/performance_manager_prod/003_ae6c6e4b2df5_extend_service_id_view.py
```diff
@@ -0,0 +1,77 @@
...
+
+    update_stop_sequences = (
```
mzappitello, line 38: not much point in being able to downgrade this, right?

mzappitello (approved):
🍰

PR #205, opened 2023-12-21
FIX: Make canonical_stop_sequence use route_pattern_typicality=5

rymarczy:
Ops Analytics identified an error in the canonical_stop_sequence calculation we were using. This error is triggered by 3 static schedules released in Dec 2023:
- Fall 2023, 2023-12-07T22:35:17+00:00, version D
- Fall 2023, 2023-12-05T18:45:28+00:00, version D
- Fall 2023, 2023-12-04T21:09:26+00:00, version D

When picking which representative_trip_id to use from route_patterns, records with route_pattern_typicality=5 should be used first for any route-pattern/direction combination and then route_pattern_typicality=1 should be used when the first option is not available. 

This change will work with static schedule going back to January 2019, but will fail for any older GTFS schedule versions.

This change updates our daily canonical_stop_sequence calculation logic and also deploys a migration for existing canonical_stop_sequence values in the staging RDS.

This change also shrinks our prod migrations into a single file, since no deploy has been pushed to the PROD RDS yet, a single migration file is sufficient.


docker-compose.yml
```diff
@@ -49,6 +48,7 @@ services:
         "run",
         "seed_metadata",
         "--clear-static",
+        "--clear-rt",
         "--seed-file",
-        "/april_2023_filepaths.json"
+        "/seed_paths.json"
```
mzappitello, line 53: good choice.

python_src/src/lamp_py/migrations/versions/performance_manager_prod/001_5d9a7ee21ae5_initial_prod_schema.py
mzappitello, line 1: we have deployed to the prod RDS. we needed to do that in order to setup the metadata table that ingestion is populating.

python_src/src/lamp_py/migrations/versions/performance_manager_staging/004_45dedc21086e_canon_stop_seq.py
```diff
@@ -0,0 +1,105 @@
...
+        FROM
+        (
+            SELECT
+                DISTINCT ON (
+                    coalesce(static_trips.branch_route_id, static_trips.trunk_route_id),
+                    static_route_patterns.direction_id,
+                    static_route_patterns.static_version_key
+                ) 
+                static_route_patterns.direction_id AS direction_id,
+                static_route_patterns.representative_trip_id AS representative_trip_id,
```
mzappitello, line 61: how will this value be chosen if there are collisions as its not in the `DISTINCT ON` clause?

python_src/src/lamp_py/migrations/versions/performance_manager_prod/001_5d9a7ee21ae5_initial_prod_schema.py
rymarczy: Are we going to tear it down and re-deploy after the Metadata RDS is deployed? It looks like it's currently configured with a read-replica, which I don't think we want anymore?

python_src/src/lamp_py/migrations/versions/performance_manager_staging/004_45dedc21086e_canon_stop_seq.py
mzappitello: i'm assuming it will be from the latest schedule based on the `ORDER BY` on line 74?

python_src/src/lamp_py/migrations/versions/performance_manager_staging/004_45dedc21086e_canon_stop_seq.py
```diff
@@ -0,0 +1,105 @@
...
+                static_route_patterns.representative_trip_id = static_trips.trip_id
+                AND static_route_patterns.static_version_key = static_trips.static_version_key
+            WHERE
+                static_route_patterns.route_pattern_typicality = 1
+                OR static_route_patterns.route_pattern_typicality = 5
+            ORDER BY
+                coalesce(static_trips.branch_route_id, static_trips.trunk_route_id),
+                static_route_patterns.direction_id,
+                static_route_patterns.static_version_key,
+                static_route_patterns.route_pattern_typicality DESC
```
mzappitello, line 78: would be interested to hear how you would solve this if the order was supposed to be 3, 5, 1 rather than 5, 1.

rymarczy: It would require additional subqueries where `route_pattern_typicality=3` and then some sort of exclusion rules on a JOIN to subquerie(s) with the other `route_pattern_typicality` values.

python_src/src/lamp_py/migrations/versions/performance_manager_staging/004_45dedc21086e_canon_stop_seq.py
```diff
@@ -0,0 +1,105 @@
...
+            canon_trips.representative_trip_id = static_stop_times.trip_id
+            AND canon_trips.static_version_key = static_stop_times.static_version_key
+        JOIN 
+            static_stops 
+        ON 
+            static_stop_times.stop_id = static_stops.stop_id
+            AND static_stop_times.static_version_key = static_stops.static_version_key
+    ) AS static_canon 
+    ON 
+        ve.parent_station = static_canon.parent_station
```
mzappitello, line 92: why is it that we're matching on the less specific parent station?

python_src/src/lamp_py/migrations/versions/performance_manager_staging/004_45dedc21086e_canon_stop_seq.py
rymarczy: Yes, the `DISTINCT ON` clause works with the `ORDER BY` to select the first result on any collisions. SO with `static_route_patterns.route_pattern_typicality DESC` in the `ORDER BY` `route_pattern_typicality=5` is selected before any lower `route_pattern_typicality` value.

python_src/src/lamp_py/migrations/versions/performance_manager_staging/004_45dedc21086e_canon_stop_seq.py
rymarczy: We do everything by `parent_station` so that if the GTFS-RT data reports multiple `stop_id` values from the same `parent_station` we don't end up recording them as separate events, which they should not be.

python_src/src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -543,12 +541,54 @@ def update_stop_sequence(db_manager: DatabaseManager) -> None:
...
+                StaticRoutePatterns.route_pattern_typicality == 5,
+            ),
+        )
+        .order_by(
+            sa.func.coalesce(
+                StaticTrips.branch_route_id, StaticTrips.trunk_route_id
+            ),
+            StaticRoutePatterns.direction_id,
+            StaticRoutePatterns.static_version_key,
+            StaticRoutePatterns.route_pattern_typicality.desc(),
```
mzappitello, line 564: i think a comment here that notes we want route pattern typicality of 5 first, falling back to 1 where we can't find it.

i think this distinction gets lost pretty easily.

python_src/src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -543,12 +541,54 @@ def update_stop_sequence(db_manager: DatabaseManager) -> None:
...
+        )
+        .subquery("canon_trips")
+    )
+    partition_by = (
```
mzappitello, line 568: i think this could be renamed `stop_sequence_partition` to be a bit more clear?

could also just add it to the `static_cannon` subquery now that its not as large.

python_src/src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -655,7 +684,7 @@ def update_stop_sequence(db_manager: DatabaseManager) -> None:
         .subquery("zero_points")
     )
 
-    # select stop_sequence number for the zerp_point parent_station of each route-branch,
+    # select stop_sequence number for the zero_point parent_station of each route-branch,
```
mzappitello, line 687: revert this please.

mzappitello (commented):
a few different questions here.

python_src/src/lamp_py/migrations/versions/performance_manager_staging/004_45dedc21086e_canon_stop_seq.py
mzappitello: yuck. but cool.

python_src/src/lamp_py/migrations/versions/performance_manager_prod/001_5d9a7ee21ae5_initial_prod_schema.py
mzappitello: we'll discuss this on the call.

mzappitello (approved):
🍰

PR #209, opened 2024-01-03
FIX: Always Create new service_id_by_date_and_route parquet

rymarczy:
The service_id_by_date_and_route VIEW dataset cannot be updated in the same manner as the other GTFS Static datasets being sent to tableau.

This change allows the service_id_by_date_and_route VIEW dataset to be re-created instead of updated whenever new static schedule data is available. 

Asana Task: https://app.asana.com/0/1205827492903547/1206273298673484


mzappitello (approved):
🍰

PR #214, opened 2024-01-03
FIX: Ingestion RT table save for yield

rymarczy:
Ingestion is running out of memory processing very large trip update tables. 

This fix changes the way RT table data is stored during ingestion loops.

Previously, `json.gz` table data was stored as a list of individual pyarrow tables, that was concatenated before writing to parquet. 

The new method concats each `json.gz` pyarrow table to a main pyarrow table when each table is extracted.


mzappitello (approved):
🍰

PR #216, opened 2024-01-04
Use ParquetWriter rather than Dataset.writedataset

mzappitello:
The pyarrow write dataset function has shown to consume a lot of memory while also taking longer to execute. It has features we use around partitioning and file visitor functions, but those features can be replicated without too much difficulty given the way we use them.

Asana Task: https://app.asana.com/0/1205827492903547/1206273298673481/f

### DIFF
--- a/python_src/src/lamp_py/aws/s3.py
+++ b/python_src/src/lamp_py/aws/s3.py
@@ -20,7 +20,7 @@ from typing import (
 import boto3
 import pandas
 import pyarrow
-import pyarrow.dataset as ds
+import pyarrow.compute as pc
 import pyarrow.parquet as pq
 from pyarrow import Table, fs
 from pyarrow.util import guid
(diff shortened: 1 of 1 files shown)


python_src/src/lamp_py/aws/s3.py
```diff
@@ -349,26 +350,34 @@ def write_parquet_file(
...
-        data=table,
-        base_dir=s3_dir,
-        filesystem=fs.S3FileSystem(),
-        format=ds.ParquetFileFormat(),
-        partitioning=partitioning,
-        file_visitor=visitor_func,
-        basename_template=basename_template,
-        existing_data_behavior="overwrite_or_ignore",
+    write_path = os.path.join(
+        s3_dir, *partition_strings, basename_template.format(i=0)
```
rymarczy, line 366: question/nitpick:

I think `write_dataset` required us to have the `{i}` notation in the `basename_template` but that isn't really required anymore and we're not using the `basename_template` parameter for it's true intended purpose, so I'm wondering if it might be worth dropping/renaming that parameter along with the `{i}` notation.

rymarczy (approved):
One nitpick/question, but otherwise LGTM.

python_src/src/lamp_py/aws/s3.py
mzappitello: i had that thought as well. i left it like this just to keep things consistent with the prior behavior.

PR #217, opened 2024-01-05
Fix rds writer process and with more typing

mzappitello:
The RDS writer process waits for items to be put on a metadata queue, pops them off, and then writes that file to the metadata. In the past, the queue was typed with any, as it was being populated with an obscure pyarrow object we couldn't deduce properly.

With the recent changes to how we write parquet files, these changed to strings. This caused the process to fail since it was attempting to pull a `path` attribute out a string, causing an uncaught exception.

I've updated the process to expect a string from the queue and put in typing hints that should prevent an error like this in the future.

Asana Task: https://app.asana.com/0/1205827492903547/1206288301107382/f

### DIFF
--- a/python_src/src/lamp_py/aws/s3.py
+++ b/python_src/src/lamp_py/aws/s3.py
@@ -312,7 +312,7 @@ def write_parquet_file(
     file_type: str,
     s3_dir: str,
     partition_cols: List[str],
-    visitor_func: Optional[Callable[..., None]] = None,
+    visitor_func: Optional[Callable[[str], None]] = None,
     basename_template: Optional[str] = None,
 ) -> None:
     """
--- a/python_src/src/lamp_py/postgres/postgres_utils.py
+++ b/python_src/src/lamp_py/postgres/postgres_utils.py
@@ -2,7 +2,8 @@ import os
 import platform
 import time
 import urllib.parse as urlparse
-from multiprocessing import Manager, Process, Queue
+from queue import Queue
+from multiprocessing import Manager, Process
 from typing import Any, Dict, List, Optional, Tuple, Union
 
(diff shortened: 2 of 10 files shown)


python_src/src/lamp_py/postgres/postgres_utils.py
```diff
@@ -462,15 +462,15 @@ def _rds_writer_process(metadata_queue: Queue) -> None:
...
 
 
-def start_rds_writer_process() -> Tuple[Queue, Process]:
+def start_rds_writer_process() -> Tuple[Queue[Optional[str]], Process]:
     """
     create metadata queue and rds writer process
 
     return metadata queue
     """
     # mypy: Function "multiprocessing.Manager" is not valid as a type
```
rymarczy, line 471: This can be dropped.

python_src/src/lamp_py/aws/s3.py
```diff
@@ -376,7 +376,7 @@ def write_parquet_file(
 
     # call the visitor function if it exists
     if visitor_func is not None:
-        visitor_func(write_path)
+        visitor_func(123)
```
rymarczy, line 379:
```suggestion
        visitor_func(write_path)
```

mzappitello: yeah. i accidentally commited that after running black

rymarczy (approved):
LGTM 

I turned off ingestion on staging and updated the metadatalog RDS table with all of the parquet paths that were written since the last manual update I did. 

Should be good to deploy and kick everything back on. On PROD, I can reconcile the parquet file and metadatalog table diffs later today and write up an SQL query for the infra team to run against the PROD rds.

PR #218, opened 2024-01-08
Don't Install Tableau API on arm64

mzappitello:
The tableauhyperapi module is unavailable on mac with arm64 processors.

* Update poetry to take advantage of dependency groups.
* Add a tableau dependency group that can be excluded when installing dependencies.
* Update python_src to fallback with error messages when using the tableau subdir that creates hyperfiles fails to import everything.
* Use new `BUILDPLATFORM` arg in dockerfile to build without tableau deps on an arm64 platform.
* Add an export for the BUILDPLATFORM variable to .envrc and pass it through via docker-compose when building the lamp py images.


.envrc
```diff
@@ -1,3 +1,7 @@
 use asdf
 
-dotenv
\ No newline at end of file
+dotenv
+
+# used in dockercompose and dockerfile 
+export BUILDPLATFORM=$(uname -m)
```
mzappitello, line 6: this is passed to docker compose, which passes it to the dockerfile, which uses it to install tableau deps or not.

docker-compose.yml
```diff
@@ -19,6 +19,8 @@ services:
     env_file: .env
     build:
       context: ./python_src
+      args:
+        BUILDPLATFORM: $BUILDPLATFORM
```
mzappitello, line 23: pass arg to dockerfile which is used to install poetry or not.

python_src/Dockerfile
```diff
@@ -16,13 +16,21 @@ RUN chmod a=r /usr/local/share/amazon-certs.pem
...
-RUN poetry install --no-dev --no-interaction --no-ansi -v
+
+# Tableau dependencies for arm64 cannot be resolved (since salesforce doesn't
+# support them yet). For that buildplatform build without those dependencies
+ARG BUILDPLATFORM
+RUN echo "Installing python dependencies for ${BUILDPLATFORM}."
+RUN if [ "$BUILDPLATFORM" = "arm64" ]; then \
+    poetry install --without tableau --no-interaction --no-ansi -v ;\
+    else poetry install --no-interaction --no-ansi -v ;\
+    fi
```
mzappitello, line 33: if its an arm64 build platform, install without the tableau dep group.

python_src/pyproject.toml
```diff
@@ -27,10 +27,15 @@ psutil = "^5.9.1"
 schedule = "^1.1.0"
 alembic = "^1.10.2"
 types-pytz = "^2023.3.0.1"
+
+[tool.poetry.group.tableau]
```
mzappitello, line 31: add dependency group for tableau dependencies. its not optional, meaning a user will have to change behavior to build without it.

python_src/src/lamp_py/tableau/__init__.py
```diff
@@ -1 +1,21 @@
-"""Utilites for Interacting with Tableau and Hyper files"""
+"""Utilities for Interacting with Tableau and Hyper files"""
```
mzappitello, line 1: update the tableau init file to catch exceptions when the tableau modues arent found.

python_src/src/lamp_py/performance_manager/pipeline.py
```diff
@@ -15,7 +15,7 @@
 from lamp_py.runtime_utils.env_validation import validate_environment
 from lamp_py.runtime_utils.process_logger import ProcessLogger
 
-from lamp_py.tableau.pipeline import start_parquet_updates
+from lamp_py.tableau import start_parquet_updates
```
mzappitello, line 18: this is the only part i don't like. if we import directly from the file we're doing to get the import error and it felt like catch that error should be left to the subdir.

rymarczy (changes requested):
This all looks like it's functioning as expected for me locally.

I think we need to add some documentation regarding why/how we are doing this in a README file. We are also missing a README for the "tableau" applications, which is my oversight when originally adding that functionality. 

@paulswartz Could you also please review this approach and see if you can find any potential issues with this approach? Are any other TID apps doing anything similar with a different approach?

python_src/src/lamp_py/tableau/__init__.py
```diff
@@ -1 +1,29 @@
...
+
```
paulswartz, line 2: suggestion(non-blocking):
```python
try:
  import .pipeline as pipeline
except ModuleNotFoundError:
  pipeline = None

def start_parquet_updates(db_manager: DatabaseManager) -> None
  if pipeline is None:
    logging.exception(...)
  else:
    pipeline.start_parquet_updates(db_manager)
```

.envrc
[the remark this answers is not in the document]
mzappitello: i tried this out locally based on this same documentation and found that the value evaluated to an empty string when I ran on my machine.

python_src/src/lamp_py/tableau/__init__.py
mzappitello: oh. i like this much better.

.envrc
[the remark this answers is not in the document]
paulswartz: This diff worked for me without needing to set any additional variables:
```diff
diff --git a/python_src/Dockerfile b/python_src/Dockerfile
index 06a3e1e..40d69a6 100644
--- a/python_src/Dockerfile
+++ b/python_src/Dockerfile
@@ -25,9 +25,9 @@ COPY pyproject.toml pyproject.toml

 # Tableau dependencies for arm64 cannot be resolved (since salesforce doesn't
 # support them yet). For that buildplatform build without those dependencies
-ARG BUILDPLATFORM
-RUN echo "Installing python dependencies for ${BUILDPLATFORM}."
-RUN if [ "$BUILDPLATFORM" = "arm64" ]; then \
+ARG TARGETARCH BUILDPLATFORM TARGETPLATFORM
+RUN echo "Installing python dependencies for build: ${BUILDPLATFORM} target: ${TARGETPLATFORM}"
+RUN if [ "$TARGETARCH" = "arm64" ]; then \
     poetry install --without tableau --no-interaction --no-ansi -v ;\
     else poetry install --no-interaction --no-ansi -v ;\
     fi
```

rymarczy (approved):
LGTM, working locally for me.

PR #219, opened 2024-01-11
Add boilerplate for Bus Performance Manager

mzappitello:
* Create bus_performance_manager with readme, init, and pipeline files in python src.
* Create basic pipeline script that starts up, sets and validates the environment, and runs an event loop that checks for sigterm.
* edit the pyproject toml file to execute the bus performance manager in the poetry environment.
* Add a service to docker compose to run the bus performance manager.
* Update github action deploy scripts.


rymarczy (approved):
LGTM

PR #223, opened 2024-01-16
Add pause utility and use after static load failure

mzappitello:
Add a new file and function `infinite_wait` in runtime utilities to pause a process without killing it, waiting for user interaction. In the static load function, pause the process if an error occurs.

Asana Task: (https://app.asana.com/0/1205827492903547/1206050538281288/f)

### DIFF
--- a/python_src/src/lamp_py/runtime_utils/infinite_wait.py
+++ b/python_src/src/lamp_py/runtime_utils/infinite_wait.py
@@ -1,0 +1,28 @@
+import logging
+import time
+
+from lamp_py.aws.ecs import check_for_sigterm
+
+
+def infinite_wait(reason: str) -> None:
+    """
+    When running on ECS, propagating an exception up the call stack and killing
+    the processes will result in the process being restarted, to keep the task
+    count at one. This method should be called instead when we want to pause
+    the process for intervention before restarting.
+    """
+    # amount of time to sleep between logging statements
+    sleep_time = 60
+    count = 0
+
+    while True:
+        check_for_sigterm()
+
+        # log every ten minutes
+        if count == 10:
+            logging.error("Pausing for %s", reason)
+            count = 0
+
+        # sleep
+        time.sleep(sleep_time)
+        count += 1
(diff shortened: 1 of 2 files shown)


python_src/src/lamp_py/runtime_utils/infinite_wait.py
```diff
@@ -0,0 +1,26 @@
...
+
+    while True:
+        # log the error
+        logging.error("Pausing for %s", reason)
+
+        # sleep
+        time.sleep(nap_time)
+
+        # adjust the sleep period
+        nap_time = max(2 * nap_time, max_sleep)
```
rymarczy, line 26: should this be min?
```suggestion
        # adjust the sleep period
        nap_time = min(2 * nap_time, max_sleep)
```

python_src/src/lamp_py/runtime_utils/infinite_wait.py
```diff
@@ -0,0 +1,26 @@
...
+    When running on ECS, propagating an exception up the call stack and killing
+    the processes will result in the process being restarted, to keep the task
+    count at one. This method should be called instead when we want to pause
+    the process for intervention before restarting.
+    """
+    # max amount of time to sleep between logging statements
+    max_sleep = 30 * 60
+
+    # initial sleep period, to be doubled between each log statement
+    nap_time = 60
```
rymarczy, line 16: I think a constant sleep duration should be sufficient, especially if we're waiting for a `SIGTERM` to reset the ECS.

python_src/src/lamp_py/runtime_utils/infinite_wait.py
```diff
@@ -0,0 +1,26 @@
...
+
+    while True:
+        # log the error
+        logging.error("Pausing for %s", reason)
```
rymarczy, line 20: We should probably be running `check_for_sigterm` in this loop, to handle a graceful shutdown, right?

rymarczy (changes requested):
Few comments/questions.

rymarczy (approved):
LGTM

PR #227, opened 2024-01-23
FIX: Incorrect WHERE clause on static_version_key sync

rymarczy:
Found this incorrect WHERE clause while debugging a separate issue raised by ops analytics.


mzappitello (approved):
🍰

PR #230, opened 2024-02-02
FEAT: Add parallel task checking to Tableau Publisher

rymarczy:
This change adds a check to the Tableau Publisher pipeline to make sure that only one is running at a time. 

Implementation was lifted from dmap-import. 

Requires ENV VARS to be added to ECS first with devops PR: https://github.com/mbta/devops/pull/1540

Asana Task: https://app.asana.com/0/1205827492903547/1206184120740467

### DIFF
--- a/python_src/src/lamp_py/aws/ecs.py
+++ b/python_src/src/lamp_py/aws/ecs.py
@@ -66,24 +66,29 @@ def check_for_parallel_tasks() -> None:
     ecs_cluster = os.environ["ECS_CLUSTER"]
     ecs_task_group = os.environ["ECS_TASK_GROUP"]
 
-    # get all of the tasks running on the cluster
-    task_arns = client.list_tasks(cluster=ecs_cluster)["taskArns"]
+    try:
+        # get all of the tasks running on the cluster
+        task_arns = client.list_tasks(cluster=ecs_cluster)["taskArns"]
 
-    # if tasks are running on the cluster, get their descriptions and check to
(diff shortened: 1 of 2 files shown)


python_src/src/lamp_py/aws/ecs.py
```diff
@@ -41,3 +43,48 @@ def check_for_sigterm(
...
+            if ecs_task_group == task["group"]:
+                match_count += 1
+
+    # if the group matches, raise an exception that will terminate the process
+    if match_count > 1:
+        exception = SystemError(f"Multiple {ecs_cluster} ECS Tasks Running")
+        process_logger.log_failure(exception)
+        raise exception
+
+    process_logger.log_complete()
```
mzappitello, line 90: might be worth putting this in a try except to check for key errors, log, and re-raise.

mzappitello (commented):
one comment.

other than that LGTM

python_src/src/lamp_py/aws/ecs.py
rymarczy: Do you mean key errors for the `tasks` and `groups` indexing from `client.describe_tasks`?

mzappitello: yeah. that would be the thing to catch i think? that way we get a failure log on this process rather than a failure on the main method.

not a huge deal, but feels a little cleaner and like a practice we should keep up imho.

PR #250, opened 2024-02-14
chore: Manually update Pandas to version 2.2.0

rymarczy:
The updated of pandas to version 2.2.0 exposed issues with our `test_whole_table` performance manager test. 

The `Dataframe.compare` function that we were using to check equivalency between our pipeline output and the source of truth CSV file was not performing as expected. Use of this function has been replaced with `pandas.testing.assert_series_equal` which produces a much more strict comparison between dataframe columns. 

Also added an additional test for the `process_vp_files` function, which was used during the debugging process, so might as well add it. 

Asana Task: https://app.asana.com/0/1205827492903547/1204354793482788


python_src/src/lamp_py/performance_manager/l0_rt_vehicle_positions.py
```diff
@@ -158,7 +158,7 @@ def transform_vp_timestamps(
         vehicle_positions,
         index=trip_stop_columns,
         columns="is_moving",
-        aggfunc={"vehicle_timestamp": min},
+        aggfunc={"vehicle_timestamp": "min"},
```
mzappitello, line 161: thats kind of a strange change. was it using pythons built in min function before and is now using some internal one?

mzappitello:
LGTM. There is the other pandas PR in here that dependabot built that we should close as well.

python_src/src/lamp_py/performance_manager/l0_rt_vehicle_positions.py
rymarczy: Just having `min` still works, but throws a warning.  All of the pandas documentation for the `pivot_table` function references all agg functions as just strings.

PR #254, opened 2024-02-16
fix: Last Stop of trip Headways Value

rymarczy:
Ops Analytics identified an issue where the performance_manager application is producing headways values for the last stop of trips. Headways are calculated on a departure-to-departure basis, and the last stop of a trip should have not have a departure.

An update to the `trips_for_headways_subquery` function calculates `next_station_move` (aka station departure timestamp) using a `lead` window function partitioned by a unique trip id. This will result in a null `next_station_move` value for the last stop of a trip. The `order_by` parameter of the window function has also been updated to coalesce the `vp_move_timestamp` and `vp_stop_timestamp` fields to correctly order station events that don't have an initial `vp_move_timestamp`.  This often occurs for vehicle events that are the first stop of a trip. 

This change to headways calculations resulted in a few new headways values being added to the `pipeline_flat_out.csv` file, this is because of the change to the `order_by` parameter, creating new headways values for the first stop events on several trips. No other headways values required updates. 

Asana Task: https://app.asana.com/0/1205827492903547/1206549592761356


python_src/tests/performance_manager/test_performance_manager.py
```diff
@@ -65,6 +65,7 @@
 from ..test_resources import springboard_dir, test_files_dir, csv_to_vp_parquet
 
 
+@pytest.mark.skip("utility function, not a test")
```
rymarczy, line 68: `pytest` always threw a warning when running this function. this seems to clean that up.

mzappitello (approved):
cake

PR #267, opened 2024-02-28
FIX: Reduce resource usage during parquet file creation

rymarczy:
The LAMP staging environment has been displaying excessively high memory usage and crashes during the creation of parquet files for Tableau publishing. The issue was traced back to the `write_to_parquet` function of our DatabaseManager object. 

SQLAlchemy result streaming through use of the `yield_per` method, was not behaving as intended. Instead of incrementally streaming results from a query, it was materializing the entire result stream in memory before iterating through the results. Updating the `write_to_parquet` function to use the `execute_options` parameter has resolved this issue and results are now correctly streamed incrementally and memory usage during streaming is contained to a single `batch_size` amount. 

Additionally, this change adds helper functions to our DatabaseManager objects to ENABLE and DISABLE the `rt_trips_update_branch_trunk` TRIGGER of our `vehicle_trips` table. This trigger was causing excessively long query team for an UPDATE operation against the `vehicle_trips` table and is generally not required. This is a bit of a band-aid fix for now, that can be fixed with architectural RDS and/or application changes in the future. 

Asana Task: https://app.asana.com/0/1205827492903547/1206703823960612

### DIFF
--- a/python_src/src/lamp_py/performance_manager/l1_rt_trips.py
+++ b/python_src/src/lamp_py/performance_manager/l1_rt_trips.py
@@ -183,7 +183,7 @@ def load_new_trip_data(db_manager: DatabaseManager) -> None:
             VehicleTrips.trip_id == distinct_update_query.c.trip_id,
         )
     )
-    db_manager.execute(trip_update_query)
+    db_manager.execute(trip_update_query, disable_trip_tigger=True)
 
     process_logger.log_complete()
 
@@ -217,7 +217,7 @@ def update_static_version_key(db_manager: DatabaseManager) -> None:
 
     process_logger = ProcessLogger("l1_trips.update_static_version_key")
     process_logger.log_start()
-    db_manager.execute(update_query)
+    db_manager.execute(update_query, disable_trip_tigger=True)
     process_logger.log_complete()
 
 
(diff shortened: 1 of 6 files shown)


python_src/src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -183,7 +183,7 @@ def load_new_trip_data(db_manager: DatabaseManager) -> None:
             VehicleTrips.trip_id == distinct_update_query.c.trip_id,
         )
     )
-    db_manager.execute(trip_update_query)
+    db_manager.execute(trip_update_query, disable_trip_tigger=True)
```
mzappitello, line 186: is the idea that we'll run this trigger _once_ per event loop rather than on each thing that touches the `VehicleTrips`?

python_src/src/lamp_py/postgres/postgres_utils.py
```diff
@@ -380,16 +401,28 @@ def write_to_parquet(
...
+
+        part_stmt = select_query.execution_options(
+            stream_results=True,
+            max_row_buffer=batch_size,
+        )
         with self.session.begin() as cursor:
-            result = cursor.execute(select_query).yield_per(batch_size)
             with pq.ParquetWriter(write_path, schema=schema) as pq_writer:
-                for part in result.partitions():
+                for part in cursor.execute(part_stmt).partitions(batch_size):
```
mzappitello, line 417: this is the real meat of the change, right?

any thoughts on why this was working before and stopped now?

mzappitello (approved):
Two questions, but LGTM

python_src/src/lamp_py/postgres/postgres_utils.py
rymarczy: My intuition is that it was never really working as we expected. Hence the large spike in memory usage at each app start, where the LAMP_ALL_RT parquet file had to be fully created. But for some reason, that memory spike was being re-claimed at some point and I'm not sure why that stopped happening.

python_src/src/lamp_py/performance_manager/l1_rt_trips.py
rymarczy: Yes, we expect the trigger to get run as part of the `update_trip_stop_counts` function, which we have commented as such. But specifically disabling for all other quries that are updating `VehcileTrips` should be a major speed up. 

Using my staging DB extract for testing, the `update_static_version_key` call drops from ~100 seconds to 38ms.

rymarczy: It's a little hacky, but seems reasonable for now??

PR #269, opened 2024-03-01
FEAT: Skip upload of LAMP_ALL_RT_fields.parquet file when no new records

rymarczy:
Currently, the `LAMP_ALL_RT_fields.parquet` file is uploaded to S3 on every event loop. 

The change adds a check between the records >= `max_start_date` in the database and in the existing `LAMP_ALL_RT_fields.parquet` file. If the counts are equivalent, the parquet file is not re-uploaded to S3.

Also includes all dependabot updates for the week.

### DIFF
--- a/python_src/pyproject.toml
+++ b/python_src/pyproject.toml
@@ -19,7 +19,7 @@ hyper_update = 'lamp_py.tableau.pipeline:start_hyper_updates'
 python = "^3.10"
 SQLAlchemy = "^2.0.27"
 pyarrow = "^15.0.0"
-boto3 = "^1.34.46"
+boto3 = "^1.34.50"
 pandas = "^2.2.1"
 numpy = "^1.26.4"
 psycopg2 = "^2.9.3"
@@ -38,8 +38,8 @@ tableauserverclient = "0.30"
 [tool.poetry.group.dev.dependencies]
 black = "^24.2.0"
 mypy = "^1.1.1"
-pylint = "^3.0.3"
-pytest = "^8.0.1"
+pylint = "^3.1.0"
+pytest = "^8.0.2"
 ipykernel = "^6.29.2"
 
 [build-system]
--- a/python_src/src/lamp_py/tableau/jobs/rt_rail.py
+++ b/python_src/src/lamp_py/tableau/jobs/rt_rail.py
@@ -185,6 +185,13 @@ class HyperRtRail(HyperJob):
             schema=self.parquet_schema,
         )
 
+        check_filter = pc.field("service_date") >= max_start_date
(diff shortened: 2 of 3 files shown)


python_src/src/lamp_py/tableau/jobs/rt_rail.py
```diff
@@ -99,6 +99,8 @@ def __init__(self) -> None:
             "   ve.service_date, vt.route_id, vt.direction_id, vt.vehicle_id, vt.start_time"
             ";"
         )
+        self.ds_batch_size = 1024 * 1024
+        self.db_parquet_path = "/tmp/db_local.parquet"
```
rymarczy, line 103: moved these because of "too many local variables" from pylint in `update_parquet`

python_src/src/lamp_py/tableau/jobs/rt_rail.py
```diff
@@ -178,17 +178,23 @@ def update_parquet(self, db_manager: DatabaseManager) -> bool:
...
+            write_path=self.db_parquet_path,
             schema=self.parquet_schema,
         )
 
+        check_filter = pc.field("service_date") >= max_start_date
+        if pd.dataset(self.db_parquet_path).count_rows() == pd.dataset(
+            self.local_parquet_path
+        ).count_rows(filter=check_filter):
+            # No new records from database, no upload requried
+            return False
```
rymarczy, line 192: meat of the change

the `pyarrow.dataset.count_rows()` method is very cheap, returns number of records from a parquet file instantly.

python_src/pyproject.toml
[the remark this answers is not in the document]
rymarczy: I threw in the dependabot changes for the week, in attempt to avoid some deployment thrash on staging.

python_src/src/lamp_py/tableau/jobs/rt_rail.py
```diff
@@ -178,17 +178,23 @@ def update_parquet(self, db_manager: DatabaseManager) -> bool:
             f" AND vt.service_date >= {max_start_date.strftime('%Y%m%d')} ",
         )
 
-        db_parquet_path = "/tmp/db_local.parquet"
         db_manager.write_to_parquet(
             select_query=sa.text(update_query),
-            write_path=db_parquet_path,
+            write_path=self.db_parquet_path,
```
mzappitello, line 183: should batch size be included in this function call?

mzappitello (commented):
a few questions.

python_src/src/lamp_py/tableau/jobs/rt_rail.py
rymarczy: We're ok with the default batch size `1024*1024` so I just left it out.

PR #271, opened 2024-03-01
FEAT: Update Tableau Parquet files based on S3 Metdata Version field

rymarczy:
Previously, all Tableau parquet files were being deleted and re-created on every application deployment.

This change utilizes the S3 `Metadata` head field to store a `lamp_version` value to determine when datasets should be re-freshed / re-created. 

Asana Task: https://app.asana.com/0/1189492770004753/1206729585163530

### DIFF
--- a/python_src/src/lamp_py/aws/s3.py
+++ b/python_src/src/lamp_py/aws/s3.py
@@ -15,6 +15,7 @@ from typing import (
     Tuple,
     Union,
     cast,
+    Dict,
 )
 
 import boto3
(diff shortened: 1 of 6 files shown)


mzappitello:
i'm tempted to say that the version of the tableau file should be tied to the version of `lamp` itself rather than a version we increment when we feel we need to.

this would result in move overwrites then we strictly need, but it would make some forms of debugging easier if we could look at a file and know the conditions it was made under. i was thinking of tying them to minor versions of lamp, so when we jump from 1.0.6 to 1.1.0 it would require a bump.

that would require we have a method for determining the version, that we run minor version bumps when this changes for any of our public files, and that we have a way to run version bumps on staging without triggering prod deploys.

PR #272, opened 2024-03-02
FIX: Add ContentType to index.html upload

rymarczy:
Our index.html page for performancedata.mbta.com is not being served by CloudFront because the S3 `ContentType` parameter is set to `binary/octet-stream` causing the file to be offered as a download. 

This change will upload index.html with `text/html` set for `ContentType` which should hopefully serve the file as a normal webpage. 

Asana Task: https://app.asana.com/0/1205827492903547/1204907226653450

### DIFF
--- a/python_src/src/lamp_py/aws/s3.py
+++ b/python_src/src/lamp_py/aws/s3.py
@@ -49,6 +49,7 @@ def upload_file(
         "s3_upload_file",
         file_name=file_name,
         object_path=object_path,
+        **extra_args,
     )
     upload_log.log_start()
 
--- a/python_src/src/lamp_py/aws/s3.py
+++ b/python_src/src/lamp_py/aws/s3.py
@@ -61,7 +61,9 @@ def upload_file(
 
         s3_client = get_s3_client()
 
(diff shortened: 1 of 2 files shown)


python_src/src/lamp_py/aws/s3.py
```diff
@@ -61,7 +61,9 @@ def upload_file(
 
         s3_client = get_s3_client()
 
-        s3_client.upload_file(file_name, bucket, object_name, extra_args)
+        s3_client.upload_file(
+            file_name, bucket, object_name, ExtraArgs=extra_args
+        )
```
rymarczy, line 66: Added `ExtraArgs` and named parameter per: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/s3-uploading-files.html

mzappitello: can we add the extra args to the `upload_log`?

mzappitello (approved):
one change but LGTM aft that goes in.

python_src/src/lamp_py/aws/s3.py
rymarczy: Incorporated!

PR #293, opened 2024-03-26
Add Investigation Group to Dependencies

mzappitello:
Add a new dependency group to pyproject.toml that contains libraries that are useful for data investigations. I've used this flow while working on the bus stop crossing investigation and it allows me to install libraries that aren't needed by LAMP applications locally while still using LAMP utilities.

The can be installed with
`poetry install --with investigation`

I also added a directory to .gitignore where I've been putting notebook and datafiles.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ erl_crash.dump
 .elixir_ls/
 /performance_manager/test.db
 /notebook
+/investigation
 
 
 __pycache__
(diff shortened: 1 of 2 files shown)


rymarczy:
Would there be any issue with just adding these to the `dev` dependencies?

rymarczy (approved):
one question, otherwise LGTM

mzappitello:
i prefer this way as it encapsulates the libraries needed for investigation as separate from dev dependencies which we use to inspect the code itself. this way we can be more selective with what optional dependencies we're installing. for example, the github python ci tests need the dev dependencies, but not the investigation ones.

PR #294, opened 2024-03-26
FIX - Run CI when dependencies are altered

mzappitello:
I noticed that we didn't run CI when I put a PR that modified `pyproject.toml`.

I modified the paths clauses on our workflow:
  * add pyproject.toml and poetry.lock to catch when dependencies are changed
  * remove the `./**` condition that seems to not be working as expected.
  * add in `src/**` and `tests/**` paths to catch code and test changes


.github/workflows/ci_python.yaml
mzappitello, line 1: this is the change that triggers the tests.

tests/performance_manager/test_performance_manager.py
mzappitello, line 1: the shape of vehicle position dataframes before the `transform_vp_timestamps` functions are one column wider now that we have the multi-carriage-detail column.

tests/test_files/SPRINGBOARD/RT_VEHICLE_POSITIONS/year=2023/month=5/day=8/hour=12/1613b49e4fa1459eabe9c83553ef1045-0.parquet
mzappitello, line 1: everything else is just adding an extra column on the source data for tests.

(one could argue that we should modify the code to handle the case where that column doesn't exist in a source field?)

rymarczy (approved):
LGTM

PR #295, opened 2024-03-27
FIX: Update poetry dot lock file

mzappitello:
When I updated the pyproject.toml file I forgot to update the poetry.lock file. Do that now before it messes things up for development.


rymarczy (approved):
LGTM

PR #297, opened 2024-03-28
Feat: Check Version on Public Archive Files

mzappitello:
For flat files in the public archive, check the metadata for version information and overwrite files that are not using the current version of the rail performance manager. Files with mismatched versions that will not be overwritten are deleted.

Asana Task: [Add Version Metadata to Public Flat Files
](https://app.asana.com/0/1189492770004753/1206931284338815/f)

### DIFF
--- a/src/lamp_py/performance_manager/flat_file.py
+++ b/src/lamp_py/performance_manager/flat_file.py
@@ -337,6 +337,13 @@ def write_daily_table(
                 VehicleEvents.vp_stop_timestamp.is_not(None),
             ),
         )
+        .order_by(
+            sa.func.coalesce(
+                VehicleEvents.vp_move_timestamp,
+                VehicleEvents.vp_stop_timestamp,
+                VehicleEvents.tu_stop_timestamp,
+            )
+        )
     )
 
     flat_schema = pyarrow.schema(
--- a/src/lamp_py/performance_manager/flat_file.py
+++ b/src/lamp_py/performance_manager/flat_file.py
@@ -3,13 +3,16 @@ import re
 from typing import Set, List
 from datetime import datetime
 
+from botocore.exceptions import ClientError
 import sqlalchemy as sa
 import pandas
 import pyarrow
 
 from lamp_py.aws.s3 import (
+    delete_object,
     file_list_from_s3,
     file_list_from_s3_with_details,
+    object_metadata,
     upload_file,
 )
 from lamp_py.performance_manager.gtfs_utils import (
@@ -36,6 +39,8 @@ class S3Archive:
         "lamp", "subway-on-time-performance-v1"
     )
     INDEX_FILENAME = "index.csv"
+    VERSION_KEY = "rpm_version"
+    RPM_VERSION = "1.0.0"
 
 
 def dates_to_update(db_manager: DatabaseManager) -> Set[datetime]:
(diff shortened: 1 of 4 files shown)


src/lamp_py/performance_manager/flat_file.py
```diff
@@ -36,14 +38,23 @@ class S3Archive:
         "lamp", "subway-on-time-performance-v1"
     )
     INDEX_FILENAME = "index.csv"
+    VERSION_KEY = "rpm_version"
+    RPM_VERSION = "1.0.0"
 
 
-def dates_to_update(db_manager: DatabaseManager) -> Set[datetime]:
+def dates_to_update(
```
mzappitello, line 45: this func now returns two sets; one is the set of dates that need to be archived, one that has old metadata and needs to be deleted.

the latter set should always be empty, since any data under an old version _should_ also be in our rds but it felt safer to do it this way.

src/lamp_py/performance_manager/flat_file.py
```diff
@@ -59,49 +70,65 @@ def db_service_dates_to_datetimes(df: pandas.DataFrame) -> Set[datetime]:
...
-                datetimes.add(
-                    datetime.strptime(match.group("date"), "%Y-%m-%d")
-                )
+                date = datetime.strptime(match.group("date"), "%Y-%m-%d")
+                # check that the version of this file is equal to the current version
+                metadata = object_metadata(filepath).get(S3Archive.VERSION_KEY)
+                if metadata == S3Archive.RPM_VERSION:
+                    this_version.add(date)
+                else:
+                    old_version.add(date)
```
mzappitello, line 98: i'm not a _big_ fan of doing this call for each file, but it feels necessary?

the other thing i considered was keeping track of it in the index csv, but it was hard to control for errors if something goes wrong mid process.

src/lamp_py/performance_manager/flat_file.py
```diff
@@ -179,6 +215,7 @@ def write_csv_index() -> None:
             S3Archive.RAIL_PERFORMANCE_PREFIX,
             S3Archive.INDEX_FILENAME,
         ),
+        extra_args={"Metadata": {S3Archive.VERSION_KEY: S3Archive.RPM_VERSION}},
```
mzappitello, line 218: add a version to the metadata for each file uploaded.

mzappitello:
the other thought i had while doing this was putting the version in `lamp_py/performance_manager/__version__.py` and using that, but i thin kwe decided on doing it per file type rather than for the whole project.

src/lamp_py/performance_manager/flat_file.py
```diff
@@ -114,33 +119,83 @@ def write_flat_files(db_manager: DatabaseManager) -> None:
...
+            bucket_name=S3Archive.BUCKET_NAME,
+            file_prefix=S3Archive.RAIL_PERFORMANCE_PREFIX,
+        )
+
+        for file in files_to_remove:
+            success = delete_object(file)
+            if not success:
+                raise RuntimeError(
+                    f"Failed to delete {file} when updating flat files"
+                )
```
mzappitello, line 198: new function that checks the metadata of the csv file for the version. if it detects a version mismatch it deletes everything in the "directory", which will cause all archive files to be re-written.

tests/performance_manager/test_performance_manager.py
```diff
@@ -204,6 +204,28 @@ def fixture_flat_file_s3_patch(monkeypatch: MonkeyPatch) -> Iterator[None]:
...
+            obj
+            == f"{test_archive_value}/lamp/subway-on-time-performance-v1/index.csv"
+        )
+        return {S3Archive.VERSION_KEY: test_version_value}
+
+    monkeypatch.setattr(
+        "lamp_py.performance_manager.flat_file.object_metadata",
+        mock__object_metadata,
+    )
+
```
mzappitello, line 228: monkeypatch object metadata to confirm we're only checking csv data and return the test version.

i didn't write new tests to ensure we delete everything if the version mismatches, but can write one if you like.

rymarczy (approved):
LGTM

rymarczy (approved):
:cake:

PR #298, opened 2024-03-29
FIX: Remove vestigial dot pylintrc

mzappitello:
Our pylint configuration is in `pyproject.toml`. We had a `.pylintrc` file from before times. It didn't matter when we had a python_src directory, but now that everything is at the top level it causes problems when running pylint via poetry locally. (It still worked correctly on github because we don't copy the .pylintrc file into our testing environments).

Remove that file to avoid conflicts in local development.


rymarczy (approved):
LGTM

PR #303, opened 2024-04-01
FEAT: run sync_stop_sequence migration

rymarczy:
During loading of data into PROD, the `sync_stop_sequence` update query was disabled to speed up processing. Now that PROD is caught up to real-time, this migration will update all `sync_stop_sequence` values for each service date. 

This branch can be deployed to production with a new version tag. 

Asana Task: https://app.asana.com/0/1205827492903547/1206680656631279


src/lamp_py/migrations/versions/performance_manager_prod/002_f09e853d5672_update_prod_stop_sync.py
```diff
@@ -0,0 +1,418 @@
...
+            ;
+        """
+        update_log = ProcessLogger(
+            "migrate_sync_stop_sequence",
+            service_date=service_date,
+            static_version_key=static_version_key,
+        )
+        update_log.log_start()
+        op.execute(stop_sync_update)
+        update_log.log_complete()
```
rymarczy, line 414: This should let us see the migration progress in splunk...

src/lamp_py/migrations/versions/performance_manager_prod/002_f09e853d5672_update_prod_stop_sync.py
```diff
@@ -0,0 +1,418 @@
...
+                            WHERE
+                                static_route_patterns.static_version_key = {static_version_key}
+                                AND (static_route_patterns.route_pattern_typicality = 1
+                                    OR static_route_patterns.route_pattern_typicality = 5)
+                            ORDER BY
+                                COALESCE(static_trips.branch_route_id,
+                                static_trips.trunk_route_id),
+                                static_route_patterns.direction_id,
+                                static_route_patterns.static_version_key,
+                                static_route_patterns.route_pattern_typicality DESC) AS canon_trips
```
mzappitello, line 99: is there a way that this could be a CTE instead of being repeated throughout this query?

not a big deal, just trying to think about how these queries work with best practices.

mzappitello:
this all looks like the same logic we've used before. a question on how its implemented, but LGTM. i guess we can merge this in now. i think the versioning commit has to wait until we switch over to using prod data.

src/lamp_py/migrations/versions/performance_manager_prod/002_f09e853d5672_update_prod_stop_sync.py
rymarczy: Unfortunately, CTE's have a performance penalty in this application, since all results of a CTE have to be materialized before a query can be executed. In-lining the entire query allows the SQL engine to optimize the query, which is supposed to improve query performance. 

Risking the performance penalty of CTE's for this could significantly extend the migration duration.

PR #307, opened 2024-04-03
Add extra logging to GTFS RT Rail Parquet Creation

mzappitello:
On prod, the ECS is failing somewhere inside of the `HyperRtRail.update_parquet` function. Add extra logging so it can be monitored both on staging and (going forward) on prod.


rymarczy (approved):
LGTM

PR #308, opened 2024-04-04
FIX: performance of l1_rt_trips.py updates

rymarczy:
This will hopefully fix a lot of the performance issues we are seeing in the `l1_rt_trips.py` file.

Two main changes:

1. 
Don't update trunk_route_id and branch_route_id with database functions, instead `update_branch_trunk_route_id` function will perform this operation locally, and then update the `vehicle_trips` table.

2.
perform updates in `update_stop_sequence` function on unique `service_date` and `static_version_key` combinations with for loop. Previous implementation used a where statement that compared to an array of values, that I think was not allowing our indexes to be used.

### DIFF
--- a/src/lamp_py/performance_manager/l1_rt_trips.py
+++ b/src/lamp_py/performance_manager/l1_rt_trips.py
@@ -1,3 +1,6 @@
+from typing import Optional
+
+import pandas
 import sqlalchemy as sa
 from sqlalchemy.dialects import postgresql
 from sqlalchemy.sql.expression import bindparam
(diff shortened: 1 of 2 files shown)


src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -326,12 +329,145 @@ def update_start_times(db_manager: DatabaseManager) -> None:
...
+                distinct_red_trips.c.pm_trip_id == VehicleEvents.pm_trip_id,
+            ),
+        )
+    )
+
+    process_logger = ProcessLogger("l1_trips.update_branch_trunk_route_id")
+    process_logger.log_start()
+
+    trips_df = db_manager.select_as_dataframe(distinct_trips)
+    red_events_df = db_manager.select_as_dataframe(red_events)
```
rymarczy, line 385: pull two dataframes from DB:

trips_df:  all trips loaded during this event loop (with comparison to TempEventCompare table)
red_events_df: `pm_trip_id` and `stop_id` for all trips found in TempEventCompare table (for redline branch_route_id assignment)

src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -326,12 +329,145 @@ def update_start_times(db_manager: DatabaseManager) -> None:
...
+        if str(route_id).startswith("Green"):
+            return "Green"
+        return route_id
+
+    def apply_branch_route_id(record: pandas.Series) -> Optional[str]:
+        if str(record["route_id"]).startswith("Green"):
+            return record["route_id"]
+        if str(record["route_id"]) == "Red":
+            return get_red_branch(record["pm_trip_id"])
+        return None
```
rymarczy, line 432: Slightly more complicated `branch_route_id` assignment using pandas apply.

src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -326,12 +329,145 @@ def update_start_times(db_manager: DatabaseManager) -> None:
...
+            "70096",
+        }
+        trip_stop_ids = set(
+            red_events_df[red_events_df["pm_trip_id"] == pm_trip_id]["stop_id"]
+        )
+        if trip_stop_ids & ashmont_stop_ids:
+            return "Red-A"
+        if trip_stop_ids & braintree_stop_ids:
+            return "Red-B"
+        return None
```
rymarczy, line 420: determine if `Red-A` or `Red-B` through use of set intersection. convert series of all stop_id's for a pm_trip_id into a set and compare to specified braintree or ashmont stop_ids

src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -374,7 +510,7 @@ def update_trip_stop_counts(db_manager: DatabaseManager) -> None:
 
     process_logger = ProcessLogger("l1_trips.update_trip_stop_counts")
     process_logger.log_start()
-    db_manager.execute(update_query)
+    db_manager.execute(update_query, disable_trip_tigger=True)
```
rymarczy, line 513: disable trigger that was applying `branch_route_id` and `trunk_route_id` previously, these vestiges of the database functions should be totally removed in a future PR

src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -517,314 +653,315 @@ def update_stop_sequence(db_manager: DatabaseManager) -> None:
...
-        .select_from(VehicleEvents)
-        .join(
-            VehicleTrips,
-            VehicleEvents.pm_trip_id == VehicleTrips.pm_trip_id,
-        )
-        .join(
-            static_canon,
-            sa.and_(
-                VehicleTrips.direction_id == static_canon.c.direction_id,
+    for record in db_manager.select_as_list(distinct_query):
```
rymarczy, line 664: These queries are the exact same, except running in a for loop with specific `service_date` and `static_version_key` values in the where clauses. In testing this dramatically improved the performance of these queries from ~100 seconds to < 5 seconds locally.

src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -326,12 +329,145 @@ def update_start_times(db_manager: DatabaseManager) -> None:
...
+            trunk_route_id=bindparam("b_trunk_route_id"),
+        )
+    )
+
+    db_manager.execute_with_data(
+        start_times_update_query,
+        trips_df,
+        disable_trip_tigger=True,
+    )
+    process_logger.log_complete()
```
mzappitello, line 465: i generally find these types of DB manipulations easier to follow. not sure what that says about me. i've often wondered about doing things this way vs some of our more complicated db computes.

src/lamp_py/performance_manager/l1_rt_trips.py
mzappitello: do we turn the trip trigger on anywhere?

src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -326,12 +329,145 @@ def update_start_times(db_manager: DatabaseManager) -> None:
...
+            "70099",
+            "70100",
+            "70101",
+            "70102",
+            "70103",
+            "70104",
+            "70105",
+            "70095",
+            "70096",
+        }
```
mzappitello, line 412: i have an intuition that these values should be stored in a different place, but i think thats more something to think about if / when we convert from DB to Flat File architecture.

mzappitello (approved):
this looks good to me.  i have it in a comment, but i've often wondered about running some of our complicated database computes locally instead of on the db using sql.

i guess thats what we're hoping to do with a flat file structure. i think it might be a good way to do the eventual refactor, with pulling the biz logic into python with pandas or poloars or pyarrow while still keeping the data in the db before removing it all together.

src/lamp_py/performance_manager/l1_rt_trips.py
rymarczy: not anymore

PR #309, opened 2024-04-04
FIX: prod metdata migration to re-process failed parquet files

rymarczy:
Execute migration on PROD metadata_log table to re-process parquet files from morning of April 3, 2024.


mzappitello (approved):
🍰

PR #310, opened 2024-04-04
feat: log add_metadata calls to ProcessLogger

rymarczy:
print log statements when `add_metadata` called on `ProcessLogger`.


mzappitello (approved):
🍰

PR #311, opened 2024-04-05
FIX: Parquet writing batch_size and profile logging

rymarczy:
Very large parquet input files on PROD were resulting in ECS crashes from excessive memory usage.

After local testing, excessive memory usage was re-produced and `batch_size` values were dialed in to limit memory usage during parquet file writing operations. 

Also added some basic system profiling to logging statements `free_disk_mb` & `free_mem_pct` will hopefully help troubleshoot these types of issues more quickly in the future.

### DIFF
--- a/src/lamp_py/runtime_utils/process_logger.py
+++ b/src/lamp_py/runtime_utils/process_logger.py
@@ -3,6 +3,7 @@ import os
 import time
 import uuid
 import shutil
+import psutil
 from typing import Any, Dict, Union, Optional
 
 MdValues = Optional[Union[str, int, float]]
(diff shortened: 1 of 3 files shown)


src/lamp_py/runtime_utils/process_logger.py
```diff
@@ -61,8 +70,10 @@ def add_metadata(self, **metadata: MdValues) -> None:
             if key in ProcessLogger.protected_keys:
                 continue
             self.metadata[str(key)] = str(value)
-        self.default_data["status"] = "add_metadata"
-        logging.info(self._get_log_string())
+
+        if self.default_data.get("status") is not None:
+            self.default_data["status"] = "add_metadata"
+            logging.info(self._get_log_string())
```
mzappitello, line 76: is this something we'll want on all the time?

mzappitello (approved):
🍰

src/lamp_py/runtime_utils/process_logger.py
rymarczy: I put this in the stop log printing before `log_start()` gets called, was kind of confusing for me to see logs in splunk before the log had actually started, but I can add that back if you'd prefer.

PR #315, opened 2024-04-17
FEAT: MSSQL Server Connection Scaffolding

rymarczy:
This change adds the scaffolding required for our Ingestion application to make a connection to the TransitMaster Microsoft SQL Server. 

With this scaffolding we are able to use the `MSSQLManager` class to return query results as a list of dicts, pandas dataframe or write directly to a parquet file. 

Asana Task: https://app.asana.com/0/1205827492903547/1206974089170980

### DIFF
--- a/.env
+++ b/.env
@@ -17,6 +17,12 @@ RPM_DB_USER=postgres
 RPM_DB_PASSWORD=postgres
 ALEMBIC_RPM_DB_NAME=performance_manager_prod
 
+# MSSQL TransitMaster database
+TM_DB_HOST=do_update
+TM_DB_NAME=do_update
+TM_DB_USER=do_update
+TM_DB_PASSWORD=do_update
+
 # s3 locations
 SPRINGBOARD_BUCKET=mbta-ctd-dataplatform-dev-springboard
 ARCHIVE_BUCKET=mbta-ctd-dataplatform-dev-archive
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,15 +14,18 @@ RUN curl https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem -o /u
 RUN chmod a=r /usr/local/share/amazon-certs.pem
 
 # Install MSSQL ODBC 18 Driver
-RUN mkdir -m 0755 -p /etc/apt/keyrings/
-RUN curl -fsSL https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor -o /etc/apt/keyrings/microsoft.gpg
-RUN echo "deb [signed-by=/etc/apt/keyrings/microsoft.gpg] https://packages.microsoft.com/debian/12/prod bookworm main" | tee /etc/apt/sources.list.d/mssql-release.list
-RUN apt-get update
-RUN ACCEPT_EULA=Y apt-get install -y msodbcsql18
+# for TransitMaster DB connection and ingestion
+RUN mkdir -m 0755 -p /etc/apt/keyrings/ \
+    && curl -fsSL https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor -o /etc/apt/keyrings/microsoft.gpg \
+    && echo "deb [signed-by=/etc/apt/keyrings/microsoft.gpg] https://packages.microsoft.com/debian/12/prod bookworm main" | tee /etc/apt/sources.list.d/mssql-release.list \
+    && apt-get update \
+    && ACCEPT_EULA=Y apt-get install -y msodbcsql18
 
-# modify openssl config to allow TLSv1 connection for TransitMaster server
-RUN sed -i 's/\[openssl_init\]/# [openssl_init]/' /etc/ssl/openssl.cnf
-RUN echo '\n\n[openssl_init]\nssl_conf = ssl_sect\n\n[ssl_sect]\nsystem_default = ssl_default_sect\n\n[ssl_default_sect]\nMinProtocol = TLSv1\nCipherString = DEFAULT@SECLEVEL=0\n' >> /etc/ssl/openssl.cnf
+# modify openssl config to allow TLSv1 connection
+# moves [openssl_init] section and creates [ssl_default_sect] section to allow TLSv1
+# for TransitMaster DB connection and ingestion
+RUN sed -i 's/\[openssl_init\]/# [openssl_init]/' /etc/ssl/openssl.cnf \
+    && echo '\n\n[openssl_init]\nssl_conf = ssl_sect\n\n[ssl_sect]\nsystem_default = ssl_default_sect\n\n[ssl_default_sect]\nMinProtocol = TLSv1\nCipherString = DEFAULT@SECLEVEL=0\n' >> /etc/ssl/openssl.cnf
 
 # Install poetry
 RUN pip install -U pip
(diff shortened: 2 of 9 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.10-slim
```
rymarczy, line 1: Changing this to specify bookworm because the Microsoft ODBC driver install has to target a specific version of debian.

Dockerfile
```diff
@@ -7,12 +7,23 @@ ENV PYTHONUNBUFFERED 1
...
+RUN mkdir -m 0755 -p /etc/apt/keyrings/
+RUN curl -fsSL https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor -o /etc/apt/keyrings/microsoft.gpg
+RUN echo "deb [signed-by=/etc/apt/keyrings/microsoft.gpg] https://packages.microsoft.com/debian/12/prod bookworm main" | tee /etc/apt/sources.list.d/mssql-release.list
+RUN apt-get update
+RUN ACCEPT_EULA=Y apt-get install -y msodbcsql18
+
+# modify openssl config to allow TLSv1 connection for TransitMaster server
+RUN sed -i 's/\[openssl_init\]/# [openssl_init]/' /etc/ssl/openssl.cnf
+RUN echo '\n\n[openssl_init]\nssl_conf = ssl_sect\n\n[ssl_sect]\nsystem_default = ssl_default_sect\n\n[ssl_default_sect]\nMinProtocol = TLSv1\nCipherString = DEFAULT@SECLEVEL=0\n' >> /etc/ssl/openssl.cnf
+
```
rymarczy, line 26: There are probably other ways to update this openssl config, but I found this one to be the most straightforward....

Dockerfile
```diff
@@ -7,12 +7,23 @@ ENV PYTHONUNBUFFERED 1
...
+RUN apt-get install -y libpq-dev gcc curl gpg
 
 # Fetch Amazon RDS certificate chain
 RUN curl https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem -o /usr/local/share/amazon-certs.pem
 RUN chmod a=r /usr/local/share/amazon-certs.pem
 
+# Install MSSQL ODBC 18 Driver
+RUN mkdir -m 0755 -p /etc/apt/keyrings/
+RUN curl -fsSL https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor -o /etc/apt/keyrings/microsoft.gpg
+RUN echo "deb [signed-by=/etc/apt/keyrings/microsoft.gpg] https://packages.microsoft.com/debian/12/prod bookworm main" | tee /etc/apt/sources.list.d/mssql-release.list
```
rymarczy, line 19: Open to suggestions on how to add this microsoft gpg key, but again this seems the most straightforward / secure.

docker-compose.yml
```diff
@@ -38,6 +38,14 @@ services:
...
     command: ["poetry", "run", "performance_manager"]
 
+  test_mssql:
+    container_name: test_mssql
+    env_file: .env
+    build: .
+    working_dir: /lamp
+    network_mode: "host"
+    command: ["poetry", "run", "test_mssql"]
+
```
rymarczy, line 48: I will drop this before the final merge of this pull request into main, but this was helpful for me locally in confirming all of this works.

pyproject.toml
```diff
@@ -14,6 +14,7 @@ performance_manager = 'lamp_py.performance_manager.pipeline:start'
 bus_performance_manager = 'lamp_py.bus_performance_manager.pipeline:start'
 seed_metadata = 'lamp_py.postgres.seed_metadata:run'
 hyper_update = 'lamp_py.tableau.pipeline:start_hyper_updates'
+test_mssql = 'lamp_py.mssql.test_connect:start'
```
rymarczy, line 17: Again, can drop before final merge

src/lamp_py/mssql/mssql_utils.py
```diff
@@ -0,0 +1,173 @@
...
+        return engine
+    except Exception as exception:
+        process_logger.log_failure(exception)
+        raise exception
+
+
+class MSSQLManager:
+    """
+    manager class for rds application operations
+    """
```
rymarczy, line 68: All of the methods in here worked for me locally

src/lamp_py/mssql/test_connect.py
rymarczy, line 1: Will drop this file before final merge into main, but this helped in confirming everything was working

.env
mzappitello, line 1: not for this pr, but we should generate a sample.env that has these things in it and then git ignore our actual .env files.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.10-slim
+FROM python:3.10-slim-bookworm
```
mzappitello, line 1: why this change?

Dockerfile
mzappitello: i looked around for other ways to modify these conf files, and it seems like using `echo` and `sed` is a fine way for doing simple changes like this.

Dockerfile
mzappitello: i think this is fine. maybe a few comments about what the curl and echo statements are doing, but this seems fine to me.

mzappitello: it might be a good idea while we're here to add comments about why each dependency is needed. like this one is only going to be used for the TransitMaster ingestion application. Future LAMP will be happy that they know why each dependency was added to this shared dockerfile.

mzappitello: one last note is that it seems like we should chain all these commands together as [per dockers best practices](https://stackoverflow.com/a/39234643)

```
# Install necessary tools and MSSQL ODBC 18 Driver
RUN mkdir -m 0755 -p /etc/apt/keyrings/ \
    && curl -fsSL https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor -o /etc/apt/keyrings/microsoft.gpg \
    && echo "deb [signed-by=/etc/apt/keyrings/microsoft.gpg] https://packages.microsoft.com/debian/12/prod bookworm main" | tee /etc/apt/sources.list.d/mssql-release.list \
    && apt-get update \
    && ACCEPT_EULA=Y apt-get install -y msodbcsql18
```

docker-compose.yml
mzappitello: i had to add a block to avoid DNS errors.

```
    extra_hosts:
      - "HSTMDM01:172.29.21.44"
 ```

Dockerfile
rymarczy: I mentioned it in a comment above, the Microsoft drivers we're installing have to target a debian version, so explicitly calling this out to avoid future conflicts

src/lamp_py/mssql/mssql_utils.py
mzappitello: these all worked for me as expected.

mzappitello:
:cake:

this lgtm. once i got it working i played around with it a bit and it worked as expected.

i did have a note on the dockerfile, but that could be a different ticket if we wanted.

mzappitello (approved):
🍰

this lgtm. once i got it working i played around with it a bit and it worked as expected.

i did have a note on the dockerfile, but that could be a different ticket if we wanted, especially since its applicable to more than just the changes you've made.

Dockerfile
rymarczy: Incorporated.

PR #318, opened 2024-04-19
FIX Correct Parquet Generation Assertion

mzappitello:
When writing parquet file, we “partition” files by creating Tables that have have the same value in all of the columns we're plan on partitioning by before writing. To ensure we processed data correctly before writing, the `write_parquet_file` has an assertion that the list of unique values in partition columns is exactly one. However, if the Table that is being passed in is empty, that list of unique values will have zero, and the assertion fill throw an exception.

* Add an additional `if` gate that checks that the table is not empty before this assertion to prevent this false exception.
* Add a message in the assertion that will allow us to debug this problem easier going forward.


src/lamp_py/aws/s3.py
```diff
@@ -476,7 +476,10 @@ def write_parquet_file(
     partition_strings = []
     for col in partition_cols:
         unique_list = pc.unique(table.column(col)).to_pylist()
-        assert len(unique_list) == 1
+        if len(table) > 0:
+            assert (
+                len(unique_list) == 1
+            ), f"Table {s3_dir} column {col} had {len(unique_list)} unique elements"
```
rymarczy, line 482: Won't this result in us writing a table to a location we don't want? Since the `partition_strings` is built as part of the final file path.

If we're having an issue with writing empty GTFS files I think we should abort this process further up the chain, and not write an empty file at all.

rymarczy (changes requested):
Think we might need a different approach?

rymarczy (approved):
:cake:

PR #325, opened 2024-04-29
FEAT: Remove Commuter Rail data from exports

rymarczy:
This change removes commuter rain data from flat file and OPMI ALL_RT export files. 

As tested on STAGING, these changes have no significant impact on query duration. 

flat file query remained constant at ~4 seconds, for one service_date

ALL_RT query stayed at ~12 seconds, for one service_date

Deployment of this change to PROD will result in re-creation of export datasets, which should be done at the end of the business day to avoid interference with Tableau dashboarding operations. 

Asana Task: https://app.asana.com/0/1205827492903547/1207059327607966


mzappitello (approved):
🍰

PR #327, opened 2024-05-03
FIX: Add paginator to file_list_from_s3_with_details

rymarczy:
`file_list_from_s3_with_details()` function was limited to returning a maximum of 1000 results without a paginator. This was breaking our writing of `index.csv` for the `subway-on-time-performance-v1` data export. 

Added `paginator` to return an unlimited number of results.


mzappitello (approved):
:cake:

PR #331, opened 2024-05-07
FEAT: Ingest GTFS-RT files continously

rymarczy:
This change updates the ingestion app to continuously update GTFS-RT parquet files.

Continuous updates are performed on parquet files saved locally, that are then synced with the expected S3 bucket. 

No more than 2 local parquet files, for each GTFS-RT type, are ever saved locally. 

The Metadata database is updated, with processed flags being reset, whenever an update is made to a parquet file. 

Asana Task: https://app.asana.com/0/1205827492903547/1207039261301265

### DIFF
--- a/src/lamp_py/ingestion/convert_gtfs_rt.py
+++ b/src/lamp_py/ingestion/convert_gtfs_rt.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import os
+import shutil
 from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass, field
 from datetime import datetime, timezone
@@ -10,8 +11,16 @@ from typing import Dict, Iterable, List, Optional, Tuple
 
 import pyarrow
 from pyarrow import fs
+import pyarrow.compute as pc
+import pyarrow.parquet as pq
+import pyarrow.dataset as pd
 
-from lamp_py.aws.s3 import move_s3_objects, write_parquet_file
+from lamp_py.aws.s3 import (
+    move_s3_objects,
+    file_list_from_s3,
+    download_file,
+    upload_file,
+)
 from lamp_py.runtime_utils.process_logger import ProcessLogger
 
 from .config_rt_alerts import RtAlertsDetail
@@ -76,10 +85,8 @@ class GtfsRtConverter(Converter):
         else:
             raise NoImplException(f"No Specialization for {config_type}")
 
-        # get the start of the current hour. any files written after this will
-        # not be ingested until the next go around.
-        now = datetime.now(tz=timezone.utc)
-        self.start_of_hour = now.replace(minute=0, second=0, microsecond=0)
+        self.pq_batch_size = 1024 * 256
+        self.tmp_folder = "/tmp/gtfs-rt-continuous"
 
         self.table_groups: Dict[datetime, TableData] = {}
 
@@ -99,18 +106,20 @@ class GtfsRtConverter(Converter):
         table_count = 0
         try:
             for table in self.process_files():
-                self.write_table(table)
+                if table.num_rows == 0:
+                    continue
+                self.continuous_pq_update(table)
                 self.move_s3_files()
 
-                # only count table if it contains data
-                if table.num_rows > 0:
-                    table_count += 1
+                table_count += 1
 
                 process_logger.add_metadata(table_count=table_count)
 
                 # limit number of tables produced on each event loop
                 if table_count >= max_tables_to_convert:
                     break
+
+            self.clean_local_folders()
 
         except Exception as exception:
             process_logger.log_failure(exception)
@@ -199,20 +208,8 @@ class GtfsRtConverter(Converter):
 
                 yield from self.yield_check(yield_threshold, process_logger)
 
-                # check if ready to end work because processing files past start_of_hour
-                # waiting for count of files > yield_threshold should allow for
-                # any work from previous hour to finish before exiting
-                if (
-                    result_dt >= self.start_of_hour
-                    and len(self.table_groups[timestamp_hr].files)
-                    > yield_threshold
-                ):
-                    break
-
-        # yield any remaining tables with next_hr_cnt > 0
-        # guaranteeing that the end of the hour was hit
-        # not sure if we would ever actually hit this
-        yield from self.yield_check(0, process_logger)
+        # yield any remaining tables
+        yield from self.yield_check(-1, process_logger)
 
         process_logger.add_metadata(file_count=0, number_of_rows=0)
         process_logger.log_complete()
@@ -231,10 +228,6 @@ class GtfsRtConverter(Converter):
         instead, ensure that a sufficient number for files from the next hour
         have been processed.
 
-        additionally, do not yield any tables from the current hour, as we want
-        limit the number of parquet files generated (ideally one per hour) and
-        more data for the current hour will be coming in later.
-
         @yield_threshold - how many files from the next hour have to be
             processed before considering _this_ hour complete.
         @process_logger - a process logger for the conversion process. log a
@@ -244,10 +237,7 @@ class GtfsRtConverter(Converter):
             data over the corse of an hour.
         """
         for iter_ts in list(self.table_groups.keys()):
-            if (
-                self.table_groups[iter_ts].next_hr_cnt > yield_threshold
-                and iter_ts < self.start_of_hour
-            ):
+            if self.table_groups[iter_ts].next_hr_cnt > yield_threshold:
                 self.archive_files += self.table_groups[iter_ts].files
 
                 table = self.table_groups[iter_ts].table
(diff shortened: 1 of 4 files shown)


tests/ingestion/test_gtfs_rt_converter.py
```diff
@@ -353,48 +353,3 @@ def test_bus_trip_updates_file_conversion() -> None:
...
-    assert converter.archive_files == [gtfs_rt_file_1]
-    assert len(converter.error_files) == 0
-
-    month_column = table.column("month")
-    for month in month_column:
-        assert month.as_py() == 1
-
-    day_column = table.column("day")
-    for day in day_column:
-        assert day.as_py() == 1
```
rymarczy, line 400: Not applicable anymore.

src/lamp_py/postgres/postgres_utils.py
```diff
@@ -581,8 +583,17 @@ def _rds_writer_process(metadata_queue: Queue[Optional[str]]) -> None:
...
+        insert_statement = (
+            postgresql.insert(MetadataLog.__table__)
+            .values(path=metadata_path)
+            .on_conflict_do_update(
+                index_elements=[MetadataLog.path],
+                set_={
+                    "rail_pm_processed": sa.false(),
+                    "rail_pm_process_fail": sa.false(),
+                    "created_on": now(),
+                },
```
rymarczy, line 595: If INSERT fails, update process flags to false, so file will be re-loaded into performance manager database.

src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -353,38 +343,151 @@ def gz_to_pyarrow(
...
+                with pq.ParquetWriter(
+                    join_table_path, schema=table.schema
+                ) as writer:
+                    for batch in join_ds.to_batches(
+                        batch_size=self.pq_batch_size
+                    ):
+                        writer.write_batch(batch)
+
+                os.replace(join_table_path, local_path)
+                os.remove(new_table_path)
```
mzappitello, line 446: this part feels like it could be its own method. that would let you keep the R0914 pylint rule on as well.

src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -353,38 +343,151 @@ def gz_to_pyarrow(
...
+
+            log.add_metadata(part_folder=part_folder, local_path=local_path)
+
+            # no existing table file, write new file
+            if local_path == "":
+                local_path = os.path.join(
+                    self.tmp_folder,
+                    part_folder,
+                    f"{datetime(year, month, day, hour).isoformat()}.parquet",
+                )
```
mzappitello, line 417: this has the building out of the filename we would look for on s3, with the s3 bucket instead of the temp folder obviously.

src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -353,38 +343,151 @@ def gz_to_pyarrow(
...
+        """
+        log = ProcessLogger("continuous_pq_update")
+        log.log_start()
+        try:
+            self.verify_table(table)
+
+            year = pc.unique(table.column("year")).to_pylist()[0]
+            month = pc.unique(table.column("month")).to_pylist()[0]
+            day = pc.unique(table.column("day")).to_pylist()[0]
+            hour = pc.unique(table.column("hour")).to_pylist()[0]
```
mzappitello, line 397: i could see these being put into a single function, `extract_datetime` that asserts if they are not all the same and returns a tuple of datetime ints.

src/lamp_py/ingestion/convert_gtfs_rt.py
mzappitello: with these changes the code looks something like

```
try:
    y, m, d, h = extract_datetime(table)
    
    file_suffix = os.path.join(
        LAMP_S3_PREFIX,
        str(self.config_type),
        f"year={year}",
        f"month={month}",
        f"day={day}",
        f"hour={hour}",
        f"{datetime(year, month, day, hour).isoformat()}.parquet",
    )
    
    if download_s3_locally(file_suffix):
        self.append_to_local_path(table, file_suffix)
    else:
        local_path = os.path.join(self.tmp_folder, file_suffix)
        pq.write_table(table, local_path)
        
    upload_path = os.path.join(os.environ["SPRINGBOARD_BUCKET"], file_suffix)
    upload_file(local_path, local_path)

    # error handling
```

src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -353,38 +343,151 @@ def gz_to_pyarrow(
...
             self.archive_files = []
+            log.log_failure(exception)
+
+    # pylint: enable=R0914
+
+    def clean_local_folders(self) -> None:
+        """
+        clean local temp folders
+        """
+        hours_to_keep = 2
```
mzappitello, line 472: wondering why you wanted to keep the last two hours.

src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -99,19 +106,21 @@ def convert(self) -> None:
         table_count = 0
         try:
             for table in self.process_files():
-                self.write_table(table)
+                if table.num_rows == 0:
```
mzappitello, line 109: it feels like this function could just work on a max number of files rather than a specific number of hours.

i guess thats all going to go away when we switch to daily archive files and if we want to handle that in that PR thats fine with me.

mzappitello (commented):
I left a few comments, mostly around the new `continuous_pq_update` function.

its outside the scope of this ticket, but i think we should start thinking about a utility class that models a parquet file we have on s3. this class would handle downloading from s3, appending data, and uploading data. we're doing that in a handful of places, and given the architecture we're working towards it'll only be more useful.

eventually, i could see this class handling the sorting problem we'll want to solve for filesize reasons and potentially extracting some data out of a file (like i used in the alerts processing).

src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -353,38 +343,151 @@ def gz_to_pyarrow(
...
+            bucket,
+            file_prefix=f"{part_folder}/",
+        )
+        if len(s3_files) == 1:
+            s3_path = s3_files[0].replace("s3://", "")
+            local_path = s3_path.replace(bucket, self.tmp_folder)
+            download_file(s3_path, local_path)
+            return local_path
+
+        return ""
```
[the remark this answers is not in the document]
rymarczy: changed this function to return a `bool`

src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -353,38 +343,151 @@ def gz_to_pyarrow(
...
+        bucket = os.environ["SPRINGBOARD_BUCKET"]
+        s3_files = file_list_from_s3(
+            bucket,
+            file_prefix=f"{part_folder}/",
+        )
+        if len(s3_files) == 1:
+            s3_path = s3_files[0].replace("s3://", "")
+            local_path = s3_path.replace(bucket, self.tmp_folder)
+            download_file(s3_path, local_path)
+            return local_path
```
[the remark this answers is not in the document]
rymarczy: This was a vestige of using UUID's as file names, where the file path could not be known ahead of time. I've updated to use the known datetime formatted path.

src/lamp_py/ingestion/convert_gtfs_rt.py
rymarczy: this function has been cleaned up and the pylint exception removed.

src/lamp_py/ingestion/convert_gtfs_rt.py
rymarczy: this has been partially incorporated, a datetime object (representing partition date info) is now returned from the function that verifies unique table values

src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -353,38 +343,160 @@ def gz_to_pyarrow(
...
+            upload_file(local_path, upload_path)
+            self.send_metadata(upload_path)
 
-        except Exception:
+            log.log_complete()
+
+        except Exception as exception:
+            shutil.rmtree(
+                os.path.join(self.tmp_folder, part_folder), ignore_errors=True
+            )
```
mzappitello, line 472: If the assertion in partition_dt throws, then this block will also throw since part folder is None.

mzappitello (commented):
The changes look good minus an exception case that needs to be handled differently.

src/lamp_py/ingestion/convert_gtfs_rt.py
rymarczy: updated this to just nuke the whole `self.config_type` folder and start fresh

mzappitello (approved):
🍰

PR #332, opened 2024-05-09
FIX: update sync_stop_sequence by trunk_route_id

rymarczy:
The sync_stop_sequence value should be updated by trunk_route_id, since it is agnostic to whichever branch_route_id the trip is associated with. 

This change updates the sync_stop_sequence query to update based on trunk_route_id and includes migration files for all environments to update old sync_stop_sequence values.

This issue was originally raised by Heather from ops analytics.


mzappitello (approved):
🍰

PR #339, opened 2024-05-16
FIX: Parquet hive partition reading failing on continuous ingest files

rymarczy:
ParquetDataset calls to read a hive partition are failing on the continuously ingested GTFS-RT parquet files. This is because I forgot to drop the hive partition data columns before writing the parquet files.


mzappitello (approved):
🍰

PR #340, opened 2024-05-16
FIX: Test deploy action that allows zero desired count

rymarczy:
I created a PR for the mbta/actions/deploy-ecs action to allow zero `desiredCount` values for ECS instances that are not running during a deploy. 

https://github.com/mbta/actions/pull/56

@paulswartz requested we test this feat before merging to the actions repo.

### DIFF
--- a/.github/workflows/deploy-base.yaml
+++ b/.github/workflows/deploy-base.yaml
@@ -86,12 +86,13 @@ jobs:
       - name: Deploy Bus Performance Manager Application
         id: deploy-bus-performance-manager
         if: ${{ inputs.deploy-bus-pm }}
-        uses: mbta/actions/deploy-ecs@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-base.yaml
```diff
@@ -86,12 +86,13 @@ jobs:
       - name: Deploy Bus Performance Manager Application
         id: deploy-bus-performance-manager
         if: ${{ inputs.deploy-bus-pm }}
-        uses: mbta/actions/deploy-ecs@v2
```
mzappitello, line 89: any idea why this is in a branch?

rymarczy: I think v2 is a version tag...

mzappitello (approved):
merge away

PR #353, opened 2024-05-25
FIX: Only pull 1 static parquet file from folder

rymarczy:
Continuous ingestion is causing 2 parquet files to be placed into a single static file partition on our springboard bucket. 

The static table loading process should never be loading more than 1 file from a folder. 

This is a stop-gap.


mzappitello (approved):
LGTM

PR #359, opened 2024-05-31
FIX: Add "s3://" prefix to Alerts S3 path string

mzappitello:
Not having this was causing the alerts hyper job to fail.


rymarczy (approved):
:cake:

PR #360, opened 2024-05-31
FIX: Add SqlType.timestamp_tz for pq -> hyper type

mzappitello:
The Alerts dataset has timestamp types with timezone information in them. Add a conversion in the `convert_parquet_dtype` method that will correctly convert these pyarrow types into hyper types.


rymarczy (approved):
:cake:

PR #361, opened 2024-06-03
FIX: GTFS Ingestion version_key based on published_dt

rymarczy:
Our previous implementation of determining a `version_key` was based on the file modified datetime of the `feed_info.txt` file in a GTFS zip archive. This method does not work for ingesting GTFS schedule information from the archive feed. 

GTFS Schedules in the archive feed have their feed_info.txt files modified during the creation of the next GTFS schedule, to update the `feed_end_date` field. Because of this, we will base the `version_key` on the `published_dt` that is extracted from the `feed_version` text of the `feed_info.txt` file.

### DIFF
--- a/src/lamp_py/ingestion/convert_gtfs.py
+++ b/src/lamp_py/ingestion/convert_gtfs.py
@@ -1,7 +1,9 @@
 import os
 import zipfile
-from datetime import datetime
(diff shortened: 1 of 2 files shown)


mzappitello (approved):
🍰

PR #364, opened 2024-06-03
Updates to Alerts Processing

mzappitello:
A few different issues were causing problems for downstream users w.r.t. the alerts dataset. First, we couldn't publish to tableau as all timestamp fields were coming through as unsigned int 64 types and hyper doesn't have a concept of unsigned ints. Second, when DuckDB reads parquet files, it converts all timezoned datetime types to UTC, causing issues for people using that workflow. Third, we weren't partitioning our files very well, making them slow to work with when polling them from s3. Fix these issues and update tests as appropriate.

* All timestamp fields are now `int64` instead of `uint64`.
* Datetime conversion functions now returns a date time that has been localized to Boston time but does not contain the tz info.
* Partition by month along "last_modified_timestamp". It can always be changed later.
* Update the version for the alerts file and create an s3 function that that checks for version information.

Asana Task:
[📣 Alerts datetime fields are not in Eastern Time](https://app.asana.com/0/1205827492903547/1207375565211452/f)
[📣 Publish alerts data to Tableau](https://app.asana.com/0/1205827492903547/1207101442990597/f)


src/lamp_py/aws/s3.py
```diff
@@ -181,6 +181,21 @@ def object_metadata(obj: str) -> Dict[str, str]:
...
+    """
+    try:
+        remote_version = object_metadata(obj).get("lamp_version", "")
+
+        return remote_version == version
+
+    except Exception:
+        return False
+
+
```
mzappitello, line 198: new utility function. i think its best to keep the key for our versions only in this file going forward to avoid typos.

should probably update the upload file signature as well.

src/lamp_py/performance_manager/alerts.py
```diff
@@ -29,8 +31,8 @@ class AlertsS3Info:
     s3_path: str = "s3://" + os.path.join(
         bucket_name, "lamp", "tableau", "alerts", "LAMP_RT_ALERTS.parquet"
     )
-    version_key: str = "version"
-    file_version: str = "1.0.0"
+    version_key: str = "lamp_version"
+    file_version: str = "1.1.0"
```
mzappitello, line 35: bump version and also use the right version key.

src/lamp_py/performance_manager/alerts.py
```diff
@@ -48,36 +50,18 @@ class AlertsS3Info:
...
-            (
-                "active_period.start_datetime",
-                pyarrow.timestamp("ms", tz="America/New_York"),
-            ),
-            ("active_period.start_timestamp", pyarrow.uint64()),
-            (
-                "active_period.end_datetime",
-                pyarrow.timestamp("ms", tz="America/New_York"),
-            ),
-            ("active_period.end_timestamp", pyarrow.uint64()),
```
mzappitello, line 80: `uint64` -> `int64`

remove tz info from timestamp column types. they are localized to boston, but without the tz info. downstream users can assume they are in boston time.

src/lamp_py/performance_manager/alerts.py
```diff
@@ -93,26 +77,28 @@ class AlertParquetHandler:
...
-    def __init__(self) -> None:
+    def __init__(self, update_historical: bool) -> None:
         self.s3_path: str = AlertsS3Info.s3_path
 
         self.local_path: str = os.path.join("/tmp", "alerts.parquet")
-        self.version_key: str = AlertsS3Info.version_key
-        self.file_version: str = AlertsS3Info.file_version
         self.parquet_schema: pyarrow.Schema = AlertsS3Info.parquet_schema
 
-        self.local_exists: bool = self.download_data()
```
mzappitello, line 104: use os.path.exists instead of this var.

src/lamp_py/performance_manager/alerts.py
```diff
@@ -93,26 +77,28 @@ class AlertParquetHandler:
     This class handles all of the interactions with alert data thats stored as a parquet file on s3.
     """
 
-    def __init__(self) -> None:
+    def __init__(self, update_historical: bool) -> None:
```
mzappitello, line 80: only pull historical data if its being updated.

src/lamp_py/performance_manager/alerts.py
```diff
@@ -125,67 +111,64 @@ def append_new_records(self, alerts: pandas.DataFrame) -> None:
         """
         append alerts to the end of the local parquet file using batches to keep memory usage lower
         """
-        self.new_data = True
```
mzappitello, line 128: update this function to use the new row group logic.

src/lamp_py/performance_manager/alerts.py
```diff
@@ -331,12 +318,16 @@ def unix_to_est(unix_time: Optional[int]) -> Optional[datetime]:
     if unix_time is None or pandas.isna(unix_time):
         return None
 
-    # Create a timezone-aware datetime object in UTC
```
mzappitello, line 334: update this function to return a datetime that is localized to boston but without the tz info.

src/lamp_py/performance_manager/alerts.py
```diff
@@ -475,10 +466,15 @@ def get_unprocessed_alert_files(
...
-    )
+    if unprocessed_only:
+        read_md = sa.select(MetadataLog.pk_id, MetadataLog.path).where(
+            (MetadataLog.rail_pm_processed == sa.false())
+            & (MetadataLog.path.contains("RT_ALERTS"))
+        )
+    else:
+        read_md = sa.select(MetadataLog.pk_id, MetadataLog.path).where(
+            (MetadataLog.path.contains("RT_ALERTS"))
+        )
```
mzappitello, line 477: we want _all_ alerts data if we're regenerating this file.

src/lamp_py/performance_manager/alerts.py
```diff
@@ -489,22 +485,35 @@ def process_alerts(md_db_manager: DatabaseManager) -> None:
     """
     process_logger = ProcessLogger("process_alerts")
     process_logger.log_start()
-    unprocessed_records = get_unprocessed_alert_files(md_db_manager)
+
+    version_match = version_check(
+        obj=AlertsS3Info.s3_path, version=AlertsS3Info.file_version
+    )
```
mzappitello, line 491: check for a version match. if regenerating, we'll pull _all_ files from springboard bucket and we won't pull historical record from s3.

tests/performance_manager/test_alerts.py
mzappitello, line 1: a few patches to make the tests match new logic.

rymarczy:
The polars upgrade that this triggered requires the `dtypes` parameter of `pl.read_csv` to be renamed to `schema_overrides`.

It only occurs in two file.
    src/lamp_py/ingestion/compress_gtfs/schedule_details.py
    src/lamp_py/ingestion/utils.py

But some other apps will probably break without this change

src/lamp_py/performance_manager/alerts.py
```diff
@@ -125,67 +111,64 @@ def append_new_records(self, alerts: pandas.DataFrame) -> None:
...
-        """
-        download the s3 parquet file to a temp location if the versions match.
-        return True if the file was downloaded successfully, False otherwise.
-        """
-        try:
-            lamp_version = object_metadata(self.s3_path).get(
-                self.version_key, ""
-            )
+        now = datetime.now(tz=BOSTON_TZ)
+        start = datetime(2028, 1, 1, tzinfo=BOSTON_TZ)
```
rymarczy, line 144: I would recommend against using a tz here, other than `UTC`, since the timestamp field we're partitioning on is a UTC value, would probably make debugging issues in partition more difficult. 

Also shouldn't need to hard-code a start datetime here, could use `pyarrow.compute.min` function on `partition_key` column of the dataset to pull this value dynamically. 

Using `now` for the final partition could also be an issue, in my testing there are a huge number of alerts that are projected to start past the current date, So maybe also pulling the max value from `partition_key` column and doing the partitions dynamically (maybe every 3 months) would be a better approach.

src/lamp_py/performance_manager/alerts.py
```diff
@@ -125,67 +111,64 @@ def append_new_records(self, alerts: pandas.DataFrame) -> None:
...
-            )
+        with pq.ParquetWriter(new_path, schema=self.parquet_schema) as writer:
+            while start < now:
+                end = start + relativedelta(months=1)
+                if end < now:
+                    table = joined_ds.filter(
+                        (pc.field(partition_key) >= start.timestamp())
+                        & (pc.field(partition_key) < end.timestamp())
+                    ).to_table()
+                else:
```
rymarczy, line 156: will this else block ever get hit the way the `while` loop is structured?

src/lamp_py/performance_manager/alerts.py
```diff
@@ -125,67 +111,64 @@ def append_new_records(self, alerts: pandas.DataFrame) -> None:
...
-            if lamp_version != self.file_version:
-                return False
+        partition_key = "last_modified_timestamp"
 
-            return download_file(
-                object_path=self.s3_path,
-                file_name=self.local_path,
-            )
+        with pq.ParquetWriter(new_path, schema=self.parquet_schema) as writer:
+            while start < now:
```
rymarczy, line 149: This is missing a block to write dataset records where ` partition_key` is `null`. Unless we can be certain there won't be any `null` values in the `parition_key` column.

rymarczy (changes requested):
A few things need to be fixed around the partition writing. 

We're you able to run through all the production data with this new process on your local machine? Or is the plan to try it out on staging and see if it goes before pushing to prod?

src/lamp_py/performance_manager/alerts.py
rymarczy: Nevermind, on this one

src/lamp_py/aws/s3.py
```diff
@@ -181,6 +181,21 @@ def object_metadata(obj: str) -> Dict[str, str]:
...
+    :return True if remote and expected version match, else False
+    """
+    try:
+        remote_version = object_metadata(obj).get("lamp_version", "")
+
+        return remote_version == version
+
+    except Exception:
+        return False
```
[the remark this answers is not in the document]
mzappitello: i'll take this function out and run it in the file itself.

i do think its a good idea to consolidate this behavior at some point.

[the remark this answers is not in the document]
mzappitello: i'll change it to only return False on 404 errors.

src/lamp_py/performance_manager/alerts.py
```diff
@@ -48,36 +50,18 @@ class AlertsS3Info:
...
-                pyarrow.timestamp("ms", tz="America/New_York"),
-            ),
-            ("active_period.start_timestamp", pyarrow.uint64()),
-            (
-                "active_period.end_datetime",
-                pyarrow.timestamp("ms", tz="America/New_York"),
-            ),
-            ("active_period.end_timestamp", pyarrow.uint64()),
+            ("created_datetime", pyarrow.timestamp("us")),
+            ("created_timestamp", pyarrow.int64()),
```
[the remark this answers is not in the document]
mzappitello: i was keeping them consistent with the timestamp fields in the `LAMP_ALL_RT_fields.parquet` file.

rymarczy (approved):
:cake:

PR #366, opened 2024-06-05
FIX dataset.to_table casting ints as floats

mzappitello:
When an alerts dataset was being filtered with Fields that compared int64 column values against python float values, these columns were cast as float32 values, which resulted in overflow errors on timestamp fields that need the wider band of values. Convert the compare values to ints before using them to filter the dataset.

Some details on the issue can be found here:
https://github.com/apache/arrow/issues/36719


rymarczy (approved):
:cake:

PR #367, opened 2024-06-05
FIX: read_parquet dataset reading not adding hive partition columns

rymarczy:
There appears to have been a change in how pyarrow dataset's read file paths with hive partitioning values. 

static table ingestion has been failing because partition columns are no longer being added to final dataframes coming out of the `read_parquet` function.


mzappitello (approved):
🍰

PR #368, opened 2024-06-06
FIX: remove timestamp tz type from hyper type map

mzappitello:
I added this type when the alerts data had timezoned datetime fields. In talking to Heather about downstream user preferences, we decided to drop the timezone part of the datetimes and ensure they were formatted to Boston time.

This type was still being used when converting the types however, causing the alerts hyper job to fail on a mismatched type. Dropping it as its no longer needed and was causing issues.


rymarczy (approved):
:cake:

PR #374, opened 2024-06-10
FEAT: Dynamic port for TM Ingestion MSSQL Connection

rymarczy:
MSSQL Port was previously hard-coded. This change makes the port dynamic.


mzappitello (approved):
🍰

PR #375, opened 2024-06-10
UPDATE: Change partition key in alerts data

mzappitello:
Initially, the processed alerts archive was partitioned into row groups via the 'last_modified_timestamp' key. After hearing back from Transit Matters, change this key to the 'active_period.start_timestamp', which should improve the types of queries they run.


rymarczy (approved):
I think the `active_period.start_timestamp` field has a lot of records that go beyond the `now` date, so this could create one very large partition at the end of the parquet file, but that can be addressed by another PR if needed.

PR #377, opened 2024-06-12
FEAT: De-Duplicate GTFS-RT Ingestion Files

rymarczy:
This change de-duplicates records for GTFS-RT parquet files. 

To achieve de-duplication, the polars library is used in combination with a record-hash function. 

This change also introduces a GTFS-RT parquet file partition column, which will partition parquet files by a specific gtfs-rt field. 

Due to the compression performance of the parquet file format, this change is not guaranteed to reduce parquet file sizes on the scale of the 1 hour partitions we are currently writing. However, the opportunity to reduce the size of parquet files with larger partition spans is available. 

Asana Task: https://app.asana.com/0/1205827492903547/1207039261301272


src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -41,14 +60,10 @@ class TableData:
...
     files: list of files that make up tables
-    next_hr_cnt: keeps track of how many files in the next hour have been
-                 processed, when this hits a certain threshold the table
-                 can be yielded
     """
 
-    table: Optional[pyarrow.table] = None
+    table: Optional[pyarrow.Table] = None
     files: List[str] = field(default_factory=list)
-    next_hr_cnt: int = 0
```
rymarczy, line 51: No longer using `next_hr_cnt` to determine when to yield a table. Table will be yielded any time it goes over 1 million records. This is conservative yielding delimiter and could be increased if needed.

src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -85,16 +100,15 @@ def __init__(
...
-        self.table_groups: Dict[datetime, TableData] = {}
+        self.data_parts: Dict[datetime, TableData] = {}
 
         self.error_files: List[str] = []
         self.archive_files: List[str] = []
 
     def convert(self) -> None:
-        max_tables_to_convert = 6
+        max_tables_to_convert = 50
         process_logger = ProcessLogger(
```
rymarczy, line 112: will convert up to 50 million rows in one convert cycle 50 * 1_000_000 yield limit

src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -174,78 +185,55 @@ def process_files(self) -> Iterable[pyarrow.table]:
...
+                    )
+                else:
+                    self.data_parts[dt_part].table = pyarrow.concat_tables(
+                        [
+                            self.data_parts[dt_part].table,
+                            self.detail.transform_for_write(rt_data),
+                        ]
+                    )
+
+                yield from self.yield_check(process_logger)
```
rymarczy, line 209: reduced logic here by dropping `next_hr_cnt`

src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -174,78 +185,55 @@ def process_files(self) -> Iterable[pyarrow.table]:
...
-            if self.table_groups[iter_ts].next_hr_cnt > yield_threshold:
-                self.archive_files += self.table_groups[iter_ts].files
-
-                table = self.table_groups[iter_ts].table
-
-                assert table is not None
+        for iter_ts in list(self.data_parts.keys()):
+            table = self.data_parts[iter_ts].table
+            if table is not None and table.num_rows > min_rows:
+                self.archive_files += self.data_parts[iter_ts].files
```
rymarczy, line 233: no more `next_hr_count` yield table if not None and over `min_rows` 1_000_000 by default, reduced to 0 when all files are processed.

src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -397,33 +385,76 @@ def sync_with_s3(self, local_path: str) -> bool:
...
+        out_ds = pd.dataset(table)
+
+        if self.sync_with_s3(local_path) is True:
+            hash_gtfs_rt_parquet(local_path)
+            out_ds = pd.dataset(
+                [
+                    pd.dataset(table),
+                    pd.dataset(local_path),
+                ]
+            )
```
rymarczy, line 406: if S3 file for partition exists, merge new table with existing data

src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -397,33 +385,76 @@ def sync_with_s3(self, local_path: str) -> bool:
...
+                for part in partitions:
+                    part_filter = pc.field(part_col) == part
+                    table = pl.from_arrow(out_ds.to_table(filter=part_filter))
+                    writer.write_table(
+                        table.sort(
+                            by=["feed_timestamp"],
+                        )
+                        .unique(subset=GTFS_RT_HASH_COL, keep="first")
+                        .to_arrow()
+                        .cast(out_ds.schema)
```
rymarczy, line 424: de-dupe operation happens here for each individual partition, doing this by partition will limit memory usage during de-duper operation.

There are most likely way to optimize/speed this up when parquet file sizes start to increase

src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -397,33 +385,76 @@ def sync_with_s3(self, local_path: str) -> bool:
...
+            with pq.ParquetWriter(tmp_pq, schema=write_schema) as writer:
+                for part in pc.unique(
+                    ds.to_table(columns=[part_col]).column(part_col)
+                ):
+                    writer.write_table(
+                        ds.to_table(
+                            columns=write_columns,
+                            filter=(pc.field(part_col) == part),
+                        ).sort_by(self.detail.table_sort_order)
+                    )
```
rymarczy, line 454: drop hash column from dataset before uploading to S3, the hash column adds a decent amount to the parquet file size and external users should not need this column

src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -449,19 +480,9 @@ def continuous_pq_update(self, table: pyarrow.Table) -> None:
...
-            else:
-                self.merge_local_pq(table, local_path)
+            self.write_local_pq(table, local_path)
 
-            upload_path = local_path.replace(
-                self.tmp_folder, os.environ["SPRINGBOARD_BUCKET"]
-            )
-            upload_file(local_path, upload_path)
-            self.send_metadata(upload_path)
+            self.upload_local_pq(local_path)
```
rymarczy, line 485: simplify this operation

src/lamp_py/ingestion/glides.py
```diff
@@ -368,17 +368,17 @@ def ingest_glides_events(
...
+    try:
+        converters = [
+            EditorChanges(),
+            OperatorSignIns(),
+            TripUpdates(),
+        ]
 
-    for record in kinesis_reader.get_records():
-        try:
+        for record in kinesis_reader.get_records():
```
rymarczy, line 381: wrapped in try-except so it doesn't kill other ingestion operations

src/lamp_py/ingestion/ingest_gtfs.py
rymarczy, line 1: This is mostly a file re-name. To indicate we're ingesting GTFS Schedule (not from S3) and GTFS-RT data.

src/lamp_py/ingestion/pipeline.py
```diff
@@ -38,21 +37,19 @@ def main() -> None:
...
-            kinesis_reader=glides_reader, metadata_queue=metadata_queue
-        )
+        ingest_gtfs(metadata_queue)
+        ingest_glides_events(glides_reader, metadata_queue)
         check_for_sigterm(metadata_queue, rds_process)
 
         process_logger.log_complete()
 
-        time.sleep(5)
+        time.sleep(30)
```
rymarczy, line 52: run every 30 seconds to avoid s3 file thrashing, to some degree

src/lamp_py/ingestion/utils.py
```diff
@@ -211,3 +217,65 @@ def explode_table_column(table: pyarrow.table, column: str) -> pyarrow.table:
         ],
         promote=True,
     )
+
+
+def hash_gtfs_rt_row(row: Any) -> Tuple[bytes]:
+    """hash row from polars dataframe"""
+    return (hashlib.md5(pickle.dumps(row), usedforsecurity=False).digest(),)
```
rymarczy, line 224: This seemed to be the most efficient hashing method for the standard library

src/lamp_py/ingestion/config_busloc_trip.py
mzappitello, line 1: wondering how you chose the partition column.

mzappitello (approved):
looks great.

src/lamp_py/ingestion/utils.py
```diff
@@ -211,3 +217,65 @@ def explode_table_column(table: pyarrow.table, column: str) -> pyarrow.table:
...
+        return
+
+    ds_columns.remove("feed_timestamp")
+    ds_columns = sorted(ds_columns)
+
+    new_schema = ds.schema.append(
+        pyarrow.field(GTFS_RT_HASH_COL, pyarrow.large_binary())
+    )
+
+    with tempfile.TemporaryDirectory() as temp_dir:
```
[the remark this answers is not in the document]
rymarczy: Yeah, we're using this often enough now that it should probably be generalized. Just need to have a think about the best way to implement that.

src/lamp_py/ingestion/config_busloc_trip.py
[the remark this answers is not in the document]
rymarczy: If `route_id` is available in a dataset that seems like the partition that will provide the most utility on these datasets. Alerts kind of just has a random partition, that file is so small it shouldn't have any significant impact

rymarczy:
@mzappitello I removed the Glides try-except changes from this PR, can be implemented in a separate change.

PR #378, opened 2024-06-12
FEAT: Update performance_manager event loop for continous ingestion

rymarczy:
Continuous ingestion will cause more thrashing in the loading of new data into the performance manager application. 

This change updates the performance manager event loop to only run Tableau parquet file updates every 30 minutes. 

This is a first attempt at an alternate event loop, more refined control of process iteration may be required depending on performance.


mzappitello (approved):
🍰 

makes sense to me.

PR #379, opened 2024-06-12
FIX: Update to Glides Ingestion before Deploy

mzappitello:
There are a few issues with glides ingestion that need to be ironed out before deploying to production.

* Operator Sign Ins had a mislabeled field that wasn't being picked up.
* Add try / except blocks and clear logging around kinesis reader and the ingest records blocks.

### DIFF
--- a/src/lamp_py/aws/kinesis.py
+++ b/src/lamp_py/aws/kinesis.py
@@ -95,36 +95,42 @@ class KinesisReader:
         )
         process_logger.log_start()
 
-        if self.shard_iterator is None:
-            self.update_shard_iterator()
-
         all_records = []
-        shard_count = 0
 
-        while True:
-            try:
-                response = self.kinesis_client.get_records(
-                    ShardIterator=self.shard_iterator
-                )
-                shard_count += 1
-                self.shard_iterator = response["NextShardIterator"]
-                records = response["Records"]
+        try:
 
-                for record in records:
-                    self.last_sequence_number = record["SequenceNumber"]
-                    all_records.append(json.loads(record["Data"]))
-
-                if response["MillisBehindLatest"] == 0:
-                    break
-
-            # thrown if the shard iterator has expired. catch it and regenerate
-            # a new shard iterator.
-            except self.kinesis_client.exceptions.ExpiredIteratorException:
+            if self.shard_iterator is None:
                 self.update_shard_iterator()
 
-        process_logger.add_metadata(
-            record_count=len(all_records), shard_count=shard_count
-        )
+            shard_count = 0
+
+            while True:
+                try:
+                    response = self.kinesis_client.get_records(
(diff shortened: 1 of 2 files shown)


src/lamp_py/aws/kinesis.py
```diff
@@ -95,36 +95,42 @@ def get_records(self) -> List[Dict]:
...
-                    all_records.append(json.loads(record["Data"]))
-
-                if response["MillisBehindLatest"] == 0:
-                    break
-
-            # thrown if the shard iterator has expired. catch it and regenerate
-            # a new shard iterator.
-            except self.kinesis_client.exceptions.ExpiredIteratorException:
+
+        try:
```
mzappitello, line 100: reading records is under a try except with logging.

src/lamp_py/ingestion/glides.py
```diff
@@ -97,29 +98,31 @@ def append_records(self) -> None:
...
-                end = start + relativedelta(months=1)
-                if end < now:
-                    table = joined_ds.filter(
-                        (pc.field("time") >= start) & (pc.field("time") < end)
-                    ).to_table()
-                else:
-                    table = joined_ds.filter(
-                        (pc.field("time") >= start)
-                    ).to_table()
+        with tempfile.TemporaryDirectory() as tmp_dir:
```
mzappitello, line 101: use tempfile here now.

src/lamp_py/ingestion/glides.py
```diff
@@ -217,7 +220,7 @@ def event_schema(self) -> pyarrow.schema:
                                     [("badgeNumber", pyarrow.string())]
                                 ),
                             ),
-                            ("signed_in_at", pyarrow.timestamp("ms")),
+                            ("signedInAt", pyarrow.timestamp("ms")),
```
mzappitello, line 223: this field was incorrectly labeled, so i wasn't picking anything up. :/

src/lamp_py/ingestion/glides.py
```diff
@@ -244,6 +247,11 @@ def unique_key(self) -> str:
         return "operator"
 
     def convert_records(self) -> pd.Dataset:
+        for record in self.records:
+            record["data"]["signedInAt"] = datetime.fromisoformat(
+                record["data"]["signedInAt"].replace("Z", "+00:00")
+            )
```
mzappitello, line 253: i have to do this if we want the record to be stored as a pyarrow.timestamp.

would also do a string if you prefer.

rymarczy: String probably makes more sense for these, doing the conversion later is pretty inconsequential for such a small dataset

rymarczy (approved):
LGTM :cake:

mzappitello:
i'll flip that to a string real quick.

rymarczy (approved):
:cake:

PR #385, opened 2024-06-25
FEAT: GTFS-RT Daily Ingest

rymarczy:
Change GTFS-RT ingestion partition from hourly to daily.

This run successfully on DEV for several days with acceptable loop duration.

TRIP_UPDATES are the limiting events for this processing. There are some potential options for optimization, if needed based on future application performance. 

Asana Task: https://app.asana.com/0/1205827492903547/1207039261301268

### DIFF
--- a/src/lamp_py/ingestion/convert_gtfs_rt.py
+++ b/src/lamp_py/ingestion/convert_gtfs_rt.py
@@ -108,7 +108,7 @@ class GtfsRtConverter(Converter):
         self.archive_files: List[str] = []
 
     def convert(self) -> None:
-        max_tables_to_convert = 50
+        max_tables_to_convert = 15
         process_logger = ProcessLogger(
             "parquet_table_creator",
             table_type="gtfs-rt",
@@ -156,7 +156,7 @@ class GtfsRtConverter(Converter):
         """
         iterate through all of the files to be converted
 
-        only yield a new table when the timestamps cross over an hour.
+        only yield a new table when table size crosses over min_rows of yield_check
         """
         max_workers = 4
 
@@ -184,7 +184,11 @@ class GtfsRtConverter(Converter):
                     continue
 
                 # create key for self.data_parts dictionary
-                dt_part = result_dt.replace(minute=0, second=0, microsecond=0)
+                dt_part = datetime(
+                    year=result_dt.year,
+                    month=result_dt.month,
+                    day=result_dt.day,
+                )
 
                 # create new self.table_groups entry for key if it doesn't exist
                 if dt_part not in self.data_parts:
@@ -213,7 +217,7 @@ class GtfsRtConverter(Converter):
         process_logger.log_complete()
 
     def yield_check(
-        self, process_logger: ProcessLogger, min_rows: int = 1_000_000
+        self, process_logger: ProcessLogger, min_rows: int = 5_000_000
     ) -> Iterable[pyarrow.table]:
         """
         yield all tables in the data_parts map that have been sufficiently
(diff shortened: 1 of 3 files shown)


src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -405,55 +402,90 @@ def write_local_pq(self, table: pyarrow.Table, local_path: str) -> None:
                 ]
```
mzappitello, line 402: just seeing this now, but the line

`if self.sync_with_s3(local_path) is True:`

block can drop the `is True` portion.

src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -494,15 +523,15 @@ def continuous_pq_update(self, table: pyarrow.Table) -> None:
                 ),
                 ignore_errors=True,
             )
-            self.error_files += self.archive_files
+            # self.error_files += self.archive_files
```
mzappitello, line 526: why are we taking this out?

src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -405,55 +402,90 @@ def write_local_pq(self, table: pyarrow.Table, local_path: str) -> None:
...
-                    )
+                        .cast(out_ds.schema),
+                        out_ds.to_table(
+                            filter=(
+                                (pc.field(self.detail.partition_column) == part)
+                                & (pc.field("feed_timestamp") < unique_ts_min)
+                            )
+                        ),
+                    ]
+                ).sort_by(self.detail.table_sort_order)
```
mzappitello, line 468: if i'm reading this right, we're doing the deduplication on everything in the out dataset that is less than 45 minutes older than the oldest piece that was ingested in this event loop?

i think it might read a little cleaner if the two tables, newer and older, were explicitly declared and then concatenated. it would take one layer of nesting out.

src/lamp_py/ingestion/convert_gtfs_rt.py
rymarczy: Incorporated

src/lamp_py/ingestion/convert_gtfs_rt.py
rymarczy: Was a holdover from testing on dev, will revert

src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -124,22 +124,19 @@ def convert(self) -> None:
...
 
-            self.clean_local_folders()
-
         except Exception as exception:
             process_logger.log_failure(exception)
         else:
             process_logger.log_complete()
+        finally:
+            self.move_s3_files()
+            self.clean_local_folders()
```
rymarczy, line 139: always move and clean folders in `convert` function

src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -186,15 +183,15 @@ def process_files(self) -> Iterable[pyarrow.table]:
...
+                    day=result_dt.day,
+                )
 
                 # create new self.table_groups entry for key if it doesn't exist
                 if dt_part not in self.data_parts:
                     self.data_parts[dt_part] = TableData()
-
-                self.data_parts[dt_part].files.append(result_filename)
-
-                if self.data_parts[dt_part].table is None:
```
rymarczy, line 197: clean up this control flow a bit

mzappitello (approved):
looks like  you have to rebase, but LGTM.

🍰

PR #386, opened 2024-06-26
FIX: [Ingestion] always clean and move files

rymarczy:
This is the only thing I can find that could be causing increasing run-times on Ingestion prod.

### DIFF
--- a/src/lamp_py/ingestion/convert_gtfs_rt.py
+++ b/src/lamp_py/ingestion/convert_gtfs_rt.py
@@ -124,22 +124,20 @@ class GtfsRtConverter(Converter):
                     continue
 
                 self.continuous_pq_update(table)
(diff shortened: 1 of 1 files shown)


mzappitello (approved):
🍰 

this looks fine to me. are we seeing the failure logs on prod though? i would think that without those this change wont do much to fix things.

its also going to effect our 'parquet_table_createor' times since the moving and cleaning are now happening after the complete and finally logs.

rymarczy:
There are these occasional error logs: https://mbta.splunkcloud.com/en-US/app/search/search?sid=1719407868.3524138

But I imagine the main issue is this conditional:
```
if table.num_rows == 0:
    continue
```
potentially blocking the moving of files or folder clean up, this would be kind of a silent blocking of these operations

PR #388, opened 2024-06-28
FEAT: Export compressed GTFS schedule to SQLITE db

rymarczy:
This change adds the ability to export compressed GTFS schedule data to an SQLITE db file. 

For each year partition folder, in the compressed gtfs archives, one SQLITE db file will be produced that contains a table for each GTFS schedule file that has been compressed.

This implementation utilizes the python built-in sqlite3 library. Locally, I am able to produce each SQLITE db file in about 2 minutes. 

This change also has all of the S3 sync and upload logic to fully implement the compressed GTFS process. 

Asana Task: https://app.asana.com/0/1205827492903547/1207450430015372

### DIFF
--- a/src/lamp_py/ingestion/compress_gtfs/gtfs_to_parquet.py
+++ b/src/lamp_py/ingestion/compress_gtfs/gtfs_to_parquet.py
@@ -10,10 +10,17 @@ import pyarrow.dataset as pd
 
 from lamp_py.runtime_utils.process_logger import ProcessLogger
 
-from .gtfs_schema_map import gtfs_schema_list
-from .gtfs_schema_map import gtfs_schema
-from .schedule_details import ScheduleDetails
-from .schedule_details import schedules_to_compress
+from lamp_py.ingestion.compress_gtfs.gtfs_schema_map import (
+    gtfs_schema_list,
+    gtfs_schema,
+)
+from lamp_py.ingestion.compress_gtfs.schedule_details import (
+    ScheduleDetails,
+    schedules_to_compress,
+    GTFS_PATH,
+)
+from lamp_py.ingestion.compress_gtfs.pq_to_sqlite import pq_folder_to_sqlite
+from lamp_py.aws.s3 import upload_file
 
 
 def frame_parquet_diffs(
@@ -57,6 +64,7 @@ def frame_parquet_diffs(
         how="anti",
         on=join_columns,
         join_nulls=True,
+        coalesce=True,
     ).drop("from_zip")
 
     # left join to create frame of old and same records
@@ -65,6 +73,7 @@ def frame_parquet_diffs(
         how="left",
         on=join_columns,
         join_nulls=True,
+        coalesce=True,
     )
     same_records = pq_frame.filter(pl.col("from_zip").eq(True)).drop("from_zip")
     old_records = pq_frame.filter(pl.col("from_zip").is_null()).drop("from_zip")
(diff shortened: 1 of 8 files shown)


src/lamp_py/ingestion/compress_gtfs/gtfs_to_parquet.py
```diff
@@ -73,51 +82,46 @@ def frame_parquet_diffs(
...
             for batch in filter_ds.to_batches(batch_size=batch_size):
                 writer.write_batch(batch)
 
         # over-write export_path file with merged dataset
-        combined_ds = pd.dataset((filter_pq_path, new_pq_path))
-        with pq.ParquetWriter(export_path, schema=pq_schema) as writer:
-            for batch in combined_ds.to_batches(batch_size=batch_size):
+        export_ds = pd.dataset((pd.dataset(tmp_path), pd.dataset(merge_df)))
+        with pq.ParquetWriter(export_path, schema=merge_df.schema) as writer:
+            for batch in export_ds.to_batches(batch_size=batch_size):
```
mzappitello, line 124: i'm guessing these datasets arent big enough to warrant partitioning them into row groups?

src/lamp_py/ingestion/compress_gtfs/gtfs_to_parquet.py
```diff
@@ -313,8 +319,10 @@ def gtfs_to_parquet() -> None:
...
     for year in set(feed["published_dt"].dt.strftime("%Y").unique()):
         year_path = os.path.join(gtfs_tmp_folder, year)
-        for _ in os.listdir(year_path):
-            # upload file to S3
-            continue
+        pq_folder_to_sqlite(year_path)
+        for file in os.listdir(year_path):
+            local_path = os.path.join(year_path, file)
+            upload_path = os.path.join(GTFS_PATH, year, file)
+            upload_file(local_path, upload_path)
```
mzappitello, line 326: wooohoo

src/lamp_py/ingestion/compress_gtfs/pq_to_sqlite.py
```diff
@@ -0,0 +1,82 @@
...
+    if "double" in pq_type:
+        return "REAL"
+    return "TEXT"
+
+
+def sqlite_table_query(table_name: str, schema: pyarrow.Schema) -> str:
+    """
+    return CREATE TABLE query for sqlite table from pyarrow schema
+    """
+    logger = ProcessLogger("sqlite_create_table")
```
mzappitello, line 30: seems like we're logging how long it takes to generate the query here rather than how long it takes to create and populate the table which might be more useful.

mzappitello (approved):
one comment on the sqlite generation for you to take or leave.

lgtm.

src/lamp_py/ingestion/compress_gtfs/gtfs_to_parquet.py
```diff
@@ -57,6 +64,7 @@ def frame_parquet_diffs(
         how="anti",
         on=join_columns,
         join_nulls=True,
+        coalesce=True,
```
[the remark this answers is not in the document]
rymarczy: It was a recent change to the library, I think the PR talks about matching the behavior of all join operations: https://github.com/pola-rs/polars/issues/13441

src/lamp_py/ingestion/compress_gtfs/gtfs_to_parquet.py
rymarczy: `stop_times` and maybe the `trips` table could benefit from that, but I think the most sensible partitioning would be quarterly (every 3 months), and the added complexity of that logic didn't currently seem worth the rub

src/lamp_py/ingestion/compress_gtfs/pq_to_sqlite.py
rymarczy: Dropped the logging of the full `CREATE TABLE` query. 

Should be able to track the duration of SQLITE table creation/insertion when `pq_folder_to_sqlite`  calls `logger.add_metadata(current_file=file)`

rymarczy:
Will wait to merge this until https://github.com/mbta/devops/pull/1947 is applied.

PR #392, opened 2024-07-01
Feat: Add Tableau Project env var

mzappitello:
We're now going to be publishing to tableau from staging in addition to prod. The Tableau Project is currently hardcoded to aim at the prod location. Create and use an environmental variable that stores the appropriate project for each instance of the publisher application.

Asana Task: [🚀 Stand up staging Publisher](https://app.asana.com/0/1205827492903547/1206391938568280/f)

### DIFF
--- a/.github/actions/run_task/action.yaml
+++ b/.github/actions/run_task/action.yaml
@@ -1,0 +1,34 @@
+name: Run LAMP ECS Task in Environment
+description: Run an existing task in the LAMP ECS Cluster
+
+inputs:
+  role-to-assume:
+    description: IAM role
+    required: true
+  aws-region:
+    description: AWS region to use
+    required: true
+    default: us-east-1
+  environment:
+    description: prod or staging
+    required: true
+  task_name:
+    description: name of the task to run in kebab case
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Setup AWS Credentials
+      uses: aws-actions/configure-aws-credentials@v1
+      with:
+        role-to-assume: ${{ inputs.role-to-assume }}
+        aws-region: ${{ inputs.aws-region }}
+    - name: Run ECS Task
+      id: run-ecs
+      run: ${{ github.action_path }}/run_task.sh
+      shell: bash
+      env:
+        AWS_REGION: ${{ inputs.aws-region }}
+        TASK_NAME: ${{ inputs.task_name }}
+        ENVIRONMENT: ${{ inputs.environment }}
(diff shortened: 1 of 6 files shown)


rymarczy (approved):
:cake: Just need to make sure we get the devops PR merged for PROD before deploying this.

.github/actions/run_task/action.yaml
mzappitello, line 1: i created an action with a bash script to run the task.

when launching from the command line, you need to define the subnet and security group to be used. the bash script pulls both of them based on the task name and envirionment.

.github/workflows/run-task.yaml
mzappitello, line 1: change this workflow to use the action.

src/lamp_py/runtime_utils/env_validation.py
```diff
@@ -45,9 +45,10 @@ def validate_environment(
         value = os.environ.get(key, None)
         if value is None:
             missing_required.append(key)
-        # do not log private variables
-        if key in private_variables:
-            value = "**********"
+        else:
+            # do not log private variables
+            if key in private_variables:
+                value = "**********"
```
mzappitello, line 51: we were only logging private vars before 👎 . log them all now.

src/lamp_py/tableau/hyper.py
```diff
@@ -42,7 +42,7 @@ def __init__(
         hyper_file_name: str,
         remote_parquet_path: str,
         lamp_version: str,
-        project_name: str = "LAMP API",
+        project_name: str = os.getenv("TABLEAU_PROJECT", ""),
```
mzappitello, line 45: project name is an env var now.

src/lamp_py/tableau/pipeline.py
```diff
@@ -30,6 +30,7 @@ def start_hyper_updates() -> None:
             "TABLEAU_USER",
             "TABLEAU_PASSWORD",
             "TABLEAU_SERVER",
+            "TABLEAU_PROJECT",
```
mzappitello, line 33: make sure we have a project name.

.github/actions/run_task/action.yaml
mzappitello: i'm still getting permissions errors when i run this. something about the way our permissions are structured for describing the vpcs is still giving us unauthed errors.

src/lamp_py/runtime_utils/env_validation.py
rymarczy: I don't think this needs to be in an `else` block

the `metadata[key] = value` line below just needs to be un-indented one level. 

This was introduced here: https://github.com/mbta/lamp/commit/36b959265ff977425013732019088fe46082d827

when the `add_metadata` calls were removed.

rymarczy (approved):
One, suggestion. Otherwise, LGTM.

rymarczy (approved):
:cake:

PR #394, opened 2024-07-03
FIX: Do not calculate dwell_time for first/last stop of trip

rymarczy:
Bug ticket submitted by Ops Analytics requested change to calculations of dwell times. This change explicitly does not calculate dwell times for first and last stops of trips (which would be considered a turn-around time). 

Also includes migration to update existing dwell times in RDS.  

Asana Task: https://app.asana.com/0/1205827492903547/1207437794217284


src/lamp_py/performance_manager/l1_cte_statements.py
```diff
@@ -231,16 +231,6 @@ def trips_for_metrics_subquery(
...
-            sa.func.coalesce(
-                static_trips_sub.c.static_trip_stop_rank,
-                rt_trips_sub.c.rt_trip_stop_rank,
-            ).label("stop_rank"),
-            sa.func.lead(rt_trips_sub.c.vp_move_timestamp)
-            .over(
-                partition_by=rt_trips_sub.c.vehicle_id,
-                order_by=rt_trips_sub.c.vp_move_timestamp,
-            )
-            .label("next_station_move"),
```
rymarczy, line 243: These subquery fields not being used

src/lamp_py/performance_manager/l1_rt_metrics.py
```diff
@@ -74,43 +74,19 @@ def update_metrics_columns(
...
-                ),
-                else_=sa.literal(None),
+            (
+                sa.func.lead(
+                    trips_for_metrics.c.move_timestamp,
+                ).over(
+                    partition_by=trips_for_metrics.c.pm_trip_id,
+                    order_by=trips_for_metrics.c.sort_timestamp,
+                )
+                - trips_for_metrics.c.stop_timestamp
```
mzappitello, line 84: how does this work for the last stop on the stop on the trip? i guess the move timestamp won't get brought up because of the `lead` call, leaving this value unset?

src/lamp_py/migrations/versions/performance_manager_prod/004_896dedd8a4db_dwell_time_update.py
```diff
@@ -0,0 +1,192 @@
...
+
+def upgrade() -> None:
+    date_query = sa.text(
+        "SELECT DISTINCT service_date, static_version_key FROM vehicle_trips ORDER BY service_date"
+    )
+
+    conn = op.get_bind()
+    result = conn.execute(date_query)
+    for service_date, static_version_key in result.fetchall():
+        stop_sync_update = f"""
```
mzappitello, line 37: this update looks like its recalculating the dwell times for each vehicle event? i guess we haven't deployed to dev or staging yet, but i would think that its going to take a while?

PR #397, opened 2024-07-09
Update Glides Parquet Files

mzappitello:
In the initial deployment of glides ingestion, location fields that include the gtfsId and todsId fields incorrectly used ID instead of Id. Correct this in the schemas for each of our record types and create a migration that will modify existing data to the new schema so that old and new datasets can be joined correctly.

While investigating, I also noticed that there were duplicate records in our files stemming from restarting the application and re-ingesting data from the past 24 hours. In the migration and in the GlidesConverter class, deduplicate the datasets based on the 'id' column.

Asana Task: [🪂 Migrate Glides Files Schema](https://app.asana.com/0/1205827492903547/1207770444568210/f)

### DIFF
--- a/src/lamp_py/ingestion/glides.py
+++ b/src/lamp_py/ingestion/glides.py
@@ -1,6 +1,6 @@
 from typing import Dict, List, Optional
 import os
-from datetime import datetime, timedelta
+from datetime import datetime
 import tempfile
 from queue import Queue
 
@@ -99,8 +99,6 @@ class GlidesConverter(ABC):
         now = datetime.now()
         start = datetime(2024, 1, 1)
 
-        unique_ts_min = now - timedelta(hours=36)
-
         with tempfile.TemporaryDirectory() as tmp_dir:
 
             new_path = os.path.join(tmp_dir, self.base_filename)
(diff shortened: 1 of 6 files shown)


src/lamp_py/ingestion/glides.py
```diff
@@ -31,8 +32,8 @@ class GlidesConverter(ABC):
 
     glides_location = pyarrow.struct(
         [
-            ("gtfsID", pyarrow.string()),
-            ("todsID", pyarrow.string()),
+            ("gtfsId", pyarrow.string()),
+            ("todsId", pyarrow.string()),
```
mzappitello, line 36: 🤦

src/lamp_py/ingestion/glides.py
```diff
@@ -88,7 +89,7 @@ def append_records(self) -> None:
 
         joined_ds = new_dataset
         if os.path.exists(self.local_path):
-            joined_ds = pd.dataset([pd.dataset(self.local_path), new_dataset])
+            joined_ds = pd.dataset([new_dataset, pd.dataset(self.local_path)])
```
mzappitello, line 92: swapped the order here so that "newer" data will be the one thats used when i do the unique later on.

src/lamp_py/ingestion/glides.py
```diff
@@ -98,6 +99,8 @@ def append_records(self) -> None:
         now = datetime.now()
         start = datetime(2024, 1, 1)
 
+        unique_ts_min = now - timedelta(hours=36)
+
```
mzappitello, line 103: used for uniquing the datasets. the kinesis stream holds data for 24 hours, so we'll run the unique on any data older than 36 hours to be safe.

rymarczy (approved):
Couple of suggestions, but otherwise LGTM

rymarczy (approved):
:cake:

PR #398, opened 2024-07-10
FIX (LAMP_ALL_RT_fields): Use Static Schedule to filter Initial Station of Trip

rymarczy:
This change drops using `canonical_stop_sequence` for filtering out initial trip stations and instead uses static schedule data to drop any events relating to the initial canonical station of a trip.

Asana Task: https://app.asana.com/0/1205827492903547/1207100677882489


src/lamp_py/tableau/jobs/rt_rail.py
```diff
@@ -73,6 +73,63 @@ def __init__(self) -> None:
...
+            "               coalesce(static_trips.branch_route_id, static_trips.trunk_route_id), "
+            "               static_route_patterns.direction_id, "
+            "               static_route_patterns.static_version_key, "
+            "               static_route_patterns.route_pattern_typicality desc) as canon_trips "
+            "       JOIN static_stop_times on "
+            "           canon_trips.representative_trip_id = static_stop_times.trip_id "
+            "           AND canon_trips.static_version_key = static_stop_times.static_version_key "
+            "       JOIN static_stops on "
+            "           static_stop_times.stop_id = static_stops.stop_id "
+            "           AND static_stop_times.static_version_key = static_stops.static_version_key ) as drop_station "
```
mzappitello, line 123: sanity check for my sake. 

this gets every station on every trip, then its filtered out to only be the first station for every trip.

we use that for the filtering later rather than the `ve.canonical_stop_sequence`

src/lamp_py/tableau/jobs/rt_rail.py
```diff
@@ -97,17 +154,14 @@ def __init__(self) -> None:
...
-            "   )"
+            "   AND drop_join.drop_flag IS NULL "
             "   AND ("
             "       ve.vp_stop_timestamp IS NOT null"
             "       OR ve.vp_move_timestamp IS NOT null"
             "   )"
             "   %s"
             " ORDER BY "
-            "   ve.service_date, vt.route_id, vt.direction_id, vt.vehicle_id, vt.start_time"
+            "   ve.service_date, vt.vehicle_id, vt.start_time"
```
mzappitello, line 164: was the order change requested?

mzappitello (approved):
a few small questions.

src/lamp_py/tableau/jobs/rt_rail.py
rymarczy: Yes, this subquery is the same one used to create `canonical_stop_sequence` but we are limiting it to the first `parent_station` in each each direction and then doing an anti-join with the `drop_flag` field.

src/lamp_py/tableau/jobs/rt_rail.py
rymarczy: It wasn't requested, but we're only doing this to reduce the parquet file size and dropping the two SORT field fields gave us a speed up in query time with almost no change in file size.

src/lamp_py/tableau/jobs/rt_rail.py
```diff
@@ -255,7 +255,7 @@ def update_parquet(self, db_manager: DatabaseManager) -> bool:
         max_start_date -= datetime.timedelta(days=1)
 
         update_query = self.table_query % (
-            f" AND vt.service_date >= {max_start_date.strftime('%Y%m%d')} ",
+            f" AND ve.service_date >= {max_start_date.strftime('%Y%m%d')} ",
```
mzappitello, line 258: faster because the base of the query is on vehicle events?

mzappitello (approved):
🍰

PR #401, opened 2024-07-15
FIX: Retry logic for `write_to_parquet` method of `MSSQLManager`

rymarczy:
TM Ingestion `write_to_parquet` operations have been failing with the following error:

```
pyodbc.Error: ('40001', '[40001] [Microsoft][ODBC Driver 18 for SQL Server][SQL Server]Transaction (Process ID 63) was deadlocked on lock resources with another process and has been chosen as the deadlock victim. Rerun the transaction. (1205) (SQLFetch)')
```

This change adds retry logic to the `write_to_parquet` method of the `MSSQLManager` class in an effort to avoid the failure of these TM Ingestion table export jobs. 

Asana Task: https://app.asana.com/0/1205827492903547/1207692276182265

### DIFF
--- a/src/lamp_py/mssql/mssql_utils.py
+++ b/src/lamp_py/mssql/mssql_utils.py
@@ -120,8 +120,6 @@ class MSSQLManager:
         with self.session.begin() as cursor:
             return [row._asdict() for row in cursor.execute(select_query)]
 
-    # pylint: disable=R0801
-    # Similar lines in 2 files
     def write_to_parquet(
         self,
         select_query: sa.sql.selectable.Select,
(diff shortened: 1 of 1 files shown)


src/lamp_py/mssql/mssql_utils.py
```diff
@@ -161,14 +159,26 @@ def write_to_parquet(
...
+                                    [row._asdict() for row in part],
+                                    schema=schema,
+                                )
+                            )
+
+                process_logger.log_complete()
+
+            except Exception as exception:
+                if retry_attempts == max_retries:
+                    process_logger.log_failure(exception)
```
mzappitello, line 184: no reason to suspect that the errors on successive retries, right? if they could happen it might be worth logging the error type each time?

mzappitello: And do we want to have a backoff with each try before rolling right into the next session?

mzappitello (commented):
two questions.

src/lamp_py/mssql/mssql_utils.py
rymarczy: We've only seen 1 error type, thus far... The ProcessLogger class doesn't currently support logging errors that don't end a process, so that logging would be occurring outside of the process_logger. 

For the error we've been seeing, I don't think a backoff would be beneficial, but that could change. The current error is the DB killing the query and we can't really know whether that will or will not happen or if a backoff would help.

src/lamp_py/mssql/mssql_utils.py
```diff
@@ -161,14 +159,27 @@ def write_to_parquet(
...
+                        ):
+                            pq_writer.write_batch(
+                                pyarrow.RecordBatch.from_pylist(
+                                    [row._asdict() for row in part],
+                                    schema=schema,
+                                )
+                            )
+
+                process_logger.log_complete()
+                break
```
rymarczy, line 181: forgot this break

src/lamp_py/mssql/mssql_utils.py
mzappitello: i think you can just call `log_failure` multiple times. that does have the effect of spamming our alerts though. probably a good future feature for our process logger.

mzappitello (approved):
🍰

PR #403, opened 2024-07-16
Fix Glides Migration

mzappitello:
In the last attempt, there was an error converting from polars back to arrow after removing duplicate data. The schema would become slightly modified, throwing errors when trying to blend it with the existing schema. Now run a cast after converting back to arrow.

### DIFF
--- a/src/lamp_py/ingestion/glides.py
+++ b/src/lamp_py/ingestion/glides.py
@@ -107,30 +107,27 @@ class GlidesConverter(ABC):
                 while start < now:
                     end = start + relativedelta(months=1)
                     if end < now:
-                        unique_table = (
-                            pl.DataFrame(
-                                joined_ds.filter(
-                                    (pc.field("time") >= start)
-                                    & (pc.field("time") < end)
-                                ).to_table()
-                            )
-                            .unique(keep="first")
-                            .to_arrow()
-                            .sort_by("time")
-                        )
-                    else:
-                        unique_table = (
-                            pl.DataFrame(
-                                joined_ds.filter(
-                                    (pc.field("time") >= start)
-                                ).to_table()
-                            )
-                            .unique(keep="first")
-                            .to_arrow()
-                            .sort_by("time")
+                        row_group = pl.DataFrame(
+                            joined_ds.filter(
+                                (pc.field("time") >= start)
+                                & (pc.field("time") < end)
+                            ).to_table()
                         )
 
-                    if unique_table.num_rows() > 0:
+                    else:
+                        row_group = pl.DataFrame(
+                            joined_ds.filter(
+                                (pc.field("time") >= start)
+                            ).to_table()
+                        )
+
+                    if not row_group.is_empty():
+                        unique_table = (
+                            row_group.unique(keep="first")
+                            .sort(by=["time"])
+                            .to_arrow()
+                            .cast(new_dataset.schema)
+                        )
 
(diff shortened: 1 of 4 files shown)


src/lamp_py/ingestion/glides.py
```diff
@@ -107,31 +107,28 @@ def append_records(self) -> None:
...
-                                joined_ds.filter(
-                                    (pc.field("time") >= start)
-                                ).to_table()
-                            )
-                            .unique(keep="first")
+                            row_group.unique(keep="first")
+                            .sort(by=["time"])
                             .to_arrow()
-                            .sort_by("time")
+                            .cast(new_dataset.schema)
```
mzappitello, line 129: i was missing this cast before, which caused incompatible schema errors. (it seems `to_arrow()` makes all strings large strings.

src/lamp_py/ingestion/glides.py
```diff
@@ -107,31 +107,28 @@ def append_records(self) -> None:
                 while start < now:
                     end = start + relativedelta(months=1)
                     if end < now:
-                        unique_table = (
```
mzappitello, line 110: a lot of the same actions were taking place in either of the `if end < now` condition, so i collapsed them.

src/lamp_py/migrations/versions/metadata_dev/002_26db393ea854_update_glides_location_column_names.py
```diff
@@ -43,21 +45,36 @@ def update_glides_archive(temp_dir: str, base_filename: str) -> None:
...
 
-        old = pq.read_table(old_local_path)
-        old_names = old.column_names
-
-        # rename columns containing gtfsID and todsID to use Id instead
-        new_names = [
-            n.replace("gtfsID", "gtfsId").replace("todsID", "todsId")
-            for n in old_names
-        ]
-        renamed = old.rename_columns(new_names)
```
mzappitello, line 54: in addition to renaming, i'll also need a schema to cast against or i'll be left with whatever pyarrow converts to.

generate a list of fields and generate a new schema from that.

src/lamp_py/migrations/versions/metadata_dev/002_26db393ea854_update_glides_location_column_names.py
```diff
@@ -43,21 +45,36 @@ def update_glides_archive(temp_dir: str, base_filename: str) -> None:
...
-        pq.write_table(new, new_local_path)
-
+        # cast to new schema (polars converts things)
+        new_table = (
+            pl.DataFrame(old_table.rename_columns(schema.names))
+            .unique()
+            .sort(by=["time"])
+            .to_arrow()
+            .cast(schema)
+        )
```
mzappitello, line 75: all the renaming, uniquing, and casting happens here.

src/lamp_py/ingestion/glides.py
```diff
@@ -107,31 +107,28 @@ def append_records(self) -> None:
...
                         )
+
                     else:
+                        row_group = pl.DataFrame(
+                            joined_ds.filter(
+                                (pc.field("time") >= start)
+                            ).to_table()
+                        )
+
+                    if row_group.is_empty() > 0:
```
rymarczy, line 124: I believe `is_empty` returns a bool

rymarczy (changes requested):
Just 1 change...

PR #405, opened 2024-07-18
Add Function to Choose Files for BPM Ingestion

mzappitello:
Add Function to Choose Files for BPM Ingestion 

Instead of using the metadata db, we plan on using last modified
timestamps on incoming buckets and comparing them with the last modified
timestamps on our outgoing vehicle event files.

* Create two new S3 Utilities
  * one gets the last modified file object in an s3 bucket / prefix.
  * one gets all of the files in a s3 bucket / prefix that have been
    modified after a given time.
* Create `remote_files.py` which will house the s3 location information
  for various files LAMP uses.
* Create `bus_performance_manager/vehicle_events.py` and create the
  function for collecting files that have been modified since the last
  update and a function that will group those files based on service
  date.

Asana Task: [🚌 Create function for choosing files for ingestion](https://github.com/mbta/lamp/commit/13851b8dfb5ac3338be37022cb2ca1ee28055394)

### DIFF
--- a/src/lamp_py/aws/s3.py
+++ b/src/lamp_py/aws/s3.py
@@ -1,8 +1,8 @@
-import datetime
 import os
 import re
 import time
 from concurrent.futures import ThreadPoolExecutor
+from datetime import datetime, timezone
 from io import BytesIO
 from threading import current_thread
 from typing import (
@@ -279,7 +279,7 @@ def file_list_from_s3_with_details(
     return_dict = {
         "s3_obj_path": "str: object path as s3://bucket-name/object-key",
         "size_bytes": "int: size of object in bytes",
-        "last_modified": "datetime.datetime: object creation date",
+        "last_modified": "datetime: object creation date",
     }
 
     :return List[return_dict]
@@ -319,6 +319,60 @@ def file_list_from_s3_with_details(
     except Exception as exception:
         process_logger.log_failure(exception)
         return []
+
+
+def file_list_after(
+    bucket_name: str, file_prefix: str, cutoff: datetime
+) -> List[str]:
+    """
+    List all of the files in a bucket matching a prefix that were modified
+    after a cutoff datetime.
+    """
+    s3_client = get_s3_client()
+    paginator = s3_client.get_paginator("list_objects_v2")
+    pages = paginator.paginate(Bucket=bucket_name, Prefix=file_prefix)
+
+    filepaths = []
+    for page in pages:
+        if page["KeyCount"] == 0:
+            continue
+        for obj in page["Contents"]:
+            if obj["Size"] == 0:
+                continue
+
+            if obj["LastModified"] >= cutoff:
+                filepaths.append(os.path.join("s3://", bucket_name, obj["Key"]))
+    return filepaths
+
+
+def get_last_modified_object(
+    bucket_name: str, file_prefix: str, version: Optional[str] = None
+) -> Optional[Dict]:
+    """
+    For a given bucket, find the last modified object that matches a prefix.
+    If a version is passed, only return the object if the versions of files
+    matching this prefix match.
+    """
+    objects = file_list_from_s3_with_details(
+        bucket_name=bucket_name, file_prefix=file_prefix
+    )
+
+    # if there are no objects, return early.
+    if len(objects) == 0:
+        return None
+
+    # start with the first object
+    newest_object = objects[0]
+
+    # if the first version fails that version check, return early
+    if version and not version_check(newest_object["s3_obj_path"], version):
+        return None
+
+    for obj in objects:
+        if obj["last_modified"] > newest_object["last_modified"]:
+            newest_object = obj
+
+    return newest_object
 
 
 def _move_s3_object(filename: str, to_bucket: str) -> Optional[str]:
@@ -557,7 +611,7 @@ def write_parquet_file(
 # pylint: enable=R0913
 
 
-def get_datetime_from_partition_path(path: str) -> datetime.datetime:
+def get_datetime_from_partition_path(path: str) -> datetime:
     """
     process and return datetime from partitioned s3 path
     """
@@ -569,17 +623,17 @@ def get_datetime_from_partition_path(path: str) -> datetime.datetime:
         hour = 0
         if "hour=" in path:
             hour = int(re.findall(r"hour=(\d{1,2})", path)[0])
-        return_date = datetime.datetime(
+        return_date = datetime(
             year=year,
             month=month,
             day=day,
             hour=hour,
-            tzinfo=datetime.timezone.utc,
+            tzinfo=timezone.utc,
         )
     except IndexError as _:
         # handle gtfs static paths
         timestamp = float(re.findall(r"timestamp=(\d{10})", path)[0])
-        return_date = datetime.datetime.fromtimestamp(timestamp)
+        return_date = datetime.fromtimestamp(timestamp)
     return return_date
 
 
--- a/src/lamp_py/aws/s3.py
+++ b/src/lamp_py/aws/s3.py
@@ -321,58 +321,35 @@ def file_list_from_s3_with_details(
         return []
 
 
-def file_list_after(
-    bucket_name: str, file_prefix: str, cutoff: datetime
-) -> List[str]:
-    """
-    List all of the files in a bucket matching a prefix that were modified
-    after a cutoff datetime.
-    """
-    s3_client = get_s3_client()
-    paginator = s3_client.get_paginator("list_objects_v2")
-    pages = paginator.paginate(Bucket=bucket_name, Prefix=file_prefix)
-
-    filepaths = []
-    for page in pages:
-        if page["KeyCount"] == 0:
-            continue
-        for obj in page["Contents"]:
-            if obj["Size"] == 0:
-                continue
-
-            if obj["LastModified"] >= cutoff:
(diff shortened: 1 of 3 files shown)


src/lamp_py/aws/s3.py
```diff
@@ -1,8 +1,8 @@
-import datetime
 import os
 import re
 import time
 from concurrent.futures import ThreadPoolExecutor
+from datetime import datetime, timezone
```
mzappitello, line 5: i changed this import statement.

src/lamp_py/aws/s3.py
```diff
@@ -321,6 +321,60 @@ def file_list_from_s3_with_details(
...
+    if version and not version_check(newest_object["s3_obj_path"], version):
+        return None
+
+    for obj in objects:
+        if obj["last_modified"] > newest_object["last_modified"]:
+            newest_object = obj
+
+    return newest_object
+
+
```
mzappitello, line 377: two new functions for getting files and last modified information.

src/lamp_py/runtime_utils/remote_files.py
mzappitello, line 1: this is an attempt at the start of the file that will house all our remote filepaths. figured it made more sense to start it here rather than build on a pattern of having this information in files that use them.

src/lamp_py/bus_performance_manager/vehicle_events.py
```diff
@@ -0,0 +1,133 @@
...
+        file_datetime = get_datetime_from_partition_path(vp_file)
+
+        # if the partition is hourly (for older files), consider the hour of
+        # the file to determine the service date from the file datetime
+        if "hour" in vp_file:
+            if file_datetime.hour < 6:
+                yesterday = (file_datetime - timedelta(days=1)).date()
+                add_vp_file(yesterday, vp_file)
+            if file_datetime.hour > 3:
+                add_vp_file(file_datetime.date(), vp_file)
```
rymarczy, line 106: `file_datetime` here should be a UTC datetime, so I don't think these hour conditionals would be applicable.

src/lamp_py/bus_performance_manager/vehicle_events.py
```diff
@@ -0,0 +1,133 @@
...
+            day = int(service_date_int[6:])
+
+            service_date = date(year=year, month=month, day=day)
+            add_tm_file(service_date, tm_file)
+        except IndexError:
+            # the tm files may have a lamp version file that will throw when
+            # pulling out a match from the regular expression. ask for
+            # forgiveness and assert that it was this file that caused the
+            # error.
+            assert "lamp_version" in tm_file
```
rymarczy, line 131: So this will end up throwing and exiting the loop if `lamp_version` isn't in the file?

src/lamp_py/bus_performance_manager/vehicle_events.py
```diff
@@ -0,0 +1,133 @@
...
+    return vp_files, tm_files
+
+
+def files_per_service_date(
+    vp_files: List[str], tm_files: List[str]
+) -> Dict[date, BusInputFiles]:
+    """
+    take a list of input files and bucket them based on the service date they
+    cover.
+    """
```
rymarczy, line 73: I believe there could be a flaw with the methodology of this function where the `get_new_event_files` function picks up `vp_files` for a certain service date, but no `tm_files` are returned for that service date (because the files aren't available based on the `cutoff` parameter. 

For the next step of this process to work correctly I believe we will need all `vp_files` and `tm_files` for a specific `service_date` whether the input files were modified or not. 

An alternate approach could be determining the list of `service_date`s that need to be processed based on `get_last_modified_object` and then pulling all available `vp_files` and `tm_files` for each of those service dates.

rymarczy (changes requested):
I think an alternate approach could be warranted for the `files_per_service_date` function, let me know if a quick chat/call would help to talk through pros/cons.

src/lamp_py/aws/s3.py
```diff
@@ -321,6 +321,60 @@ def file_list_from_s3_with_details(
...
+    # if there are no objects, return early.
+    if len(objects) == 0:
+        return None
+
+    # start with the first object
+    newest_object = objects[0]
+
+    # if the first version fails that version check, return early
+    if version and not version_check(newest_object["s3_obj_path"], version):
+        return None
```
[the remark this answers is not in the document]
mzappitello: it does. i guess the alternative approach would be to sort the files by last modified and then reverse through them until we have a version match.

src/lamp_py/bus_performance_manager/vehicle_events.py
mzappitello: i think that makes sense.

* pull service dates from file names
* another round of s3 calls to get all files for those service dates
* return the maps of service dates to files.

src/lamp_py/bus_performance_manager/vehicle_events.py
mzappitello: yeah. thats the only file in the tm_files list that shouldn't be caught by the regular expression. if there is another file in there that causes an empty list on the match we would want to know about it.

src/lamp_py/aws/s3.py
```diff
@@ -321,6 +321,37 @@ def file_list_from_s3_with_details(
...
+
+    # if there is a version, return the first object that matches the version
+    for file in files:
+        if version_check(file["s3_obj_path"], version):
+            return file
+
+    # if we were unable to match an object, return None
+    return None
+
+
```
mzappitello, line 354: i updated this function to only return the file that also has a version match.

mzappitello: i ran this against mbta-ctd-dataplatform-dev-springboard/lamp/RT_VEHICLE_POSITIONS, which has 4054 files with a version match and it took 6 and half minutes. if i drop the version match it takes 1.2 seconds. i get a similar runtime if i do a vesion match against the empty string (the version info of all these files).

src/lamp_py/bus_performance_manager/vehicle_events.py
```diff
@@ -0,0 +1,166 @@
...
+from lamp_py.aws.s3 import (
+    file_list_from_s3_with_details,
+    get_datetime_from_partition_path,
+    get_last_modified_object,
+)
+
+from lamp_py.runtime_utils.remote_files import RemoteFileLocations
+
+
+def get_new_event_files() -> pl.DataFrame:
```
mzappitello, line 16: i replaced how this file works, using polars. it will generate a df where each record is a service date and a two lists (gtfs_rt and tm) of files associated with that service date. on dev it takes just over a second.

rymarczy, line 16: suggestion: Returning a dataframe here creates a bit of a black box with regards to our CI tools. Something like a named tuple of dictionaries could help with using the output of this function by downstream parts of the application.

src/lamp_py/bus_performance_manager/vehicle_events.py
```diff
@@ -0,0 +1,166 @@
...
+    )
+
+    # filter all files to only files associated with new service dates
+    # aggregate records by source and pivot on service date.
+    #
+    # each record of the new dataframe will have a list of gtfs_rt and tm input
+    # files.
+    grouped_files = (
+        all_files.filter(
+            pl.col("service_date").is_in(new_service_dates["service_date"])
```
rymarczy, line 159: Some of this filtering logic seems a bit overly complicated. 

Is there a way, using `get_last_modified_object`, to determine the range of service dates needing to be processed, and then filter `all_files` by `service_date`?

rymarczy (commented):
A few suggestions on this. 

One functional comment. Currently, for a specific service date, this implementation can return `transit_master` files and no `gtfs_rt` file or vice versa, meaning a downstream function would have to parse the dataframe records and determine which service dates to actually process.

Should this function check that, for a specific service date, files of both types are being returned and throw errors/warnings if only transit master or only gtfs files are being returned?

src/lamp_py/bus_performance_manager/vehicle_events.py
```diff
@@ -0,0 +1,120 @@
...
+        | pl.col("s3_obj_path").str.contains("hour=2")
+        | pl.col("s3_obj_path").str.contains("hour=3")
+        | pl.col("s3_obj_path").str.contains("hour=4")
+        | pl.col("s3_obj_path").str.contains("hour=5")
+        | pl.col("s3_obj_path").str.contains("hour=6")
+        | pl.col("s3_obj_path").str.contains("hour=7")
+        | pl.col("s3_obj_path").str.contains("hour=8")
+        | ~pl.col("s3_obj_path").str.contains("hour")
+    ).with_columns(
+        [(pl.col("service_date") - timedelta(days=1)).alias("service_date")]
```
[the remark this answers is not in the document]
mzappitello: these values came out of testing february and april dates, so i think it should be ok? we could do something where we take dst into account in the filter, but that feels overly complicated without much bennefit.

src/lamp_py/bus_performance_manager/vehicle_events.py
mzappitello: i was trying to not to build too many assumptions into the logic. we could do something where we get the service date from the last modified object and assume that we only service dates after that one to do?

rymarczy (approved):
We'll still need to figure out how to handle mis-matched service_date results from this function (eg service dates with transit master files, but no gtfs_rt files). But we that can be handled by some separate work scope.

PR #409, opened 2024-07-24
FIX: name tableau dataset based on deploy environment

rymarczy:
Ops Analytics has requested that Tableau datasets have different names based on development environment (prod, staging, dev).

This change renames staging and dev environment datasets to append `_staging` or `_dev`. Production datasets names remain un-changed. 

Asana Task: https://app.asana.com/0/1205827492903547/1207800774457790


mzappitello (approved):
🍰

PR #410, opened 2024-07-24
Convert GTFS RT Records to Bus Events for BPM

mzappitello:
Add some logic for taking incoming gtfs realtime files, creating a dataframe of vehicle positions, reducing it to arrivals and departures, and pivoting it into one record per trip/stop.

Asana Task: [🚌 Convert GTFS-RT data into bus vehicle events](https://app.asana.com/0/1205827492903547/1207771349226041/f)


src/lamp_py/bus_performance_manager/vehicle_events.py
```diff
@@ -152,3 +155,115 @@ def get_service_date_from_filename(tm_filename: str) -> Optional[date]:
...
+            pl.col("vehicle.trip.direction_id").alias("direction_id"),
+            pl.col("vehicle.trip.start_time").alias("start_time"),
+            pl.col("vehicle.trip.start_date").alias("service_date"),
+            pl.col("vehicle.vehicle.id").cast(pl.String).alias("vehicle_id"),
+            pl.col("vehicle.current_status").alias("current_status"),
+            pl.from_epoch("vehicle.timestamp")
+            .dt.convert_time_zone("EST5EDT")
+            .alias("vehicle_timestamp"),
+        )
+        .unique()
```
rymarczy, line 221: Does having this `unique()` in front of the next `unique` serve any specific purpose?

src/lamp_py/bus_performance_manager/vehicle_events.py
```diff
@@ -152,3 +155,115 @@ def get_service_date_from_filename(tm_filename: str) -> Optional[date]:
...
+        .sort(["vehicle_id", "vehicle_timestamp"])
+        .collect()
+    )
+
+    # using the vehicle positions dataframe, create a dataframe for each event
+    # by pivoting and mapping the current status onto arrivals and departures.
+    # initially, they map to arrival times and traveling towards times. shift
+    # the traveling towards datetime over trip id to get the departure time.
+    vehicle_events = (
+        vehicle_positions.pivot(
```
rymarczy, line 243: Pivoting on `current_status` could potentially result in 3 or 4 additional dataframe columns (depending on how NULL values are handled). 

If we're pivoting on this column we should do something to handle NULL values explicitly and, however in-frequently it appears, the `INCOMING_AT` - `VehicleStopStatus` ENUM

src/lamp_py/bus_performance_manager/vehicle_events.py
```diff
@@ -152,3 +155,115 @@ def get_service_date_from_filename(tm_filename: str) -> Optional[date]:
...
+            {
+                "STOPPED_AT": "arrival_gtfs",
+                "IN_TRANSIT_TO": "started_towards_datetime",
+            }
+        )
+        .with_columns(
+            pl.col("started_towards_datetime")
+            .shift(-1)
+            .over("trip_id")
+            .alias("departure_gtfs")
```
rymarczy, line 265: I would advise against doing any shift operations until after this dataset is joined with Transit  Master data. Since we would presumably want to do the same shift operations against that dataset as well.

src/lamp_py/bus_performance_manager/vehicle_events.py
```diff
@@ -152,3 +155,115 @@ def get_service_date_from_filename(tm_filename: str) -> Optional[date]:
...
+                "route_id",
+                "trip_id",
+                "stop_id",
+                "direction_id",
+                "vehicle_id",
+                "current_status",
+            ],
+            keep="first",
+        )
+        .sort(["vehicle_id", "vehicle_timestamp"])
```
rymarczy, line 234: Probably don't need to do this `sort` here, as it will likely get un-sorted when joined to transit master data and then re-sorted for another purpose.

src/lamp_py/bus_performance_manager/vehicle_events.py
```diff
@@ -152,3 +155,115 @@ def get_service_date_from_filename(tm_filename: str) -> Optional[date]:
...
+    """get a list of bus route ids for a given service date"""
+    routes_file = get_gtfs_parquet_file(
+        year=service_date.year, filename="routes.parquet"
+    ).get_s3_path()
+
+    # generate an integer date for the service date
+    target_date = int(service_date.strftime("%Y%m%d"))
+
+    bus_routes = (
+        pl.scan_parquet(routes_file)
```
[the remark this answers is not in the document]
mzappitello: i think we can build some kind of caching decorator on these calls in an upcoming ticket.

src/lamp_py/bus_performance_manager/vehicle_events.py
mzappitello: i had it to speed up the sort, but only because our daily partitions aren't unique coming in. happy to remove it.

src/lamp_py/bus_performance_manager/vehicle_events.py
```diff
@@ -152,3 +155,115 @@ def get_service_date_from_filename(tm_filename: str) -> Optional[date]:
...
+        )
+        .collect()
+    )
+
+    return bus_routes["route_id"].unique().to_list()
+
+
+def generate_gtfs_rt_events(
+    service_date: date, gtfs_rt_files: List[str]
+) -> pl.DataFrame:
```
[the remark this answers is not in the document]
mzappitello: i'm splitting up, going with one for getting files to ingest, one for gtfs utilities, one for gtfs_rt ingestion and one for tm ingestion.

src/lamp_py/bus_performance_manager/vehicle_events.py
mzappitello: i filtered out null current status from the start and am converting all `INCOMING_AT` to `STOPPED_AT` columns later on.

src/lamp_py/bus_performance_manager/gtfs_rt_ingestion.py
```diff
@@ -0,0 +1,115 @@
...
+            (pl.col("vehicle.trip.route_id").is_in(bus_routes))
+            & (
+                pl.col("vehicle.trip.start_date")
+                == service_date.strftime("%Y%m%d")
+            )
+            & pl.col("vehicle.current_status").is_not_null()
+            & pl.col("vehicle.stop_id").is_not_null()
+            & pl.col("vehicle.trip.trip_id").is_not_null()
+            & pl.col("vehicle.vehicle.id").is_not_null()
+            & pl.col("vehicle.timestamp").is_not_null()
```
mzappitello, line 39: additional filters to ensure our data is useable.

src/lamp_py/bus_performance_manager/gtfs_rt_ingestion.py
```diff
@@ -0,0 +1,115 @@
...
+            .cast(pl.String)
+            .alias("start_time"),
+            pl.col("vehicle.trip.start_date")
+            .cast(pl.String)
+            .alias("service_date"),
+            pl.col("vehicle.vehicle.id").cast(pl.String).alias("vehicle_id"),
+            pl.col("vehicle.current_status")
+            .cast(pl.String)
+            .alias("current_status"),
+            pl.from_epoch("vehicle.timestamp").alias("vehicle_timestamp"),
```
mzappitello, line 58: everything is cast.

src/lamp_py/bus_performance_manager/gtfs_rt_ingestion.py
```diff
@@ -0,0 +1,115 @@
...
+        )
+        .with_columns(
+            pl.when(pl.col("current_status") == "INCOMING_AT")
+            .then(pl.lit("STOPPED_AT"))
+            .otherwise(pl.col("current_status"))
+            .cast(pl.String)
+            .alias("current_status"),
```
mzappitello, line 65: modify current status in cases where INCOMING_AT is used (which seems rare if at all)

src/lamp_py/bus_performance_manager/gtfs_rt_ingestion.py
```diff
@@ -0,0 +1,115 @@
...
+        .collect()
+    )
+
+    if vehicle_positions.is_empty():
+        # if there are now elements, adjust the columns to match the return
+        # dataframe schema.
+        vehicle_events = vehicle_positions.with_columns(
+            pl.lit(None).cast(pl.Datetime).alias("arrival_gtfs"),
+            pl.lit(None).cast(pl.Datetime).alias("travel_towards_gtfs"),
+        ).drop("current_status", "vehicle_timestamp")
```
mzappitello, line 88: modify the columns in the case where the df is empty. this ensures it will match the return schema.

src/lamp_py/bus_performance_manager/gtfs_rt_ingestion.py
```diff
@@ -0,0 +1,115 @@
...
+                "service_date",
+                "vehicle_id",
+            ],
+            columns="current_status",
+        ).rename(
+            {
+                "STOPPED_AT": "arrival_gtfs",
+                "IN_TRANSIT_TO": "travel_towards_gtfs",
+            }
+        )
```
mzappitello, line 113: pivot the df on current status to get arrival and departure times.

src/lamp_py/bus_performance_manager/gtfs_utils.py
```diff
@@ -0,0 +1,29 @@
+from datetime import date
+from typing import List
+
+import polars as pl
+
+from lamp_py.runtime_utils.remote_files import get_gtfs_parquet_file
+
+
+def bus_routes_for_service_date(service_date: date) -> List[str]:
```
mzappitello, line 9: i moved this function into this file to keep concerns separated.

src/lamp_py/runtime_utils/remote_files.py
```diff
@@ -11,6 +11,10 @@ class S3Location:
     bucket_name: str
     file_prefix: str
 
+    def get_s3_path(self) -> str:
+        """generate the full s3 url that can be used to describe the location"""
+        return f"s3://{self.bucket_name}/{self.file_prefix}"
```
mzappitello, line 16: useful utility when a location is a parquet file.

src/lamp_py/runtime_utils/remote_files.py
```diff
@@ -55,3 +59,15 @@ class RemoteFileLocations:
...
+def get_gtfs_parquet_file(year: int, filename: str) -> S3Location:
+    """
+    generate an S3Location instance for a gtfs schedule parquet file for a
+    given type in a given year.
+    """
+    file_object = os.path.join("lamp", "gtfs_archive", str(year), filename)
+
+    return S3Location(
+        bucket_name=RemoteFileLocations.public_bucket, file_prefix=file_object
+    )
```
mzappitello, line 73: generate parquet file path for a gtfs schedule parquet file.

tests/conftest.py
```diff
@@ -52,3 +56,94 @@ def mock__get_pyarrow_dataset(
     )
 
     yield
+
+
+@pytest.fixture(autouse=True, name="remote_file_locations_patch")
```
mzappitello, line 61: fixture to patch over s3 locations with local locations.

tests/performance_manager/test_performance_manager.py
```diff
@@ -85,7 +85,10 @@ def test_files() -> List[str]:
         for file in files:
             if "DS_Store" in file:
                 continue
-            paths.append(os.path.join(root, file))
+            path = os.path.join(root, file)
+            if "year=2024" in path:
+                continue
+            paths.append(path)
```
mzappitello, line 91: temp patch to stop this test from consuming new data.

ideally, we would have one set of vp files that work for both tests.

src/lamp_py/bus_performance_manager/gtfs_rt_ingestion.py
```diff
@@ -0,0 +1,115 @@
...
+            pl.col("vehicle.vehicle.id").cast(pl.String).alias("vehicle_id"),
+            pl.col("vehicle.current_status")
+            .cast(pl.String)
+            .alias("current_status"),
+            pl.from_epoch("vehicle.timestamp").alias("vehicle_timestamp"),
+        )
+        .with_columns(
+            pl.when(pl.col("current_status") == "INCOMING_AT")
+            .then(pl.lit("STOPPED_AT"))
+            .otherwise(pl.col("current_status"))
```
rymarczy, line 63: `INCOMING_AT` technically means the vehicle is still moving, with rail we convert this to `IN_TRANSIT_TO`

src/lamp_py/bus_performance_manager/gtfs_rt_ingestion.py
```diff
@@ -0,0 +1,115 @@
...
+            subset=[
+                "route_id",
+                "trip_id",
+                "stop_id",
+                "direction_id",
+                "vehicle_id",
+                "current_status",
+            ],
+            keep="first",
+        )
```
rymarczy, line 78: This `.unique` call is semi-redundant to the `pivot` table that is called later. Calling `pivot` with a `min` aggregation function on `vehicle_timestamp` would probably be slightly more explicit in what we are trying to do here.

src/lamp_py/bus_performance_manager/gtfs_rt_ingestion.py
```diff
@@ -0,0 +1,115 @@
...
+                == service_date.strftime("%Y%m%d")
+            )
+            & pl.col("vehicle.current_status").is_not_null()
+            & pl.col("vehicle.stop_id").is_not_null()
+            & pl.col("vehicle.trip.trip_id").is_not_null()
+            & pl.col("vehicle.vehicle.id").is_not_null()
+            & pl.col("vehicle.timestamp").is_not_null()
+        )
+        .select(
+            pl.col("vehicle.trip.route_id").cast(pl.String).alias("route_id"),
```
rymarczy, line 42: I think we're missing a few columns needing to be pulled, as defined by the bus requirements notion Doc. 

such as:
stop_sequence
vehicle_label

rymarczy (changes requested):
Just a few final changes. 

We also might want to consider increasing the max line length from 80 to something like 120. That 80 character limit is making these polars statements very ugly.

src/lamp_py/bus_performance_manager/gtfs_rt_ingestion.py
mzappitello: i went through the doc and pulled all the additional gtfs_rt sourced fields. it was just these two.

src/lamp_py/bus_performance_manager/gtfs_rt_ingestion.py
```diff
@@ -0,0 +1,109 @@
...
+                "direction_id",
+                "trip_id",
+                "stop_id",
+                "stop_sequence",
+                "start_time",
+                "service_date",
+                "vehicle_id",
+                "vehicle_label"
+            ],
+            columns="current_status",
```
rymarczy, line 101: Coming out of this `pivot` operations, we're not 100% guaranteed to have a `STOPPED_AT` and `IN_TRANSIT_TO` column, so we should be validating that and or/adding the columns if they don't exist before the `rename` operation.

rymarczy (commented):
One last comment, and fix the black formatting and we should be good to go!

src/lamp_py/bus_performance_manager/gtfs_rt_ingestion.py
mzappitello: i split the main ingestion function into two functions to make testing easier.

in the pivot method, i check for columns, add null columns if one (or both) is missing, rename, and then run a select to ensure the order is always the same.

PR #415, opened 2024-07-29
FEAT: Export Transit Master Waiver Table

rymarczy:
Ingest waiver table at the request of Ops Analytics. 

This PR is awaiting confirmation from Ops Analytics that this is the correct table to be ingesting. 

Asana Task: https://app.asana.com/0/1205827492903547/1207790219568026


mzappitello:
it looks like this wavier table is empty on aws as well. :/

PR #420, opened 2024-08-09
Ingest More Transit Master Tables

mzappitello:
* Add Block, Operator, Run and WorkPiece to the whole table ingestion jobs.
* Create new abstract base class `TMDailyTable` that will pull in daily tables from transit master. (Its largely an abstraction based on the existing `TMDailyLogStopCrossing` class.
* Add Daily Work Piece and Schedule Adherence Waiver tables to the daily log jobs.
* All TM Ingestion jobs now use Remote File Location constants to define s3 locations.
* Add tests to ensure all TM Ingestion Jobs are run in the pipeline.

Asana Task: [🚌 Ingest TM data for block, run, & operator](https://app.asana.com/0/1205827492903547/1207935724478899/f)

### DIFF
--- a/src/lamp_py/ingestion_tm/ingest.py
+++ b/src/lamp_py/ingestion_tm/ingest.py
@@ -11,11 +11,11 @@ from lamp_py.ingestion_tm.jobs.whole_table import (
     TMMainOperator,
     TMMainRun,
     TMMainWorkPiece,
+    TMDailyLogDailySchedAdhereWaiver,
 )
 from lamp_py.ingestion_tm.jobs.parition_table import (
     TMDailyLogStopCrossing,
     TMDailyLogDailyWorkPiece,
-    TMDailyLogDailySchedAdhereWaiver,
 )
 
 
--- a/src/lamp_py/ingestion_tm/ingest.py
+++ b/src/lamp_py/ingestion_tm/ingest.py
@@ -7,8 +7,35 @@ from lamp_py.ingestion_tm.jobs.whole_table import (
     TMMainRoute,
     TMMainTrip,
     TMMainVehicle,
+    TMMainBlock,
+    TMMainOperator,
+    TMMainRun,
+    TMMainWorkPiece,
 )
-from lamp_py.ingestion_tm.jobs.parition_table import TMDailyLogStopCrossing
+from lamp_py.ingestion_tm.jobs.parition_table import (
+    TMDailyLogStopCrossing,
+    TMDailyLogDailyWorkPiece,
+    TMDailyLogDailySchedAdhereWaiver,
+)
+
+
+def get_ingestion_jobs() -> List[TMExport]:
+    """
+    get a list of all ingestion jobs that
+    """
+    return [
+        TMMainGeoNode(),
+        TMMainRoute(),
+        TMMainTrip(),
+        TMMainVehicle(),
+        TMMainBlock(),
+        TMMainOperator(),
+        TMMainRun(),
+        TMMainWorkPiece(),
+        TMDailyLogStopCrossing(),
+        TMDailyLogDailyWorkPiece(),
+        TMDailyLogDailySchedAdhereWaiver(),
+    ]
 
 
 def ingest_tables() -> None:
@@ -16,14 +43,7 @@ def ingest_tables() -> None:
     ingest tables from transmaster database
     """
     tm_db = MSSQLManager(verbose=True)
-
-    jobs: List[TMExport] = [
-        TMMainGeoNode(),
-        TMMainRoute(),
-        TMMainTrip(),
-        TMMainVehicle(),
-        TMDailyLogStopCrossing(),
-    ]
+    jobs: List[TMExport] = get_ingestion_jobs()
 
     for job in jobs:
         job.run_export(tm_db)
--- a/src/lamp_py/ingestion_tm/jobs/parition_table.py
+++ b/src/lamp_py/ingestion_tm/jobs/parition_table.py
@@ -160,12 +160,11 @@ class TMDailyTable(TMExport):
                     )
                     upload_file(local_pq, s3_export_path)
 
+                self.update_version_file()
                 logger.log_complete()
 
             except Exception as exception:
                 logger.log_failure(exception)
-
-        self.update_version_file()
 
 
 class TMDailyLogStopCrossing(TMDailyTable):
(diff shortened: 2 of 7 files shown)


src/lamp_py/ingestion_tm/ingest.py
```diff
@@ -7,23 +7,43 @@
...
+from lamp_py.ingestion_tm.jobs.parition_table import (
+    TMDailyLogStopCrossing,
+    TMDailyLogDailyWorkPiece,
+    TMDailyLogDailySchedAdhereWaiver,
 )
-from lamp_py.ingestion_tm.jobs.parition_table import TMDailyLogStopCrossing
 
 
-def ingest_tables() -> None:
+def get_ingestion_jobs() -> List[TMExport]:
```
mzappitello, line 22: put this into its own function so we can test it for completeness

src/lamp_py/ingestion_tm/ingest.py
```diff
@@ -7,23 +7,43 @@
...
     TMMainVehicle,
+    TMMainBlock,
+    TMMainOperator,
+    TMMainRun,
+    TMMainWorkPiece,
+)
+from lamp_py.ingestion_tm.jobs.parition_table import (
+    TMDailyLogStopCrossing,
+    TMDailyLogDailyWorkPiece,
+    TMDailyLogDailySchedAdhereWaiver,
```
mzappitello, line 18: new ingestion cleasses.

src/lamp_py/ingestion_tm/jobs/parition_table.py
```diff
@@ -12,61 +12,33 @@
...
 )
 
 
-class TMDailyLogStopCrossing(TMExport):
-    """Export STOP_CROSSING table from TMDailyLog"""
-
-    def __init__(self) -> None:
-        TMExport.__init__(self)
+class TMDailyTable(TMExport):
+    """Export Daily table from TMDailyLog"""
```
mzappitello, line 25: i converted this to a base class that other daily log tables can use for export.

src/lamp_py/ingestion_tm/jobs/parition_table.py
```diff
@@ -149,26 +123,30 @@ def dates_to_export(self, tm_db: MSSQLManager) -> List[int]:
     def run_export(self, tm_db: MSSQLManager) -> None:
         table_columns = ",".join([col.name for col in self.export_schema])
 
-        logger = ProcessLogger("tm_stop_crossing_export")
```
mzappitello, line 152: if `tm_db.write_to_parquet`fails, this process was still running as if nothing was wrong, making debugging difficult because we can't see what subclass has the problem. now it'll be caught here with more debugging information.

src/lamp_py/ingestion_tm/jobs/parition_table.py
```diff
@@ -182,8 +160,122 @@ def run_export(self, tm_db: MSSQLManager) -> None:
...
 
-        except Exception as exception:
-            logger.log_failure(exception)
+            except Exception as exception:
+                logger.log_failure(exception)
+
+        self.update_version_file()
+
+
+class TMDailyLogStopCrossing(TMDailyTable):
```
mzappitello, line 171: impl of the StopCrossing table.

src/lamp_py/ingestion_tm/jobs/parition_table.py
```diff
@@ -182,8 +160,122 @@ def run_export(self, tm_db: MSSQLManager) -> None:
...
+                ("CROSSING_TYPE_ID", pyarrow.int64()),
+                ("OPERATOR_ID", pyarrow.int64()),
+                ("CANCELLED_FLAG", pyarrow.bool_()),
+                ("IS_LAYOVER", pyarrow.bool_()),
+                ("ROUTE_ID", pyarrow.int64()),
+            ]
+        )
+
+
+class TMDailyLogDailyWorkPiece(TMDailyTable):
```
mzappitello, line 216: This table links operators to pieces of work, allowing us to figure out who drove what runs, blocks, and trips.

src/lamp_py/ingestion_tm/jobs/parition_table.py
```diff
@@ -182,8 +160,122 @@ def run_export(self, tm_db: MSSQLManager) -> None:
...
+                ("PULLOUT_ID", pyarrow.int64()),
+                ("RUN_ID", pyarrow.int64()),
+                ("ASSIGNED_VEHICLE_ID", pyarrow.int64()),
+                ("CURRENT_VEHICLE_ID", pyarrow.int64()),
+                ("INSERTED_FLAG", pyarrow.bool_()),
+            ]
+        )
+
+
+class TMDailyLogDailySchedAdhereWaiver(TMDailyTable):
```
mzappitello, line 250: i think this where we can get waivers that Ops Analytics was requesting.

src/lamp_py/ingestion_tm/jobs/whole_table.py
```diff
@@ -15,19 +16,11 @@ class TMWholeTable(TMExport):
 
     def __init__(
         self,
-        pq_file_name: str,
+        s3_location: S3Location,
```
mzappitello, line 19: use `S3Location` instead so that filepaths aren't defined here.

src/lamp_py/ingestion_tm/jobs/whole_table.py
```diff
@@ -216,3 +209,124 @@ def export_schema(self) -> pyarrow.schema:
                 ("RADIO_DEVICE_ID", pyarrow.string()),
             ]
         )
+
+
+class TMMainOperator(TMWholeTable):
```
mzappitello, line 214: add new tables

src/lamp_py/ingestion_tm/tm_export.py
```diff
@@ -12,11 +11,6 @@ class TMExport(ABC):
     Abstract Base Class for TM Export jobs
     """
 
-    def __init__(
-        self,
-    ) -> None:
-        self.export_bucket = os.getenv("SPRINGBOARD_BUCKET")
-
```
mzappitello, line 19: this is a part of the S3Location class each child is initialized with.

src/lamp_py/runtime_utils/remote_files.py
```diff
@@ -37,6 +37,14 @@ class RemoteFileLocations:
         bucket_name=springboard_bucket,
         file_prefix=os.path.join(tm_prefix, "STOP_CROSSING"),
     )
+    tm_daily_work_piece = S3Location(
```
mzappitello, line 40: add more S3Locations.

i feel like this class is ok for now, but could see a world where we want to break it up into TransitMasterLocations, GtfsRtLocations, RailPerformanceLocations, BusPerformanceLocations, etc.

tests/ingestion_tm/test_ingest.py
```diff
@@ -0,0 +1,40 @@
+import inspect
```
mzappitello, line 1: small test file, but it makes sure that when we're adding new TMExport classes, we remember to add them to the pipeline.

src/lamp_py/ingestion_tm/jobs/parition_table.py
```diff
@@ -182,8 +160,122 @@ def run_export(self, tm_db: MSSQLManager) -> None:
...
-            self.update_version_file()
-            logger.log_complete()
+                logger.log_complete()
 
-        except Exception as exception:
-            logger.log_failure(exception)
+            except Exception as exception:
+                logger.log_failure(exception)
+
+        self.update_version_file()
```
rymarczy, line 168: `update_version_file` should be in the `try` block and before `log_complete` or have it's own logging/error catching

rymarczy (changes requested):
Few comments/questions.

src/lamp_py/ingestion_tm/jobs/parition_table.py
mzappitello: moved it into the for loop.

src/lamp_py/mssql/mssql_utils.py
```diff
@@ -186,3 +186,4 @@ def write_to_parquet(
                     process_logger.log_failure(exception)
                 else:
                     logging.exception(exception)
+                    raise exception
```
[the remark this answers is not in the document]
mzappitello: ah. good call.

src/lamp_py/ingestion_tm/jobs/parition_table.py
```diff
@@ -182,8 +160,122 @@ def run_export(self, tm_db: MSSQLManager) -> None:
...
+        TMDailyTable.__init__(
+            self,
+            s3_location=RemoteFileLocations.tm_daily_sched_adherence_waiver,
+            tm_table="TMDailyLog.dbo.SCHED_ADHERE_WAIVER",
+            lamp_version="0.0.1",
+        )
+
+    @property
+    def export_schema(self) -> pyarrow.schema:
+        return pyarrow.schema(
```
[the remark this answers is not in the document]
mzappitello: as of today its 124,566 records. i'll move it to the full table export.

src/lamp_py/ingestion_tm/jobs/parition_table.py
```diff
@@ -182,8 +160,122 @@ def run_export(self, tm_db: MSSQLManager) -> None:
...
+
+class TMDailyLogDailyWorkPiece(TMDailyTable):
+    """Export DAILY_WORK_PIECE table from TMDailyLog"""
+
+    def __init__(self) -> None:
+        TMDailyTable.__init__(
+            self,
+            s3_location=RemoteFileLocations.tm_daily_work_piece,
+            tm_table="TMDailyLog.dbo.DAILY_WORK_PIECE",
+            lamp_version="0.0.1",
```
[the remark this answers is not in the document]
mzappitello: i looked into this and they appear to be related, but i'm not sure the cadence of the updates.

The `DAILY_WORK_PIECE_HISTORY` table also has the OPERATOR_ID and their logon / loggoff times and a Vehicle ID. However, there is more than one record per piece of work, with a "FORCED_LOGOFF" flag that seems to be flipped for different records on the same piece of work.

The Operator and Vehicle ids match, but the longon / logoff times only match one of the records. It also doesn't have the Pullout Id, which I think we may eventually want?

rymarczy (approved):
:cake:

PR #422, opened 2024-08-15
FEAT: Ingestion Light Rail Raw GPS files

rymarczy:
This change allows the Ingestion app to ingest Light Rail RAW GPS files. 

The process is currently configured to process a maximum batch size of ~5000 files on each Ingestion loop. 

All processing is done in memory with polars, this dataset is relatively small and 5000 files worth of data utilizes less than 2GB of data. A full event loop of 5000 files takes ~2 mins to process locally. This can be adjusted if event loop is taking too long on AWS. 

The process will also merge export files with existing files found in the public S3 bucket. 

Asana Task: https://app.asana.com/0/1205827492903547/1207059327607972


src/lamp_py/aws/s3.py
```diff
@@ -191,6 +192,36 @@ def object_metadata(obj: str) -> Dict[str, str]:
...
 
 
+def object_exists(obj: str) -> bool:
+    """
+    check if s3 object exists
+
+    :param obj - expected as 's3://my_bucket/object' or 'my_bucket/object'
+
+    :return: True is object exists, otherwise false
+    """
```
rymarczy, line 202: Convenience function to check if object exists on S3

src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -97,6 +98,8 @@ def __init__(
             self.detail = RtBusVehicleDetail()
         elif config_type == ConfigType.BUS_TRIP_UPDATES:
             self.detail = RtBusTripDetail()
+        elif config_type == ConfigType.LIGHT_RAIL:
+            raise IgnoreIngestion("Ignore LIGHT_RAIL files")
```
rymarczy, line 102: This seemed like the least intrusive way to handle skipping these files in GTFS-RT processing. Specific exception will trigger the ignoring of this file type.

src/lamp_py/ingestion/ingest_gtfs.py
```diff
@@ -100,6 +101,8 @@ def ingest_s3_files(metadata_queue: Queue[Optional[str]]) -> None:
                         config_type, metadata_queue
                     )
                 converters[config_type].add_files(file_group)
+            except IgnoreIngestion:
+                continue
```
rymarczy, line 105: Skip ignored files, don't move them

src/lamp_py/ingestion/light_rail_gps.py
```diff
@@ -0,0 +1,225 @@
...
+    """
+    gzip to dataframe converter function meant to be run in ThreadPool
+
+    :param path: path to gzip that will be converted to polars dataframe
+    """
+    file_system = current_thread().__dict__["file_system"]
+    path = path.replace("s3://", "")
+
+    logger = ProcessLogger(process_name="light_rail_gps_to_frame", path=path)
+    logger.log_start()
```
rymarczy, line 59: This is similar to our gzip process for RT files. In testing, about 16 threads produced the maximum throughput with minimal resource usage.

src/lamp_py/ingestion/light_rail_gps.py
```diff
@@ -0,0 +1,225 @@
...
+            max_workers=16, initializer=thread_init, initargs=(init_file,)
+        ) as pool:
+            for df, path in pool.map(thread_gps_to_frame, files):
+                if df is not None:
+                    archive_files.append(path)
+                    dfs.append(df)
+                else:
+                    error_files.append(path)
+
+        dataframe: pl.DataFrame = pl.concat(dfs)
```
rymarczy, line 140: Gather all records from all files into single dataframe.

src/lamp_py/ingestion/light_rail_gps.py
```diff
@@ -0,0 +1,225 @@
...
+    """
+    sync and write parquet files partitioned by date column
+
+    will merge any existing parquet files in memory
+    """
+    for date in dataframe.get_column("date").unique():
+        logger = ProcessLogger(
+            process_name="light_rail_write_parquet", date=date
+        )
+        logger.log_start()
```
rymarczy, line 167: Write parquet files by date pulled from "updated_at" column. 

This will merge existing parquet files from S3, if they exist and de-dupe the entire dataframe.

tests/ingestion/test_light_rail_gps.py
```diff
@@ -0,0 +1,28 @@
...
+        "INCOMING/2024-05-01T02:30:11Z_s3_mbta_ctd_trc_data_rtr_prod_LightRailRawGPS.json.gz",
+    )
+]
+
+
+def test_light_rail_gps() -> None:
+    """
+    test gtfs_events_for_date pipeline
+    """
+    dataframe, archive_files, error_files = dataframe_from_gz(mock_file_list)
```
rymarczy, line 20: Not totally sure if it's worth testing the rest of the process, which just includes a `file_list_from_s3` call and the parquet writing.

src/lamp_py/ingestion/light_rail_gps.py
```diff
@@ -0,0 +1,225 @@
...
+                        .str.to_date()
+                        .alias("date")
+                    ),
+                    pl.col("data").struct.field("updated_at").cast(pl.String),
+                    pl.col("data").struct.field("bearing").cast(pl.Float64),
+                    pl.col("data").struct.field("latitude").cast(pl.String),
+                    pl.col("data").struct.field("longitude").cast(pl.String),
+                    pl.col("data").struct.field("car").cast(pl.String),
+                )
+            )
```
mzappitello, line 86: this seems to be a nicer way to do this than using pyarrow.

rymarczy: Yeah, I also think most conversions between pyarrow and polars are zero-copy operations. The only tricky thing is making sure we're handling any type conversions correctly.

src/lamp_py/aws/s3.py
```diff
@@ -191,6 +192,36 @@ def object_metadata(obj: str) -> Dict[str, str]:
...
+            Bucket=bucket,
+            Key=obj,
+        )
+        return True
+
+    except botocore.exceptions.ClientError as exception:
+        if exception.response["Error"]["Code"] == "404":
+            return False
+
+    return False
```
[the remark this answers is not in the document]
rymarczy: per this documentation: https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html#API_HeadObject_Errors

the "404" error is the only thing that should be returned if the object does not exist in the bucket. Any other error would be un-defined behavior. Such as a permissions error, that should probably raise so that it's not masked by the function.

src/lamp_py/aws/s3.py
```diff
@@ -191,6 +192,37 @@ def object_metadata(obj: str) -> Dict[str, str]:
...
+        s3_client.head_object(
+            Bucket=bucket,
+            Key=obj,
+        )
+        return True
+
+    except botocore.exceptions.ClientError as exception:
+        if exception.response["Error"]["Code"] == "404":
+            return False
+        raise exception
```
rymarczy, line 221: raise all other exceptions, The 404 error code is the only thing that should be returned if the object does not exist. Any other exceptions would be undefined behavior

PR #423, opened 2024-08-16
FEAT: S3 remote file paths in one location

rymarczy:
This changes attempts to consolidate references to S3 paths all in one location. 

Original version of `remote_files.py` was modified to reduce the verbosity of calling resources defined by the file. Also added all bucket references and lamp file prefix constants to the file. 

Hopefully I found all the places where we reference S3 file locations....

Asana Task: https://app.asana.com/0/1205827492903547/1207771349226027

PR #427, opened 2024-08-20
FIX: Bad Compress GTFS Local Folder

rymarczy:
PR #423 re-factor broke compressed GTFS process by prefixing the `gtfs_temp_folder` declaration with s3://. 

This change fixes that oversight.


mzappitello (approved):
looks good to me. i'm assuming this issue didn't pop up anywhere else on staging?

rymarczy:
> looks good to me. i'm assuming this issue didn't pop up anywhere else on staging?

I don't think so... I was trying to get a little clever with that variable assignment and that's never a good idea.

I double checked all the other changes where the `.s3_uri` property was used and they all expect to have a `s3://` prefix

PR #440, opened 2024-09-17
FEAT: Join BUS GTFS-RT Vehicle Events with Transit Master Vehicle Events

rymarczy:
The change allows for the joining of Bus events from GTFS-RT and Transit Master (TM) data sources. 

GTFS-RT events are joined to TM events with an "asof" join. 

This type of join first performs a regular LEFT JOIN on the columns of "route_id", "trip_id", "vehicle_label" and "stop_id" and then performs a nearest match on "stop_sequence". The resulting dataframe retains a "stop_sequence" (from gtfs-rt) and "tm_stop_sequence" column to verify accuracy of stop_sequence nearest join. 

The dataframe of joined events will be used as input to create performance metrics that have been requested by Ops Analytics. 

Asana Task: https://app.asana.com/0/1205827492903547/1207771349226047


arkadyan (approved):
👍 

Sorry for the delay!

PR #446, opened 2024-10-01
FEAT: Add 'revenue' field to Rail data pipeline

rymarczy:
This change adds the GTFS-RT `revenue` field to the LAMP Rail Performance Data pipeline.

1. Create db migrations to add `revenue` column to `vehicle_trips` table
2. Read `vehicle.trip.revenue` column from VehiclePositions parquet files
3. Load `revenue` data from parquet files into DB table
4. Add `is_revenue` column to `LAMP_ALL_RT_fields` OPMI export
5. Add where clause for only `revenue` trips to Subway Performance Data export

The use of the `revenue` field differs between the OPMI export and Subway Performance Data export that TransitMatters utilizes. For the OMPI export an additional field is added so that revenue trips can be filtered out at their discretion. The Subway Performance Data export was never meant to include non-revenue event data, so it is filtered out from the export all together. 

Asana Task: https://app.asana.com/0/1205827492903547/1208216161546522


arkadyan (approved):
👍

PR #448, opened 2024-10-04
FEAT: Ad-Hoc Runner Script to add revenue to Springboard file

rymarczy:
This change creates and Ad-Hoc Runner script that will add a "vehicle.trip.revenue" column to RT_VEHICLE_POSITION springboard parquet files starting from December 20th, 2023. 

The Ad-Hoc runner will re-read GTFS-RT VehiclePosition input file, from the archive bucket, and produce new daily partitioned parquet files that include a "vehicle.trip.revenue" column. Existing parquet files that do not have a "vehicle.trip.revenue" column will be overwritten.

After the files have been created they are added to our metadata RDS table so that the performance manger application may re-ingest the files and add an accurate "revenue" values to the vehicle_event_trips table of the performance manger RDS. 

This process will run in parallel with LAMP's real-time performance pipeline so that customer data deliver will not be impacted. 

This process has already been testing on the dev environment, but will also be run on staging to confirm no adverse conditions are created. 

Asana Task: https://app.asana.com/0/1205827492903547/1208216161546522
Asana Task: https://app.asana.com/0/1205827492903547/1207556073650893

### DIFF
--- a/src/lamp_py/ad_hoc/runner_001.py
+++ b/src/lamp_py/ad_hoc/runner_001.py
@@ -1,9 +1,126 @@
+import os
+from datetime import date
+from datetime import timedelta
+
+import pyarrow.parquet as pq
+
 from lamp_py.runtime_utils.process_logger import ProcessLogger
+from lamp_py.postgres.postgres_utils import start_rds_writer_process
+from lamp_py.runtime_utils.remote_files import S3_ARCHIVE
+from lamp_py.runtime_utils.remote_files import LAMP
+from lamp_py.runtime_utils.remote_files import S3_SPRINGBOARD
+from lamp_py.aws.s3 import file_list_from_s3
+from lamp_py.aws.s3 import download_file
+from lamp_py.ingestion.convert_gtfs_rt import GtfsRtConverter
+from lamp_py.ingestion.converter import ConfigType
+
+
+# pylint: disable=R0801
+class AdHocConverter(GtfsRtConverter):
+    """Custom ad-hoc Converter Class"""
+
+    def convert(self) -> None:
+        """
+        ad-hoc Convert process to add revenue field to parquet files
+
+        will read from Archive bucket and not move any files after processing
+        """
+        process_logger = ProcessLogger(
+            "ad_hoc_converter",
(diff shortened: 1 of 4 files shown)


src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -222,7 +222,7 @@ def process_files(self) -> Iterable[pyarrow.table]:
         process_logger.log_complete()
 
     def yield_check(
-        self, process_logger: ProcessLogger, min_rows: int = 5_000_000
+        self, process_logger: ProcessLogger, min_rows: int = 2_000_000
```
rymarczy, line 225: Reduced yield size to allow for the smaller ECS instance size of the ad-hoc process. This appears to not have any significant impact on ingestion processing speed.

src/lamp_py/ad_hoc/runner_001.py
```diff
@@ -1,9 +1,126 @@
...
+                process_logger.add_metadata(table_count=table_count)
+
+        except Exception as exception:
+            process_logger.log_failure(exception)
+        else:
+            process_logger.log_complete()
+        finally:
+            self.clean_local_folders()
+
+    def sync_with_s3(self, local_path: str) -> bool:
```
arkadyan, line 52: question: Just curious, is this function used anywhere?

arkadyan (approved):
👍

src/lamp_py/ad_hoc/runner_001.py
rymarczy: Yes, it gets called in the base class: https://github.com/mbta/lamp/blob/42bb38b8afb459bf019542d54ffe5df04e98860f/src/lamp_py/ingestion/convert_gtfs_rt.py#L399

The `AdHocConverter` class just modifies the minimum functionality of the base class that allows these migration operations. 

The main modifications are removing `self.move_s3_files()` from the `convert` method. This would usually move files from our `incoming` S3 bucket to `archive`, but this process just reads straight from the `archive` bucket and moves nothing. 

And the `sync_with_s3` modifications that I put in the PR description.

PR #456, opened 2024-10-24
FEAT: Add `last_trip` to ingestion Trip Descriptor for GTFS-RT

rymarczy:
Performed an audit of available GTFS-RT fields to confirm no other fields were missing from the LAMP ingestion application. 

`last_trip` is the only field found that was not already being ingested. 

Asana Task: https://app.asana.com/0/1205827492903547/1208279222032058


arkadyan (approved):
👍

PR #457, opened 2024-10-25
FIX: Reset RT_ALERTS Ingestion dataset on schema change

rymarczy:
The RT_ALERTS dataset that we ingest produces parquet files that contain columns consisting of struct types. If the struct type of the column changes, a newly ingested dataset can not be merged with the existing dataset from S3. 

If this situation emerges, the S3 dataset will be over-written by the newly ingested dataset, avoiding the merge errors for the day. This will only be applied to Alerts datasets, as the contents of Alerts RT data does not change often throughout the day. Since Alerts events are persisted with started/complete/modified timestamp fields.


arkadyan (approved):
👍

PR #458, opened 2024-10-25
FEAT: Add tripID field to glides trip_updates ingestion

rymarczy:
Ingest new `tripId` field from Glides kinesis stream. This field is flattened into it's own column, so merging with existing datasets should cause no issues. 

Asana Task: https://app.asana.com/0/1205827492903547/1208481082941326


arkadyan (approved):
👍

PR #473, opened 2025-01-07
FIX: Upgrade pm_event_id coumns to BIGINT Type

rymarczy:
Staging started throwing errors because the pm_event_id column ran out of integer numbers.

This change updates the pm_event_id columns from int to bigint so we don't run out of keys.


arkadyan (approved):
👍

PR #475, opened 2025-02-12
Bus datetime conversion

ahu42:
stop_arrival_dt and stop_departure_dt are currently in UTC, but Ops Analytics would like them to be in Eastern Time for all flows and dashboards

Asana Task: [<asana_ticket_url>](https://app.asana.com/0/1205827492903547/1209312681633177)

### DIFF
--- a/.github/workflows/ci_python.yaml
+++ b/.github/workflows/ci_python.yaml
@@ -124,8 +124,10 @@ jobs:
           --cov-report term-missing \
           --cov-branch \
           --cov=lamp_py
+
       - name: Setup LCOV
         uses: hrishikesh-kadam/setup-lcov@v1
+        
       # Upload Coverage as an Artifact for Subsequent Jobs
       - name: Report code coverage
         uses: mbta/github-actions-report-lcov@v4
--- a/.github/workflows/ci_python.yaml
+++ b/.github/workflows/ci_python.yaml
@@ -129,6 +129,6 @@ jobs:
       - name: Report code coverage
         uses: mbta/github-actions-report-lcov@v4
         with:
-          coverage-files: coverage/lcov.*.info
+          coverage-files: coverage.info
           artifact-name: python-code-coverage
           github-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/ci_python.yaml
+++ b/.github/workflows/ci_python.yaml
@@ -126,11 +126,11 @@ jobs:
           --cov=lamp_py
 
       # Upload Coverage as an Artifact for Subsequent Jobs
-      - name: Setup LCOV
-        uses: hrishikesh-kadam/setup-lcov@v1
+      # - name: Setup LCOV
+      #   uses: hrishikesh-kadam/setup-lcov@v1
       - name: Report code coverage
         uses: mbta/github-actions-report-lcov@v4
         with:
           coverage-files: coverage.info
-          artifact-name: python-code-coverage
+          artifact-name: code-coverage-report
           github-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/ci_python.yaml
+++ b/.github/workflows/ci_python.yaml
@@ -125,13 +125,10 @@ jobs:
           --cov-branch \
           --cov=lamp_py
 
-      - name: Setup LCOV
-        uses: hrishikesh-kadam/setup-lcov@v1
-
       # Upload Coverage as an Artifact for Subsequent Jobs
       - name: Report code coverage
         uses: mbta/github-actions-report-lcov@v4
         with:
-          coverage-files: coverage/lcov.*.info
+          coverage-files: coverage.info
           artifact-name: python-code-coverage
           github-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/ci_python.yaml
+++ b/.github/workflows/ci_python.yaml
@@ -126,6 +126,8 @@ jobs:
           --cov=lamp_py
 
       # Upload Coverage as an Artifact for Subsequent Jobs
+      - name: Setup LCOV
+        uses: hrishikesh-kadam/setup-lcov@v1
       - name: Report code coverage
         uses: mbta/github-actions-report-lcov@v4
         with:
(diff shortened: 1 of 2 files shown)


src/lamp_py/tableau/jobs/bus_performance.py
```diff
@@ -76,7 +78,25 @@ def create_bus_parquet(job: HyperJob, num_files: Optional[int]) -> None:
...
+            arrow_table = pyarrow.Table.from_batches([batch])
+
+            polars_df = pl.from_arrow(arrow_table)
+            # convert to df if series
+            if isinstance(polars_df, pl.Series):
+                polars_df = polars_df.to_frame()
+            polars_df = polars_df.with_columns(
+                pl.col("stop_arrival_dt")
+                .dt.convert_time_zone(time_zone="US/Eastern")
+                .dt.replace_time_zone(None)
```
huangh, line 90: it looks like convert_time_zone gives it a timezone, so the replace_time_zone is unnecessary

ahu42: It is necessary bc I have to remove the tz marker and have it be a naive datetime otherwise it won't properly upload in Tableau. 
convert_timezone() is converting to Eastern & replace_time_zone() is removing that tz marker to be naive.

huangh: got it!

.github/workflows/ci_python.yaml
```diff
@@ -126,9 +126,11 @@ jobs:
...
 
       # Upload Coverage as an Artifact for Subsequent Jobs
+      - name: Setup LCOV
+        uses: hrishikesh-kadam/setup-lcov@v1
       - name: Report code coverage
-        uses: mbta/github-actions-report-lcov@v1
+        uses: mbta/github-actions-report-lcov@v4
         with:
-          coverage-files: coverage.info
+          coverage-files: coverage/lcov.*.info
```
rymarczy, line 134: This needs to be reverted.

src/lamp_py/tableau/jobs/bus_performance.py
```diff
@@ -7,6 +7,8 @@
 import pyarrow.dataset as pd
 from pyarrow.fs import S3FileSystem
 
+import polars as pl
```
rudiejd, line 10: woo, polars! 🐻‍❄️

.github/workflows/ci_python.yaml
```diff
@@ -126,8 +126,10 @@ jobs:
           --cov=lamp_py
 
       # Upload Coverage as an Artifact for Subsequent Jobs
+      - name: Setup LCOV
+        uses: hrishikesh-kadam/setup-lcov@v1
```
rymarczy, line 130: Can you try removing this and see if the `pytest` CI is successful? I can't see why would need this added based on the changes from v1 to v4 in the `mbta/github-actions-report-lcov` action.

src/lamp_py/tableau/jobs/bus_performance.py
```diff
@@ -76,7 +78,29 @@ def create_bus_parquet(job: HyperJob, num_files: Optional[int]) -> None:
 
     with pq.ParquetWriter(job.local_parquet_path, schema=job.parquet_schema) as writer:
         for batch in ds.to_batches(batch_size=500_000):
-            writer.write_batch(batch)
+            arrow_table = pyarrow.Table.from_batches([batch])
+
+            polars_df = pl.from_arrow(arrow_table)
```
rymarczy, line 83: `polars.from_arrow` can read in a RecordBatch or Table, so `batch` can be passed directly into this function and your `arrow_table` assignment can be removed.

src/lamp_py/tableau/jobs/bus_performance.py
```diff
@@ -76,7 +78,29 @@ def create_bus_parquet(job: HyperJob, num_files: Optional[int]) -> None:
...
+            # convert to df if series
+            if isinstance(polars_df, pl.Series):
+                polars_df = polars_df.to_frame()
```
rymarczy, line 86: I know this is a type guard, forced by mypy, but if `polars_df` does happen to be a `Series` instead of a `DataFrame`, something has gone horribly wrong, this if check should raise some kind of Exception instead of converting to a DataFrame.

src/lamp_py/tableau/jobs/bus_performance.py
```diff
@@ -76,7 +78,29 @@ def create_bus_parquet(job: HyperJob, num_files: Optional[int]) -> None:
...
+                .dt.replace_time_zone(None)
+                .alias("stop_departure_dt"),
+                pl.col("gtfs_travel_to_dt")
+                .dt.convert_time_zone(time_zone="US/Eastern")
+                .dt.replace_time_zone(None)
+                .alias("gtfs_travel_to_dt"),
+            )
+            table = polars_df.to_arrow()
+
+            writer.write_table(table)
```
rymarczy, line 103: nitpick: you could avoid this extra assignment and pass `.to_arrow()` directly into `.write_table()`

rymarczy (changes requested):
Looks good!

Just a couple requests.

src/lamp_py/tableau/jobs/bus_performance.py
```diff
@@ -76,7 +78,29 @@ def create_bus_parquet(job: HyperJob, num_files: Optional[int]) -> None:
...
+
+            polars_df = pl.from_arrow(arrow_table)
+            # convert to df if series
+            if isinstance(polars_df, pl.Series):
+                polars_df = polars_df.to_frame()
+            polars_df = polars_df.with_columns(
+                pl.col("stop_arrival_dt")
+                .dt.convert_time_zone(time_zone="US/Eastern")
+                .dt.replace_time_zone(None)
+                .alias("stop_arrival_dt"),
```
rymarczy, line 91: Nitpick: The `.alias()` calls could be dropped since you are not changing the column name.

.github/workflows/ci_python.yaml
rymarczy: Ok, looks like that setup IS needed! Thanks for checking, you can disregard this one.

rymarczy (approved):
Looks good! :cake: 

Make sure you squash-merge to close the PR. Try to provide a title and description for the squash-merge that will make the commit on `main` easy to distinguish. 

You can also add a new version-tag to trigger a deploy to prod so that Ops Analytics will be able to see the changes in Tableau.

PR #476, opened 2025-02-13
Bus dt conversion

ahu42:
Update seconds columns too
Asana Task: [<asana_ticket_url>](https://app.asana.com/0/1205827492903547/1209312681633177)


rymarczy (changes requested):
Could you please re-base this PR on the current `main` branch. The end result should be that only the commits related to this PR show up in the "Commits" tab.

PR #478, opened 2025-02-14
Add conversion for seconds after midnight cols

ahu42:
Add conversion for seconds cols as requested in ticket below:
Asana Task: [<asana_ticket_url>](https://app.asana.com/0/1189492770004753/1209312681633177)


rymarczy (approved):
Looks good!

One suggestion would be to add a comment that these conversions could be moved to the `bus_performance_manager` portion of the application if you end up touching that for other changes. 

Be sure to add a descriptive message to your squash-merge commit on main, explaining what changed with this PR.

ahu42:
these conversions could be moved to the bus_performance_manager portion of the application if you end up touching that for other changes.

ahu42:
these conversions could be moved to the bus_performance_manager portion of the application if we end up touching that for other changes.

PR #479, opened 2025-02-19
update opmi link & include Ops Analytics as well

ahu42:
Update OPMI link as per Heather's request and include Ops Analytics.


rymarczy (approved):
:cake:

PR #485, opened 2025-02-27
FEAT: Add TransitMaster fields to aid validation of GTFS vs Transitmaster bus data

huangh:
[Asana Task](https://app.asana.com/0/1189492770004753/1209391790454623)

Save off actual_departure_time, actual_arrival_time, and scheduled_time.

Trigger a regeneration of the whole bus_events datasets by bumping the version.

### DIFF
--- a/src/lamp_py/bus_performance_manager/events_joined.py
+++ b/src/lamp_py/bus_performance_manager/events_joined.py
@@ -217,8 +217,12 @@ def join_tm_to_rt(gtfs: pl.DataFrame, tm: pl.DataFrame) -> pl.DataFrame:
     Join gtfs-rt and transit master (tm) event dataframes
 
     :return added-columns:
+        tm_scheduled_time_dt -> Datetime
         tm_actual_arrival_dt -> Datetime
         tm_actual_departure_dt -> Datetime
+        tm_scheduled_time_sam -> Int64
+        tm_actual_arrival_time_sam -> Int64
+        tm_actual_departure_time_sam -> Int64
     """
 
     # join gtfs and tm datasets using "asof" strategy for stop_sequence columns
--- a/src/lamp_py/bus_performance_manager/events_joined.py
+++ b/src/lamp_py/bus_performance_manager/events_joined.py
@@ -217,8 +217,8 @@ def join_tm_to_rt(gtfs: pl.DataFrame, tm: pl.DataFrame) -> pl.DataFrame:
     Join gtfs-rt and transit master (tm) event dataframes
 
     :return added-columns:
-        tm_arrival_dt -> Datetime
-        tm_departure_dt -> Datetime
+        tm_actual_arrival_dt -> Datetime
+        tm_actual_departure_dt -> Datetime
     """
 
     # join gtfs and tm datasets using "asof" strategy for stop_sequence columns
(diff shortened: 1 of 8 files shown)


src/lamp_py/bus_performance_manager/events_joined.py
```diff
@@ -217,8 +217,8 @@ def join_tm_to_rt(gtfs: pl.DataFrame, tm: pl.DataFrame) -> pl.DataFrame:
     Join gtfs-rt and transit master (tm) event dataframes
 
     :return added-columns:
-        tm_arrival_dt -> Datetime
-        tm_departure_dt -> Datetime
+        tm_actual_arrival_dt -> Datetime
+        tm_actual_departure_dt -> Datetime
```
rymarczy, line 221: also adding:

-tm_actual_arrival_dt
-tm_actual_arrival_time_sam
-tm_actual_departure_time_sam
-tm_scheduled_time_sam

src/lamp_py/bus_performance_manager/events_metrics.py
```diff
@@ -126,12 +130,14 @@ def bus_performance_metrics(service_date: date, gtfs_files: List[str], tm_files:
...
+        # drop temp fields and dev validation fields
         .drop(
             [
                 "gtfs_departure_dt",
                 "gtfs_arrival_dt",
-                "tm_departure_dt",
-                "tm_arrival_dt",
+                "tm_actual_arrival_time_sam",
+                "tm_actual_departure_time_sam",
+                "tm_scheduled_time_sam",
```
rymarczy, line 140: Do you want to drop these here? they won't make it into the tableau dataset if they're dropped...

rymarczy (changes requested):
Few requests/questions. 

To get these changes into tableau we'll also need a change to the `tableau_publisher` job. I'm assuming that will be a separate PR?

huangh (commented):
i hadn't gotten to it, but i think i'll make the tableau publisher changes in this PR as well

src/lamp_py/bus_performance_manager/events_metrics.py
huangh: the fields we want to keep are the service date + seconds after midnight as _dt (EST) per the discussion yesterday. so I thought i would just drop the seconds after midnight ones. i guess i can future proof by providing both so we can validate against the raw SAM values in analysis.

src/lamp_py/bus_performance_manager/events_tm.py
```diff
@@ -116,6 +121,7 @@ def generate_tm_events(tm_files: List[str]) -> pl.DataFrame:
                 & pl.col("GEO_NODE_ID").is_not_null()
                 & pl.col("TRIP_ID").is_not_null()
                 & pl.col("VEHICLE_ID").is_not_null()
+                & pl.col("SCHEDULED_TIME").is_not_null()
```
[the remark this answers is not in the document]
huangh: this makes sense

rymarczy (approved):
This looks good to me. As long as it runs on staging without error, should be good to push to prod. 

Tableau Publisher might throw some errors while the `bus_events` datasets are being re-processed. I'm not totally sure how that will work...

PR #490, opened 2025-03-24
FIX: Convert Times back to strings to handle when times are greater than midnight

huangh:
[asana_ticket](https://app.asana.com/0/1205827492903547/1209285127534399/f)

Times are in String HH:MM:SS format, from glides https://mbta.github.io/schemas/events/glides/com.mbta.ctd.glides.trips_updated.v1#time

This is not valid when converted to hyper times. Those have a valid range of 00:00:00 - 24:00:00
https://tableau.github.io/hyper-db/docs/sql/datatype/datetime

Returning back to strings to match existing handling.


rymarczy (approved):
:cake:

rymarczy (approved):

PR #491, opened 2025-03-25
FIX: Add Glides project_name to hyperjobs to write them to the correct project

huangh:
Currently using default which sends them to a OPMI project


rymarczy (approved):
:cake:

PR #492, opened 2025-03-26
FIX: Glides -> LAMP Data for the project to put glides*.hyper on tableau server

huangh:
change tableau project from "Glides" to "LAMP Data"


rymarczy:
@huangh I think the original intent for this change has resolved, correct?

huangh:
did not need - permissions issue on tableau server

PR #499, opened 2025-04-16
(temp) light rail gps is failing ingestion. this is unused, disabling for now

huangh:
light rail gps is failing and crashing ingestion. temporary disable while we investigate and triage. 

nobody uses this yet, so it doesn't hurt anyone


rymarczy (approved):
:cake:

PR #505, opened 2025-04-17
(fix) update polars types to match database types for vehicle_trips 

huangh:
Cleanup and fixes for the polars reimplementation of database queries for `backup_rt_static_trip_match`

We were unable to test these changes until getting the previous PR to staging. These changes fix the remaining issues found. 

Mostly datatype issues, one was an issue with polars -> pandas converting a column to float.

### DIFF
--- a/src/lamp_py/ingestion/convert_gtfs_rt.py
+++ b/src/lamp_py/ingestion/convert_gtfs_rt.py
@@ -407,9 +407,6 @@ class GtfsRtConverter(Converter):
         :param table: pyarrow Table
         :param local_path: path to local parquet file
         """
-        logger = ProcessLogger("write_local_pq", local_path=local_path, table_rows=table.num_rows)
-        logger.log_start()
-
         out_ds = self.make_hash_dataset(table, local_path)
 
         unique_ts_min = pc.min(table.column("feed_timestamp")).as_py() - (60 * 45)
@@ -428,14 +425,10 @@ class GtfsRtConverter(Converter):
                 out_ds.to_table(columns=[self.detail.partition_column]).column(self.detail.partition_column)
             )
             for part in partitions:
-                logger.add_metadata(table_part=str(part), part_rows=0, part_mbs=0)
                 write_table = out_ds.to_table(
                     filter=(
                         (pc.field(self.detail.partition_column) == part) & (pc.field("feed_timestamp") < unique_ts_min)
                     )
-                )
-                logger.add_metadata(
-                    part_rows=write_table.num_rows, part_mbs=round(write_table.nbytes / (1024 * 1024), 2)
                 )
                 hash_writer.write_table(write_table)
                 upload_writer.write_table(write_table.drop_columns(GTFS_RT_HASH_COL))
@@ -453,9 +446,6 @@ class GtfsRtConverter(Converter):
                     .to_arrow()
                     .cast(out_ds.schema)
                 )
-                logger.add_metadata(
-                    part_rows=write_table.num_rows, part_mbs=round(write_table.nbytes / (1024 * 1024), 2)
-                )
                 hash_writer.write_table(write_table)
                 upload_writer.write_table(write_table.drop_columns(GTFS_RT_HASH_COL))
 
@@ -467,8 +457,6 @@ class GtfsRtConverter(Converter):
                 upload_path,
                 local_path.replace(self.tmp_folder, S3_SPRINGBOARD),
             )
-
-        logger.log_complete()
 
     # pylint: enable=R0914
 
--- a/src/lamp_py/performance_manager/l1_cte_statements.py
+++ b/src/lamp_py/performance_manager/l1_cte_statements.py
@@ -1,4 +1,5 @@
 from datetime import datetime
+import os
 import sqlalchemy as sa
 import polars as pl
 from sqlalchemy.sql.functions import rank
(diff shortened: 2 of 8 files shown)


src/lamp_py/postgres/postgres_utils.py
```diff
@@ -312,6 +313,33 @@ def execute(
 
         return result  # type: ignore
 
+    def execute_with_data_pl(
```
huangh, line 316: I know this split is dumb, but i was unwilling to wrangle with mypy and get it to properly accept polars.to_dicts() vs pandas.to_dict(orient="records"))

If we really hate this I can try to just merge the two back and check the type inside the call? or change the type of data to be list:dict instead...this might be the way. Thoughts?

rymarczy: I think this is fine.

rymarczy (changes requested):
Let's add back all of the test you disabled in PR 500. As they would have caught this error. 

You'll have to pull out part of the `static_prefix` var from the `static_trips_subquery_pl` into a file declared variable. 

Then you can add a `unittest.mock @patch` decorator to change the variable to point to the https endpiont for testing in github:

```
GTFS_ARCHIVE = "s3://mbta-performance/lamp/gtfs_archive"

def static_trips_subquery_pl(service_date: int) -> pl.DataFrame:
    """
    Polars implementation of static_trips_subquery in backup_rt_static_trip_match
    """
    service_year = str(service_date)[:4]
    static_prefix = f"{GTFS_ARCHIVE}/{service_year}"
```

test files:
```
@patch("lamp_py.performance_manager.l1_cte_statements.GTFS_ARCHIVE","https://performancedata.mbta.com/lamp/gtfs_archive")
def test_vp_only(...
```

src/lamp_py/performance_manager/l1_cte_statements.py
```diff
@@ -137,7 +137,9 @@ def apply_branch_route_id(series_list) -> str:  # type: ignore
     static_trips = static_trips.with_columns(
         pl.duration(hours=pl.col("hour"), minutes=pl.col("minute"), seconds=pl.col("second"))
         .dt.total_seconds()
-        .alias("static_start_time")
```
rymarczy, line 140: nit: these cast changes are not required now that you're never converting to a pandas dataframe

src/lamp_py/postgres/postgres_utils.py
```diff
@@ -312,6 +313,33 @@ def execute(
...
+        self,
+        statement: Union[
+            sa.sql.selectable.Select,
+            sa.sql.dml.Update,
+            sa.sql.dml.Delete,
+            sa.sql.dml.Insert,
+        ],
+        data: pl.DataFrame,
+        disable_trip_tigger: bool = False,
```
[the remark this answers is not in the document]
huangh: i'm going to keep the parameter so it lines up with pandas implementation, and pass in True to this one. Or do we want to switch the default to true for both methods? 

I just don't think they should diverge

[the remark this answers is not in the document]
huangh: the pandas execute_with_data makes use of the trip=False once in our codebase, for TempEventCompare. So I'll default them both to True and flip that one instead.

src/lamp_py/performance_manager/l1_rt_trips.py
```diff
@@ -1016,11 +1028,12 @@ def backup_rt_static_trip_match(
         )
     )
 
-    db_manager.execute_with_data(
+    db_manager.execute_with_data_pl(
         update_query,
-        backup_trips_match_df.to_pandas(),  # we'll have to clean up the execute_with_data method
+        backup_trips_match_df,  # we'll have to clean up the execute_with_data method
```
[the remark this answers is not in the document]
huangh: yes

src/lamp_py/postgres/postgres_utils.py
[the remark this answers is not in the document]
rymarczy: I think the default `False` value is only used once, and every other invocation gets called with `True` so `True` is definitely the default. 

The triggers are a hold over from when DB triggers were used to update some fields. But they're extremely expensive and opaque operations and most of that functionality has been moved out of the DB.

src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -407,9 +407,6 @@ def write_local_pq(self, table: pyarrow.Table, local_path: str) -> None:
         :param table: pyarrow Table
         :param local_path: path to local parquet file
         """
-        logger = ProcessLogger("write_local_pq", local_path=local_path, table_rows=table.num_rows)
```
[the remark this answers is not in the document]
huangh: just cleanup logging. this was removed

src/lamp_py/performance_manager/l1_cte_statements.py
huangh: are we sure? i just wanted to keep all the dtypes lined up with what was in the database schema. i think total_seconds() returns i64, which is why i cast it here

src/lamp_py/postgres/postgres_utils.py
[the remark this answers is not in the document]
huangh: I think my changes satisfy this now

src/lamp_py/performance_manager/l1_cte_statements.py
rymarczy: Yeah, the `to_dict()` call on a polars frame just converts the values into python types, so `pl.Int32`, `pl.Int16`, `pl.Int64` all just go to a python `int`

rymarczy: Your issue before with the `to_pandas()` call had to do with pandas not correctly inferring the polars type columns and converting the ints to floats.

rymarczy (changes requested):
Can you also please enable the tests in `/lamp/tests/performance_manager/test_performance_manager.py`, except for `test_whole_table`. 

Also fix lint failure.

src/lamp_py/performance_manager/l1_cte_statements.py
huangh: got it. I'm going to stick a comment in to this effect, but I'd like to leave the casts in place to let us operate on these values as if they were in the database, to catch any issues with types in the future.

tests/performance_manager/test_performance_manager.py
```diff
@@ -601,6 +603,9 @@ def test_gtfs_rt_processing(
 
 
 @pytest.mark.skip(reason="Temporarily disabled: PR #500")
```
rymarczy, line 605: I don't understand why we're still skipping these, they should pass with the `@patch`.

huangh: do you want me to figure it out in this ticket?

rymarczy: Yeah, those tests run for me locally. The service date being in 2023 shouldn't matter. The GTFS archive has GTFS data available going back to 2009.

rymarczy (approved):
:cake:

PR #507, opened 2025-04-22
(fix) tableau output fixes - datatypes, archive bucket, and sort order

huangh:
Various fixes

* Float types not being properly converted to hyper SqlTypes - error observed when trying to convert to hyper. 
* Output dev-green data to internal archive instead of public bucket - [Ticket](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1210039594771502?focus=true)
* Fix sort order of listed files to get the right files into the "last-N-days" - logical fix

### DIFF
--- a/src/lamp_py/tableau/hyper.py
+++ b/src/lamp_py/tableau/hyper.py
@@ -95,14 +95,16 @@ class HyperJob(ABC):  # pylint: disable=R0902
         dtype = str(dtype)
         dtype_map = {
             "int8": SqlType.small_int(),
+            "uint8": SqlType.small_int(),
             "int16": SqlType.small_int(),
+            "uint16": SqlType.int(),
             "int32": SqlType.int(),
             "uint32": SqlType.big_int(),
             "int64": SqlType.big_int(),
             "bool": SqlType.bool(),
-            "float16": SqlType.double(),
-            "float32": SqlType.double(),
-            "float64": SqlType.double(),
+            "halffloat": SqlType.double(),
+            "float": SqlType.double(),
+            "double": SqlType.double(),
         }
 
         map_check = dtype_map.get(dtype)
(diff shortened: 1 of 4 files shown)


src/lamp_py/tableau/hyper.py
```diff
@@ -95,14 +95,16 @@ def convert_parquet_dtype(self, dtype: pyarrow.DataType) -> SqlType:
         dtype = str(dtype)
         dtype_map = {
             "int8": SqlType.small_int(),
+            "uint8": SqlType.small_int(),
```
rymarczy, line 98: I believe `SqlType.small_int()` is a 2 bytes signed integer, which could be overflowed by an 8bit unsigned integer. I think this has to move to an `SqlType.int()`

src/lamp_py/tableau/jobs/filtered_hyper.py
```diff
@@ -106,7 +107,15 @@ def create_tableau_parquet(self, num_files: Optional[int]) -> None:
...
             bucket_name=self.remote_input_location.bucket, file_prefix=file_prefix, in_filter=self.object_filter
         )
+
         ds_paths = [s.replace("s3://", "") for s in s3_uris]
+
+        # s3 list returns in lexicographical order,
+        # so month=4/day=4 comes after month=4/day=30. This sort grabs the last part,
+        # e.g. 2025-04-22T00:00:00.parquet as the sort key, and re-orders by that instead
+        # to get it in date order
+        ds_paths = sorted(ds_paths, key=lambda x: os.path.split(x)[1])
```
rymarczy, line 117: suggestion:  `os.path.basename()` would avoid having to index this result.

rymarczy (changes requested):
One change.

src/lamp_py/tableau/hyper.py
huangh: 2^8 max is 256, small_int is 2 byte range is -2^16 - 2^16-1, yes? i think there's no problem here

rymarczy: ha, whoops, 2 bytes = 16 bits :melting_face:

PR #508, opened 2025-04-23
(fix) backfill rt rail performance manager data from 4/4 to 4/22

huangh:
backfill process involves resetting flags in the metadata db and clearing out data from the performance_manager db. 

these have been verified to work as expected via SELECT... on the staging db.


rudiejd (approved):

src/lamp_py/migrations/versions/metadata_staging/003_5e3066f113ff_backfill_rt_rail_2025_04_04_to_2025_04_18.py
rymarczy, line 1: It would be a bit safer to do this in the `performance_manager` migration. Otherwise there is potentially a race condition where this migration runs, while an old deploy of performance_manager is still in operation and begins loading these files before the `performance_manager` migration.

src/lamp_py/migrations/versions/metadata_staging/003_5e3066f113ff_backfill_rt_rail_2025_04_04_to_2025_04_18.py
```diff
@@ -0,0 +1,64 @@
...
+
+    update_query = """
+    UPDATE
+        metadata_log
+    SET
+        rail_pm_process_fail = false
+        , rail_pm_processed = false
+    WHERE
+        created_on > '2025-04-04 00:00:00'
+        and created_on < '2024-04-22 23:59:59'
```
rymarczy, line 53: Would be cleaner to craft a where clause that selects the exact files you want to re-process based on the `path` values you want to re-load. 

something like 

```
substring(path, '\d{4}-\d{2}-\d{2}')::date >= '2025-04-05'
and substring(path, '\d{4}-\d{2}-\d{2}')::date <= '2025-04-22'
```

src/lamp_py/migrations/versions/metadata_staging/003_5e3066f113ff_backfill_rt_rail_2025_04_04_to_2025_04_18.py
huangh: I was running into issues where in the CI environment wouldn't recognize the metadata table when I tried to run the query in the performance manager directory. I see that it did work before, but wasn't sure where/how the CI database got seeded with schemas etc.

src/lamp_py/migrations/versions/metadata_staging/003_5e3066f113ff_backfill_rt_rail_2025_04_04_to_2025_04_18.py
huangh: while there could be differences between paths and created_on date, i think my range successfully gathers all the suspect days, so i think it looks clearer as is. 

Is it not recommended to use created_on?

rymarczy: I would say it's not recommended because you're not actually targeting records by `created_on` you're targeting specific file paths, so the query should reflect that. `created_on` could also vary between environments, and by targeting that actual path names you'd have a query that is totally environment agnostic.

src/lamp_py/migrations/versions/metadata_staging/003_5e3066f113ff_backfill_rt_rail_2025_04_04_to_2025_04_18.py
rymarczy: You can throw the metadata calls in a try-except block that will catch for CI where the metadata table doesn't exist, something similar is done here: https://github.com/mbta/lamp/blob/6d7c4be2193a3ff5b6d8f5e889da04403c60e8f4/src/lamp_py/migrations/versions/metadata_prod/001_07903947aabe_initial_changes.py#L71

huangh: incorporated

src/lamp_py/migrations/versions/metadata_staging/003_5e3066f113ff_backfill_rt_rail_2025_04_04_to_2025_04_18.py
huangh: i think in this instance it is safe as is, based on what we know of the behavior of the performance degradation incident. also there's a day buffer at the end as we're handling the duplicated dev-green data in here as well, when the queries were back to running at speed.

rymarczy (approved):
Looks good to send to staging and monitor before a deploy to prod. 

:cake:

PR #509, opened 2025-04-24
(feat) add date range based s3 list query

huangh:
[ticket](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1210046905852611?focus=true)

add capability to specifically query a date range. Wraps the existing list functionality with a helper method that constructs the formatted date string, and then loops over the paths on s3. 

useful to solve issue where s3 paths are in lexicographical order not date order.


src/lamp_py/tableau/jobs/filtered_hyper.py
```diff
@@ -92,39 +86,34 @@ def update_parquet(self, _: None) -> bool:
...
         Join files into single parquet file for upload to Tableau. apply filter and conversions as necessary
         """
-        if self.bucket_filter is not None:
-            file_prefix = os.path.join(self.remote_input_location.prefix, self.bucket_filter)
-        else:
-            file_prefix = self.remote_input_location.prefix
 
-        s3_uris = file_list_from_s3(
-            bucket_name=self.remote_input_location.bucket, file_prefix=file_prefix, in_filter=self.object_filter
+        end_date = datetime.now()
```
rymarczy, line 97: This is a pretty complicated solution for this problem. My understanding is that the basic issue here is the very old parquet files in the SPRINGBOARD partitions are messing with the standard date formatting. 

However this process is never really expected to use those very old parquet files, and should only be operating with recent files that follow the format of ending with `YYYY-MM-DDT00:00:00.parquet`. 

In that case, this two liner should solve it:
```
import re
import os
date_re = re.compile(r"\d{4}-\d{2}-\d{2}T00:00:00\.parquet")
s3_uris = sorted([s for s in s3_uris if date_re.search(s)], key=lambda x: os.path.basename(x))
```

rymarczy (approved):
One pretty big suggestion...

src/lamp_py/tableau/jobs/filtered_hyper.py
huangh: I think having the ability to arbitrarily select date ranges is something we can't currently do well and is something I've found to hinder work already. I think my solution here is a small piece of a long term effort to move the code base towards more explicit pinpointing of inputs and outputs across the app. 

The meat of this is really like 10 lines of code, and I think we agree it does a fairly mundane, concrete task, so I wouldn't classify it as complicated. 

Regex's though 😅

PR #510, opened 2025-04-24
(fix) fix migration query for both staging and prod - bump staging uuid

huangh:
typo in query (date < 2024) resulted in no change in metadata table. correcting that and other typo. verified in select statement

### DIFF
--- a/src/lamp_py/migrations/versions/performance_manager_staging/011_5e3066f113ff_backfill_rt_rail_2025_04_04_to_2025_04_18.py
+++ b/src/lamp_py/migrations/versions/performance_manager_staging/011_5e3066f113ff_backfill_rt_rail_2025_04_04_to_2025_04_18.py
@@ -47,11 +47,19 @@ def upgrade() -> None:
     op.execute(clear_trips)
 
     # Query to Check
-    # SELECT created_on, rail_pm_processed, rail_pm_process_fail
+    # SELECT
+    #     created_on,
+    #     rail_pm_process_fail,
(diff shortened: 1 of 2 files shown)


src/lamp_py/migrations/versions/performance_manager_staging/011_5e3066f113ff_backfill_rt_rail_2025_04_04_to_2025_04_18.py
```diff
@@ -47,11 +47,19 @@ def upgrade() -> None:
     op.execute(clear_trips)
 
     # Query to Check
-    # SELECT created_on, rail_pm_processed, rail_pm_process_fail
```
rymarczy, line 50: It would be quite confusing having two migrations doing the exact same thing back to back. If you're going to modify this one, could you turn it into a no-op with a comment that it failed on the original run and was replaced by the next migration?

rymarczy (approved):
One suggestion.

src/lamp_py/migrations/versions/performance_manager_staging/011_5e3066f113ff_backfill_rt_rail_2025_04_04_to_2025_04_18.py
huangh: Noted - gonna send this through to see if its working as expected, and address this in the follow-on to when I bump versions to get the downstream outputs to regenerate.

PR #513, opened 2025-04-28
(feat/mem) set batch_readahead and fragment_readahead for all intensive to_batches() calls

huangh:
Rail Performance Manager was dangerously close to OOM, and fell over once in the time window I was observing. Setting these params reduce memory pressure significantly.


rymarczy (approved):
Might want to keep an eye on if these changes slow down any of these operations significantly. 

:cake:

PR #514, opened 2025-04-29
(feat) Add new output for prod GTFS-RT to tableau

huangh:
[ticket](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1209347095723228?focus=true)

follow on with additional requirements from stakeholders

* add some error handling for empty datasets
* shift the rolling window to today() - 1 day to eliminate OSError contention with ingestion and perf_mgr reading the same file
* add heavyrail filters and outputs to tableau

### DIFF
--- a/src/lamp_py/tableau/jobs/filtered_hyper.py
+++ b/src/lamp_py/tableau/jobs/filtered_hyper.py
@@ -74,7 +74,6 @@ class FilteredHyperJob(HyperJob):
             self.create_tableau_parquet(num_days=self.rollup_num_days)
             self.first_run = False
             return True
-
         # only run once per day after 11AM UTC
         if object_exists(self.remote_input_location.s3_uri):
             now_utc = datetime.now(tz=timezone.utc)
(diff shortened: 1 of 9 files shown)


src/lamp_py/tableau/jobs/filtered_hyper.py
```diff
@@ -115,6 +114,9 @@ def create_tableau_parquet(self, num_days: Optional[int]) -> None:
         )
         process_logger = ProcessLogger("filtered_hyper_create", num_days=num_days)
         process_logger.log_start()
+        if len(ds_paths) == 0:
+            process_logger.add_metadata(n_paths_zero=len(ds_paths))
+            return
```
rymarczy, line 119: I'm pretty sure the way you have this configured, the process is expecting that you'll create a parquet file that will be uploaded by this process. By returning early, and not creating a local parquet file for upload, this will throw an error in the `HyperJob` framework. 

If you will potentially not create a file, with your `self.create_tableau_parquet` method, you should return a bool indicating whether a file was created or not and pass the bool back to be returned by `self.update_parquet`.

rymarczy (commented):
Not sure your empty dataset solution will work.

src/lamp_py/tableau/jobs/filtered_hyper.py
huangh: addressed

PR #515, opened 2025-05-01
(fix) rerun migration 4/22 for missing data

huangh:
Previous migration filter excluded the last day in question (4/22) due to matching on created date, which was 4/23, past the end of the filter. 

Reset flags in metadata db to rerun 4/22 and 4/23 to account for EST/UTC overlap. 

Bump versions in flat_file and rt_rail to regenerate output products

### DIFF
--- a/src/lamp_py/performance_manager/flat_file.py
+++ b/src/lamp_py/performance_manager/flat_file.py
@@ -43,7 +43,7 @@ class S3Archive:
     RAIL_PERFORMANCE_PREFIX = os.path.join(LAMP, "subway-on-time-performance-v1")
     INDEX_FILENAME = "index.csv"
     VERSION_KEY = "rpm_version"
-    RPM_VERSION = "1.2.1"
+    RPM_VERSION = "1.2.2"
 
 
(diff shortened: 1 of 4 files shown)


src/lamp_py/performance_manager/flat_file.py
```diff
@@ -43,7 +43,7 @@ class S3Archive:
     RAIL_PERFORMANCE_PREFIX = os.path.join(LAMP, "subway-on-time-performance-v1")
     INDEX_FILENAME = "index.csv"
     VERSION_KEY = "rpm_version"
-    RPM_VERSION = "1.2.1"
```
rymarczy, line 46: suggestion: do these version bumps in another PR, after you've confirmed the migration was successful and at a time that won't interfere with weekly dashboard updates for stakeholders.

rymarczy (approved):
One suggestion.

src/lamp_py/performance_manager/flat_file.py
huangh: I figured i'd try 😅

rymarczy (approved):
:cake:

PR #518, opened 2025-05-01
(feat) various cleanup tasks and analysis tools - no deployed impact

huangh:
* remove impl of failed migration - add comment stating migration did not occur
* add some analysis tools - to be improved
* add migration template generator - to be improved
* add myself to authors


rymarczy (approved):
:cake:

PR #519, opened 2025-05-07
(fix) bus performance manager stop_arrival_dt and stop_departure_dt logic regressions and various fixes

huangh:
[Asana Ticket](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1209939714274586?focus=true)

* The existing logic did not properly fill out when either TM date or GTFS date were available. added logic to populate whenever either is available. Added tests to check this logic works for the last stop on 20 sample trips on 5/5/25. This resolves the above ticket.

* Fixed a related issue where UTC timestamps were being compared to EST timestamps for these time fields. This was introduced in #485. timezones were added to the columns to eliminate this class of errors via typing. Added timezones test to verify this works as expected

* Fixed issue where missing TM files were inadvertently causing assertion errors and spamming our alerts. Handle these cases (when no data available) This was thought to be related at first but turned out to be not root cause


src/lamp_py/bus_performance_manager/events_gtfs_rt.py
```diff
@@ -223,6 +223,10 @@ def positions_to_events(vehicle_positions: pl.DataFrame) -> pl.DataFrame:
                 "IN_TRANSIT_TO": "gtfs_travel_to_dt",
             }
         )
+        .with_columns(
```
rymarczy, line 226: nit: this could be combined with the next `.with_columns()` call, my rule of thumb is to only chain a new call if a subsequent call needs to reference the new column you are creating.

src/lamp_py/bus_performance_manager/write_events.py
```diff
@@ -36,6 +41,12 @@ def write_bus_metrics() -> None:
             day_logger.log_failure(FileNotFoundError(f"No RT_VEHICLE_POSITION files found for {service_date}"))
             continue
 
+        # need gtfs_rt files to run process - this happens regularly at ~12am, so this
+        # is an expected case.
+        if len(tm_files) == 0:
```
rymarczy, line 46: This process will work with `tm_files == 0` I don't think this is protecting anything.

src/lamp_py/bus_performance_manager/write_events.py
```diff
@@ -50,8 +61,18 @@ def write_bus_metrics() -> None:
                     extra_args={"Metadata": {VERSION_KEY: bus_events.version}},
                 )
 
-            day_logger.log_complete()
+            # if any day succeeds, flip true - triggers upload to tableau
+            successful_metrics = True
+
+        except (LampExpectedNotFoundError, LampInvalidProcessingError):
```
rymarczy, line 67: The way these exceptions are currently thrown, `LampInvalidProcessingError` means something unexpected has gone wrong with the process. This should not be continued on, the error should be logged.

src/lamp_py/bus_performance_manager/pipeline.py
```diff
@@ -49,8 +49,8 @@ def iteration() -> None:
         process_logger = ProcessLogger("event_loop")
         process_logger.log_start()
         try:
-            write_bus_metrics()
-            start_bus_parquet_updates()
+            if write_bus_metrics():
```
rymarczy, line 52: Returning `bool` from here is necessary. 

`write_bus_metrics` can create parquet files for multiple days. With this change, an error on one day will preclude valid updates from another day being reflected in the Tableau datasets. 

`continue` or Excepting out of a specific `service_date` in `write_bus_metrics` is sufficient protection from polluting the export data.

src/lamp_py/bus_performance_manager/events_joined.py
```diff
@@ -147,7 +148,12 @@ def join_schedule_to_rt(gtfs: pl.DataFrame) -> pl.DataFrame:
         plan_direction_destination_headway_seconds -> Int64
     """
     service_dates = gtfs.get_column("service_date").unique()
-    assert len(service_dates) == 1, f"more than 1 service_date found: {service_dates}"
+
+    if len(service_dates) == 0:
```
rymarczy, line 152: suggestion: Feels a little odd to raise here on a `join` function. 

Could be slightly more concise to raise in `events_gtfs_rt.generate_gtfs_rt_events()` and claify in the doc string that this function should always return results.

rymarczy (changes requested):
A few things, also looks like tests need a little cleanup.

src/lamp_py/bus_performance_manager/events_joined.py
huangh: will address later

huangh: https://app.asana.com/1/15492006741476/project/1189492770004753/task/1210224824486631?focus=true

tests/utils/timezones.py
huangh, line 1: test to ensure timezone types enforce as expected. this proves out adding types to the dataframes is worth it

src/lamp_py/tableau/conversions/convert_bus_performance_data.py
huangh, line 1: do the final conversion for EST timezone at the end, right before export

src/lamp_py/bus_performance_manager/events_tm.py
```diff
@@ -49,9 +49,9 @@ def generate_tm_events(tm_files: List[str]) -> pl.DataFrame:
         trip_id -> String
         stop_id -> String
         tm_stop_sequence -> Int64
-        tm_scheduled_time_dt -> Datetime(time_unit='us', time_zone=None) as EST
-        tm_actual_arrival_dt -> Datetime(time_unit='us', time_zone=None) as EST
-        tm_actual_departure_dt -> Datetime(time_unit='us', time_zone=None) as EST
+        tm_scheduled_time_dt -> Datetime(time_unit='us', time_zone="UTC")
+        tm_actual_arrival_dt -> Datetime(time_unit='us', time_zone="UTC")
```
huangh, line 53: this was the bug - we ended up subtracting EST times from UTC times in events_metrics - adding timezones to everything so this won't happen again

rymarczy (approved):
:1st_place_medal:

PR #521, opened 2025-05-14
(fix) correct bug where bus plan times were incorrectly calculated

huangh:
[asana](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1210058693545796?focus=true)


rymarczy (approved):
Nice catch!

:pie:

rymarczy (approved):
:motor_boat:

PR #522, opened 2025-05-22
(fix) adjust tableau upload order for operational data

huangh:
[Ticket](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1210342783044783?focus=true)

### DIFF
--- a/src/lamp_py/tableau/jobs/filtered_hyper.py
+++ b/src/lamp_py/tableau/jobs/filtered_hyper.py
@@ -47,14 +47,6 @@ class FilteredHyperJob(HyperJob):
         self.parquet_filter = parquet_filter  # level 2 | by column and simple filter
         self.dataframe_filter = dataframe_filter  # level 3 | complex filter
 
-        # this flag is entirely a developer nice to have. This will ensure that
-        # the hyper job will run immediately when the tableau job is called, and will
-        # process and upload a hyper file now, rather than on the hour or after 7am
-        # this relies on the FilteredHyperJob persisting across runs - currently it is
-        # constructed on library load, but if it is reconstructed on each run_hyper() invocation,
-        # this will no longer hold.
-        self.first_run = True
-
     @property
     def parquet_schema(self) -> pyarrow.schema:
         return self.processed_schema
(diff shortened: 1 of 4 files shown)


src/lamp_py/tableau/jobs/glides.py
```diff
@@ -190,15 +190,20 @@ def create_parquet(self, _: None) -> None:
         self.update_parquet(None)
 
     def update_parquet(self, _: None) -> bool:
-        # only run once per day after 11AM UTC - 6 or 7 AM EST
+        if self.first_run:  # type: ignore
```
rymarczy, line 193: This Class is newly instantiated on every call to `start_parquet_updates` so this will always be True??

src/lamp_py/tableau/jobs/glides.py
```diff
@@ -190,15 +190,20 @@ def create_parquet(self, _: None) -> None:
...
+            create_trips_updated_glides_parquet(self, None)
+            return True
+
+        # only run once per day after 10AM UTC - 5 or 6 AM EST
         if object_exists(tableau_glides_all_trips_updated.s3_uri):
             now_utc = datetime.now(tz=timezone.utc)
             last_mod: datetime = file_list_from_s3_with_details(
                 bucket_name=tableau_glides_all_trips_updated.bucket,
                 file_prefix=tableau_glides_all_trips_updated.prefix,
             )[0]["last_modified"]
```
rymarczy, line 204: If the goal is to get this to run on every call to `start_parquet_updates` it makes much more sense to delete, or at least comment out, this logic that is attempting to only run this job once per day. 

Having unreachable code will be very confusing for anyone having to update this in the future.

src/lamp_py/tableau/jobs/filtered_hyper.py
```diff
@@ -70,7 +62,7 @@ def update_parquet(self, _: None) -> bool:
         # this relies on the FilteredHyperJob persisting across runs - currently it is
         # constructed on library load, but if it is reconstructed on each run_hyper() invocation,
         # this will no longer hold.
-        if self.first_run:
+        if self.first_run:  # type: ignore
```
rymarczy, line 65: suggestion: having to make mypy ignore this simple bool variable is very odd, and likely points to some other issue with the data structure. I'd make all attempts to remove this, if possible.

rymarczy (changes requested):
Just need to do something with the unreachable portion of `update_parquet`.

rymarczy (approved):
:cake:

PR #523, opened 2025-05-27
(feat) add additional filters for heavy rail

huangh:
[Asana Task](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1210328317084048?focus=true)

* make filter for heavy rail more robust. refactor to grab stops dynamically, and move hardcodes to unit test
* tested that all 15 stop_ids now are returned in test parquet file. the additional "exit only" stops are not listed in the returned result, matching expected result of ticket
* (extra) move utils from bus to generic utils - fix fallout 

todo: follow-on..do the same cleanup for light rail filters


src/lamp_py/tableau/conversions/convert_gtfs_rt_trip_updates.py
```diff
@@ -69,10 +69,10 @@ def heavyrail(polars_df: pl.DataFrame) -> pl.DataFrame:
     """
     Function to apply final conversions to lamp data before outputting for tableau consumption
     """
-    terminal_stop_ids = list(map(str, [70001, 70036, 70038, 70059, 70061, 70094, 70105]))
+
     polars_df = polars_df.filter(
         ~pl.col("trip_update.stop_time_update.departure.time").is_null()
-        & pl.col("trip_update.stop_time_update.stop_id").is_in(terminal_stop_ids)
+        & pl.col("trip_update.stop_time_update.stop_id").is_in(HeavyRailFilter.terminal_stop_ids)
```
rymarczy, line 75: `HeavyRailFilter.terminal_stop_ids` is a method now.

rymarczy: Using the class in this way will also be pretty expensive. A new `HeavyRailFilter` object will be created on every call to this function, which includes a call to read in a remote parquet file. While functional this is an anti-pattern.

src/lamp_py/utils/filter_bank.py
```diff
@@ -1,5 +1,77 @@
...
+        When called, dynamically populates terminal_stop_ids based on the
+        _terminal_stop_place_names. This has to live outside of the constructor
+        because we want to be able to patch the GTFS_ARCHIVE_FILTER_BANK for
+        unit tests (s3 path vs public performancedata path) which is not possible
+        when terminal_stop_ids is populated on class construction.
+        """
+        heavy_rail_filter = pl.col("vehicle_type") == 1
+
+        service_date = datetime.now()
+        stops = pl.read_parquet(f"{GTFS_ARCHIVE_FILTER_BANK}/{service_date.year}/stops.parquet")
```
rymarczy, line 69: Reading this file should also be filtered by a specific service date. While the `stops` file is relatively stable. Not filtering by service date may eventually result in unexpected results. 

I think there is also an open ticket on the parquet GTFS Archive files where when a new year rolls over, a new year parquet file is not automatically created, so there is the potential for this call to fail during the first days of a new year.

rymarczy: the `gtfs_from_parquet` function in `gtfs_utils.py` accounts for all of this.

src/lamp_py/utils/gtfs_utils.py
```diff
@@ -50,10 +49,17 @@ def gtfs_from_parquet(file: str, service_date: date) -> pl.DataFrame:
...
+def bus_route_ids_for_service_date(service_date: date) -> List[str]:
     """get a list of bus route ids for a given service date"""
     bus_routes = (
         gtfs_from_parquet("routes", service_date).filter((pl.col("route_type") == 3)).get_column("route_id").unique()
     )
 
     return bus_routes.to_list()
+
+
+def routes_for_service_date(service_date: date) -> pl.DataFrame:
```
rymarczy, line 61: Did you mean to use the function?

rymarczy (changes requested):
a few fixes.

src/lamp_py/utils/filter_bank.py
```diff
@@ -1,5 +1,77 @@
+from datetime import datetime
+from typing import List, Optional
 import pyarrow.compute as pc
 import pyarrow as pa
+import polars as pl
+
+GTFS_ARCHIVE_FILTER_BANK = "s3://mbta-performance/lamp/gtfs_archive"
```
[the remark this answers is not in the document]
huangh: addressed by swapping dynamic code with hardcoded - only dynamic stuff is in test now

src/lamp_py/tableau/conversions/convert_gtfs_rt_trip_updates.py
huangh: addressed by swapping dynamic code with hardcoded - only dynamic stuff is in test now

src/lamp_py/tableau/pipeline.py
huangh, line 1: define how each of these methods get called to reduce confusion - they're called from separate processes, but live in the same file.

rymarczy (approved):
:rocket:

PR #525, opened 2025-06-06
(fix) add timezones to default stop_crossings to resolve SchemaError

huangh:
[Asana Task](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1210268410561929?focus=true)

This issue was introduced in #519 

Resolve issue by adding timezones to default stop_crossings dataframe when no TM data is available.


src/lamp_py/bus_performance_manager/events_tm.py
```diff
@@ -25,14 +25,14 @@ def _empty_stop_crossing() -> pl.DataFrame:
...
-        "tm_actual_departure_dt": pl.Datetime,
+        "tm_scheduled_time_dt": pl.Datetime(time_zone="UTC"),
+        "tm_actual_arrival_dt": pl.Datetime(time_zone="UTC"),
+        "tm_actual_departure_dt": pl.Datetime(time_zone="UTC"),
         "tm_scheduled_time_sam": pl.Int64,
         "tm_actual_arrival_time_sam": pl.Int64,
         "tm_actual_departure_time_sam": pl.Int64,
     }
-    return pl.DataFrame(schema=schema)
+    return pl.DataFrame(schema=schema)  # type: ignore
```
rymarczy, line 35: Do we know why mypy didn't need to ignore this previously?

rymarczy (approved):
One question. 

:1st_place_medal:

src/lamp_py/bus_performance_manager/events_tm.py
huangh: pl.Datatype it recognizes as a type. it thinks pl.Datatype(timezone=timezone) is an object, which violates the { str: type } dict it expects. 

I tried to wrap it in `type(pl.Datatype(timezone)` (didn't work) and I tried `.base_type()` but that just gave me a Datetime (no timezone) again. 

 [arg-type] is the actual error it complains with

PR #526, opened 2025-06-06
(feat) regenerate bus data - add new fields to output rollup

huangh:
[Ticket](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1210404550176784?focus=true)

Originally planned on just using the .parquet, but without the timestamp conversion, and the rollup for the recent 7 days, it did not satisfy analyst requirements. Regenerating bus data to enable analysis.


rymarczy (dismissed):
🛫

PR #535, opened 2025-06-23
(fix/temp) limit BUS_ALL to 1 year temporarily

huangh:
Temporary/Bus all needs to be split up as it is becoming unwieldy to query in Tableau uncompressed. 

Lowering limit to 1 year temporarily while we are still in development of the Bus Performance Manager metrics. 

Will be implemented in this [ticket](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1210469663680236?focus=true)

Undo task: change YEAR for bus all to ALL (split up files)


rymarczy (approved):
🌮

PR #536, opened 2025-07-01
(feat) update Manual deploy to include run-name with details of manual run job

huangh:
Before it just says "Manual Deploy"

Now it tells you the environment deployed to, and which tasks were deployed under that deployment.


rymarczy (dismissed):
:+1:

PR #537, opened 2025-07-01
(feat) Ingest PatternGeoNodeXref, TimePoint, LoggedMessage tables from TM

huangh:
[Ticket](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1210610620622073?focus=true)

In support of Bus Performance Manager, need PatternGeoNodeXref and TimePoint tables as well

LoggedMessages also grabbed for future work with comparing LatLon coordinates


src/lamp_py/ingestion_tm/jobs/parition_table.py
```diff
@@ -240,3 +241,73 @@ def export_schema(self) -> pyarrow.schema:
...
+                logger.log_start()
+
+                query = sa.text(
+                    f"""
+                    SELECT
+                        {table_columns}
+                    FROM
+                        {self.tm_table}
+                    WHERE
+                        CALENDAR_ID = {date} AND MESSAGE_TYPE_ID = 16
```
huangh, line 286: duplicated run_export() to insert `MESSAGE_TYPE_ID` - without this, the amount of data extracted is an order of magnitude higher and not needed for now.

src/lamp_py/bus_performance_manager/events_tm.py
```diff
@@ -165,7 +209,11 @@ def generate_tm_events(tm_files: List[str]) -> pl.DataFrame:
                 pl.col("TRIP_SERIAL_NUMBER").cast(pl.String).alias("trip_id"),
                 pl.col("GEO_NODE_ABBR").cast(pl.String).alias("stop_id"),
                 pl.col("PATTERN_GEO_NODE_SEQ").cast(pl.Int64).alias("tm_stop_sequence"),
+                pl.col("tm_planned_sequence_end"),
```
huangh, line 212: grab the sequence_end to see later if a trip ends on the expected tm_stop_sequence

src/lamp_py/ingestion_tm/jobs/parition_table.py
```diff
@@ -126,8 +125,15 @@ def run_export(self, tm_db: MSSQLManager) -> None:
                 logger = ProcessLogger("tm_daily_log_export", tm_table=self.tm_table, date=date)
                 logger.log_start()
 
-                query = sa.text(
-                    f"""
+                if self.query_specialized_template is not None:
+                    query = sa.text(
+                        self.query_specialized_template.format(
```
huangh, line 130: format the template string if there's specific query logic needed...this is probably not generic enough, but currently ok for my usecase where i'm just adding an AND clause.

doing it this way to not duplicate the entire run_export() method in the child class

huangh:
I've tested and generated these tables in STAGING already to test - since I have no other way to access the TM dev environment. The only part that is new is the cleaned up run_export() method with the template.

src/lamp_py/bus_performance_manager/events_tm.py
rymarczy, line 212: Can you add some more detail to either the asana ticket or PR on why these tables/columns are being added to the `tm_stop_crossings` output. 

I understand pulling some new tables from the TM db and saving them as parquet files, but not why they would need to be added to any bus_performance_manager output.

src/lamp_py/ingestion_tm/jobs/parition_table.py
```diff
@@ -240,3 +246,39 @@ def export_schema(self) -> pyarrow.schema:
...
+            s3_location=tm_daily_logged_message,
+            tm_table="TMDailyLog.dbo.LOGGED_MESSAGE",
+            lamp_version="0.0.2",
+            specialized_template="""
+                    SELECT
+                        {table_columns}
+                    FROM
+                        {tm_table}
+                    WHERE
+                        CALENDAR_ID = {date}  AND MESSAGE_TYPE_ID = 16
```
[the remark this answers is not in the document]
rymarczy: 1-10mb per day seems pretty reasonable, are you sure you won't need other message types in the future?

PR #539, opened 2025-07-01
(feat) local regeneration of bus_recent dataset

huangh:
[Ticket](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1210578092051395?focus=true)

Implement local regeneration of BUS_RECENT dataset. 
Changes to analysis scripts to look at bus data - 

Additionally apply the regeneration of the most recent 7 days of bus_vehicle_events to allow BUS_RECENT datasets to be automatically up to date when columns are added.

### DIFF
--- a/src/lamp_py/bus_performance_manager/event_files.py
+++ b/src/lamp_py/bus_performance_manager/event_files.py
@@ -126,9 +126,13 @@ def transit_master_files_as_frame() -> pl.DataFrame:
     )
 
 
-def event_files_to_load() -> Dict[date, Dict[str, List[str]]]:
+def event_files_to_load(start_date: Optional[date], end_date: Optional[date]) -> Dict[date, Dict[str, List[str]]]:
     """
-    Generate a dictionary containing a record for every service date to be processed.
+    Generate a dictionary containing a record for a range of service_dates
+    to be re-processed.
+
+    If no dates are provided, processes the latest service date that was last modified
+
     * Collect all of the potential input filepaths, their last modified
         timestamp, and potential service dates.
     * Get the last modified timestamp for the output filepaths.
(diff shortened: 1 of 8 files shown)


rymarczy (changes requested):
This is another one with no details in the PR or asana ticket. Could you explain, in more detail, the purpose of these changes? 

If this is ONLY for local generation of these files, it doesn't seem worth modifying the application functions. If you need the files locally, can't you just download the ones already created by the application from S3?

src/lamp_py/bus_performance_manager/write_events.py
```diff
@@ -10,14 +12,16 @@
 from lamp_py.aws.s3 import upload_file
 
 
-def write_bus_metrics() -> None:
+def write_bus_metrics(
```
rymarczy, line 15: The core of this change seems to calling `bus_performance_metrics` with a user defined date range and then writing to a local path. 

This could be accomplished more easily with a new function that doesn't add such complexity to existing functions meant to be run by the ECS.

`event_files_to_load` and `write_bus_metrics` are designed to provide a specific scope of functionality to the ECS application. They are scaffolding around a call of `bus_performance_metrics` to allow the ECS to produce the expected output of parquet files. 

Could you re-write the changes as a new `local_write_bus_metrics` function to maintain a separation of concerns?

src/lamp_py/tableau/jobs/filtered_hyper.py
```diff
@@ -65,7 +65,7 @@ def create_tableau_parquet(self, num_days: Optional[int]) -> bool:
         # previous days uploads working
         # need to implement new input daily_upload_hour as well
         # this will currently update hourly. will monitor
-        end_date = datetime.now()
+        end_date = datetime.now().date()
```
rymarczy, line 68: You've got the `end_date` param of `file_list_from_s3_date_range` as accepting `datetime.datetime` type, not `datetime.date`

src/lamp_py/bus_performance_manager/write_events.py
```diff
@@ -8,16 +12,19 @@
...
 from lamp_py.aws.s3 import upload_file
+from lamp_py.tableau.jobs.bus_performance import BUS_RECENT_NDAYS
 
 
-def write_bus_metrics() -> None:
+def write_bus_metrics(
+    start_date: Optional[date] = None, end_date: Optional[date] = None, write_local_only: bool = False
+) -> None:
     """
     Write bus-performance parquet files to S3 for service dates neeing to be processed
```
rymarczy, line 22: suggestion: keep function signature comment up  to date with params/purpose

src/lamp_py/bus_performance_manager/event_files.py
```diff
@@ -126,9 +126,13 @@ def transit_master_files_as_frame() -> pl.DataFrame:
...
 
-def event_files_to_load() -> Dict[date, Dict[str, List[str]]]:
+def event_files_to_load(start_date: Optional[date], end_date: Optional[date]) -> Dict[date, Dict[str, List[str]]]:
     """
-    Generate a dictionary containing a record for every service date to be processed.
+    Generate a dictionary containing a record for a range of service_dates
+    to be re-processed.
+
+    If no dates are provided, processes the latest service date that was last modified
+
```
rymarczy, line 135: suggestion: keep function comment up to date with parameters, whether start/end dates and inclusive/exclusive

src/lamp_py/bus_performance_manager/write_events.py
```diff
@@ -64,3 +75,33 @@ def write_bus_metrics() -> None:
         day_logger.log_complete()
 
     logger.log_complete()
+
+
+def regenerate_bus_metrics_recent(num_days: int = BUS_RECENT_NDAYS) -> None:
```
rymarczy, line 80: I would advise against this approach, it seems like it will result in a mix of parquet files with different schemas for different date ranges. This will likely result in failures when try to read/merge all the files in a complete dataset. 

Updates to metrics schema should be an all or nothing event. Purposely updating only a portion of file schemas will result in future issues.

huangh: The point of this whole change is to allow changes to the schema to be immediately represented in the BUS_RECENT dataset without manual regeneration for analysts. 

I'm adding an additional check that ensures the schemas are compatible.

I will add an additional ticket to consider automatically updating the lamp_version with a minor version whenever i make these changes to allow us to track it, but I think for now the 2 cases possible are handled here:

1) schema are incompatible - now the check will not pass, and we will just update a single one. To regenerate the dataset, version bump is required and will be surfaced when tableau job fails
2) schema are compatible - check will pass, latest will update the recent 7 days

src/lamp_py/bus_performance_manager/event_files.py
huangh: incorporated

rymarczy (approved):
:zap:

PR #547, opened 2025-08-01
feat(spare/tableau): Ingest, flatten, and upload Spare resources to tableau

huangh:
[Prototype import spare to dev](https://app.asana.com/1/15492006741476/project/1209584755824899/task/1209698688344824?focus=true)
[Import all spare to dev/prod](https://app.asana.com/1/15492006741476/project/1209584755824899/task/1210535979320155?focus=true)

Related Devops/Terraform: 
https://github.com/mbta/terraform_modules/pull/287
https://github.com/mbta/terraform_modules/pull/289
https://github.com/mbta/terraform_modules/pull/295

https://github.com/mbta/devops/pull/3042

## Why do we need changes?

Spare/Paratransit tech is reusing the existing Tableau architecture developed by LAMP to upload data to tableau for analysts. all spare data is in the `/spare` partition of the same `dataplatform` buckets, and require specific permissions to access (as implemented in the devops tickets above. 

The core work here is to generically convert and flatten the files into a tableau hyperapi compliant shape. A simple way to do this is to simply convert everything to strings, ~~but here we added a utility to flatten/explode/convert all fields into a fully flat structure.~~ _Based on Comments, implemented just string flattening as a reasonable default._ This is reusable and generic for other Tableau uploading tasks. 

We also autogenerated the input/output and processing methods for each of these inputs, so adding a new source is simply adding it to a whilelist, and it will get processed and uploaded. 

Tested locally, tested main conversion method, tested in staging (since there is no environment to test in via dev...)

Limitations: 

We removed the ability to do "custom" conversions for now as they were not necessary, but an improvement in the future may be to add includes/excludes, or as requested leave structs as just a "string" if it causes analysis issues by being too exploded/deep. 

Another thing to watch is the runtime of this, as we've added ~46 new resources to upload, and this all runs in the Tableau cron job - which runs every hour.

### DIFF
--- a/tests/tableau/test_convert_types.py
+++ b/tests/tableau/test_convert_types.py
@@ -21,6 +21,6 @@ def test_convert_to_tableau_flat_schema() -> None:
     )
     print(df1)
 
-    expected = pl.read_json('tests/tableau/test_convert_types_expected.json')
+    expected = pl.read_json("tests/tableau/test_convert_types_expected.json")
 
     pt.assert_frame_equal(expected, convert_to_tableau_flat_schema(df1))
--- a/src/lamp_py/tableau/spare/default_converter.py
+++ b/src/lamp_py/tableau/spare/default_converter.py
@@ -37,20 +37,22 @@ def convert_to_tableau_flat_schema(self: pl.DataFrame, seperator="."):
         if len(struct_columns):
             self = _unnest_all(struct_columns=struct_columns)
         if len(list_columns):
-            # don't expand lists - leave it as a string repr for now. 
+            # don't expand lists - leave it as a string repr for now.
             # self = self.explode(list_columns).with_columns(pl.col(list_columns).cast(pl.String))
             # self['roles'].dtype.inner
             for col in list_columns:
-                # self = self.explode(columns=col)            
+                # self = self.explode(columns=col)
                 try:
                     if isinstance(self[col].dtype.inner, pl.Categorical) | isinstance(self[col].dtype.inner, pl.String):
-                        self = self.with_columns(pl.col(col).cast(pl.List(pl.String)).list.join(','))
+                        self = self.with_columns(pl.col(col).cast(pl.List(pl.String)).list.join(","))
                         continue
                     if isinstance(self[col].dtype.inner, pl.Struct):
-                        self  = self.with_columns(pl.col(col).list.eval(pl.element().struct.json_encode()).list.join(','))
+                        self = self.with_columns(
+                            pl.col(col).list.eval(pl.element().struct.json_encode()).list.join(",")
+                        )
                         continue
                     if isinstance(self[col].dtype.inner, pl.String):
-                        self = self.with_columns(pl.col(col).cast(pl.List(pl.String)).list.join(','))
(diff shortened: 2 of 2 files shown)


tests/tableau/test_convert_types.py
```diff
@@ -1,122 +1,27 @@
...
+            "data2": [
+                [{"name": "Darcy", "age": 30}, {"name": "Effie", "age": 25}],
+                [{"name": "Gerard", "age": 40}, {"name": "Henry", "age": 30}, {"name": "Iggy", "age": 25}],
+                [{"name": "Juliet", "age": 40}],
+            ],
+        }
+    )
+    print(df1)
+
+    json_str = '[{"data.name":"Alice","data.age":30,"data2.name":"Darcy","data2.age":30},{"data.name":"Alice","data.age":30,"data2.name":"Effie","data2.age":25},{"data.name":"Bob","data.age":25,"data2.name":"Darcy","data2.age":30},{"data.name":"Bob","data.age":25,"data2.name":"Effie","data2.age":25},{"data.name":"Charlie","data.age":40,"data2.name":"Gerard","data2.age":40},{"data.name":"Charlie","data.age":40,"data2.name":"Henry","data2.age":30},{"data.name":"Charlie","data.age":40,"data2.name":"Iggy","data2.age":25},{"data.name":null,"data.age":null,"data2.name":"Juliet","data2.age":40}]'
```
skyqrose, line 24: This is doing a cartesian product between each list column. That is one way to flatten data, but I don't think it's the right approach for most of this data. For example, a rider has a list of mobilityaids and a list of disabilities and a list of conditions. You don't really want to explode each rider into one row per combination of mobilityaid,disability,condition.

Here are a couple other options:
1. Skip uploading some list columns. E.g. I grabbed a column `rider.notificationPreference` that happened to be in the data but probably isn't actually important for analytics. We'd have to chat about each column cuz it requires domain knowledge, and there will still be many list columns to keep.
2. Flatten each column individually into a string-encoded list. E.g. `data: ["[{name:"Alice",age:30},{name:"Bob",age:25}]", "...", "[]"]`. This ruins the ability to do analytics on the data without lots of effort, but indicates to analysists that the data exists and they should ask for solution 3 for columns they care about.
3. Expand some lists into multiple columns. E.g. for mobilitiyAids, there could be a boolean `hasWalker` column and a `hasWheelchair` column, etc. This is high effort and requires knowing a finite list of all possible list members, so I don't want to do it except for columns we know are important. This should probably happen when producing the parquet file, so it'd be just a simple boolean column for the tableau publisher. (Also works for some non-boolean data, like instead of a `riderCounts` list column, having `numSeniors`, `numAdults`, `numChildren` int columns.)
4. Make a new table with a one-to-many relationship. This is high effort, so I don't think it's a good first option, but it would be a good structure if we decided it was worth doing it Right.
5. For lists of finite max length, have one column per potential element. `data_0.name, data_0.age, data_1.name, data_1.age, data_2.name, data_2.age`. But it's hard to analyze and requires knowing which lists are finite length, so I don't think it's a good option.

In my opinion, I think 2 is a fine option for just getting something down, though it's obviously not a good permanent solution for analysis.

PR #559, opened 2025-08-14
Change threshold for subway service_date from 3am to 4am

runkelcorey:
Asana Task: https://app.asana.com/1/15492006741476/project/1210964140773280/task/1211027583249655?focus=true

# Why do we need changes?

Currently, we assign a small amount of subway TripUpdates and VehiclePositions by comparing the `timestamp`/`vehicle_timestamp` with 3am. Extended late-night service on subways in the Fall 2025 rating means that, if we kept that logic intact, then it's possible that records with timestamps between 3am and 4am could have service_dates greater than later records on different trips and earlier records on the same trip.

## Edits
1. Threshold for determining subway service dates from 3am to 4am in `lamp_py.performance_manager.gfts_utils.service_date_from_timestamp`
2. Associated unit test

1. Unit tested with expected results.

1. Should we instead filter these records out, as BusPM does?


rymarczy (approved):
I reviewed all of the RT_VEHICLE_POSITION springboard source parquet files for the year. When accounting for the pre-filters performance_manager executes on the parquet files (https://github.com/mbta/lamp/blob/8859fc58e171dbed45a4e7761bb0aa69f84302a0/src/lamp_py/performance_manager/l0_rt_vehicle_positions.py#L45) `service_date_from_timestamp` has been applied to 0 records in the past 226 days. (`start_date` has not been NULL for any VehiclePosition records used in the performance_manager pipeline)

So I do think of this change as essentially a no-op. Ideally, this PR would show how many RT_VEHICLE_POSITION records are currently being produced between 3am and 4am, to get an idea of how many records this could potentially impact. Are we currently seeing any trips for the current start_date beginning after 3am, for which this change would incorrectly assign a start_date? 

The asana ticket referenced for this change also states that subway trips will go until ~2:45am, which is already inside the currently defined 3am cutoff. 

Approving because of the no-op status, but still slightly murky on the justification for pushing this value back to 4am. 

--updated for clarity

PR #560, opened 2025-08-14
(feat) Refactor bus schedule join order

huangh:
Asana Ticket:  [ticket](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1210966395794050?focus=true)

## Why do we need changes?

OA needs non-revenue timepoints in the bus dataset to properly evaluate runtimes between stops e.g. duration through ted williams tunnel. without these points in the set, the duration would be measured between Silverline way and Logan...which is not precise. 

Existing schedule method started with gtfs events joined with tm events, followed by joining gtfs schedule on, but this meant non-revenue TM timepoints were never represented in the final dataset. Prior efforts to address this involved concatenating the additional rows onto the dataset, but this had issues where it was difficult to fill in the rest of the columns in the concat rows. 

Instead, reworking the base schedule to do Schedules + Schedules, then join events on to the schedule dataframe. 

### Changes
- GTFS schedule joined with TM schedule
- fix fallout and reorganize other methods given schedule is now joined first

- Validation script as committed. Checking expected shape and contents of gtfs schedule, tm schedule, and joined schedule

---
  - https://app.asana.com/0/0/1210941486415515


src/lamp_py/bus_performance_manager/events_tm_schedule.py
huangh, line 1: need to join unfiltered tm_pattern_geo_node_xref to get all of the timepoint ids - so we have a way to sort the entire df

src/lamp_py/bus_performance_manager/combined_bus_schedule.py
```diff
@@ -0,0 +1,153 @@
...
+                "tm_planned_sequence_end",
+                "timepoint_id",
+                "timepoint_abbr",
+                "timepoint_name",
+                "pattern_id",
+            ]
+        )
+    ).with_row_index()
+
+    schedule = schedule.with_columns(
```
huangh, line 126: this section below is non-blocking error checking/logging

src/lamp_py/bus_performance_manager/events_tm.py
```diff
@@ -70,98 +70,6 @@ def generate_tm_events(tm_files: List[str]) -> pl.DataFrame:
     """
     logger = ProcessLogger("generate_tm_events", tm_files=tm_files)
     logger.log_start()
-    # the geo node id is the transit master key and the geo node abbr is the
```
huangh, line 73: TM schedule tables are joined before and passed into this method now

src/lamp_py/tableau/conversions/convert_bus_performance_data.py
```diff
@@ -1,5 +1,38 @@
 import polars as pl
 from pyarrow import Table
+import pyarrow
+
+
+def convert_bus_recent_to_tableau_compatible_schema(
```
runkelcorey, line 6: thanks for this! It *feels* like there should be an even-more-elegant way to handle these conversions but who has the time to consider those?

pyproject.toml
```diff
@@ -76,7 +76,7 @@ pretty = true
 python_version = "3.12"
 warn_unreachable = true
 warn_unused_ignores = true
-exclude = ["investigation/", "runners", "analysis/"]
+exclude = ["investigation/", "runners", "analysis/", "src/lamp_py/tableau/spare/"]
```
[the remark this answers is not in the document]
huangh: Confirm this is unrelated - but this is already in main, this comment here acknowledges this, but not worth taking out again as there's no code impact.

src/lamp_py/tableau/conversions/convert_bus_performance_data.py
huangh: i've implemented the elegant in the spare PR, but didn't migrate this all over yet.

PR #566, opened 2025-09-08
Add dataframely typing

runkelcorey:
Asana Task: https://app.asana.com/1/15492006741476/project/1189492770004753/task/1211291821750054

# Why do we need changes?

1. Mentally keeping track of expected input and output dataframes grows more difficult as we compose more functions and is difficult to dive back into for debugging
2. It's important to test data-transformation functions, but unit-testing complex transformations requires lots of code to setup the test

# What changes does this PR propose?

## Adds
1. `dataframely` to `dev` group and `mypy` plugins. [This library](https://dataframely.readthedocs.io/en/latest/index.html) enables Pydantic-style typing of Polars dataframes. In the future, we can substitute type testing for the unit tests that check the output schema of dataframes
2. `dataframely` classes for selected bus PM dataframes to demonstrate its usage

I looked for Python frameworks with 4 features:
- suggests a way to validate data transformations in a CI/CD pipeline
- provides a built-in way to check cross-dataframe properties, such as referential integrity or [state changes between upstream and downstream dataframes](https://app.asana.com/1/15492006741476/project/1210964140773280/task/1210058693545805?focus=true)
- declarative way to define (and run) custom data-validation logic
- incremental way to apply to existing codebase

I considered several libraries and found `dataframely` the most promising, especially for it's support of cross-dataframe testing:
| framework          | CI/CD support | cross-dataframe | custom logic | incremental |
|--------------------|---------------|-----------------|--------------|-------------|
| Great Expectations | ✅             | ✅               | ✅            | ❌           |
| cuallee            | ❌             | ❌               | ⚠️           | ✅           |
| pointblank         | ❌             | ⚠️              | ✅            | ✅           |
| pandera            | ⚠️            | ❌               | ✅            | ✅           |
| patito             | ✅             | ❌               | ✅            | ✅           |
| dataframely        | ✅             | ✅               | ✅            | ✅           |

## Edits
1. Bumped Python version for `black` to 3.12
3. Refactored `bus_performance_metrics` to separate the joining from the additional fields

# How were these changes validated?

1. `mypy` succeeded
2. `pytest` succeeded
3. `run_bus_recent_events` succeeded

# What questions should reviewers consider?

1. How worried are you about stability of `dataframely`? `pandera` is the trusty steed of this realm but doesn't support data generation for polars dataframes or declarative testing of dataframe relationships
2. Should I create a unit test for my new function and, if so, what would I test separate from what dataframely tests?

PR #567, opened 2025-09-09
Reveal (and address) mypy errors

runkelcorey:
Asana Task: https://app.asana.com/1/15492006741476/project/1189492770004753/task/1211291821750054

# Why do we need changes?

`mypy` silently ignores typing issues throughout the codebase. Once those errors are revealed, they need to be fixed. Also, #566 included a commit tree that I wanted to split into a different PR.

## Adds
1. `src/lamp_py/__init__.py`, allowing `mypy` to recognize `src/lamp_py` as a python module
2. `src/lamp_py/py.typing`, allowing `mypy` to recognize that it should type check `lamp_py` when it checks `tests/`

## Edits
1. `[tools.mypy]` in `pyproject.toml` to ignore `union-attr` errors because they are simply too numerous to deal with right now; these occur when any of the possible types in a union of types lacks the attribute or method referenced. For example, imagine an argument `x` typed `Optional[Union[float, str]]`: 
    - a reference to `x.str` will throw a `union-attr` error unless x is asserted to be one of `float` or `str`
    - a reference to `x.endswith` will throw a `union-attr` error unless `x` is asserted to be `str`

1. GitHub Actions + my local see that `mypy` has 0 errors

1. Do any of these annotation changes raise flags?

### DIFF
--- a/src/lamp_py/tableau/jobs/filtered_hyper.py
+++ b/src/lamp_py/tableau/jobs/filtered_hyper.py
@@ -10,7 +10,7 @@ import polars as pl
 
 from lamp_py.runtime_utils.process_logger import ProcessLogger
 from lamp_py.tableau.hyper import HyperJob
-
+from lamp_py.postgres.postgres_utils import DatabaseManager
 from lamp_py.runtime_utils.remote_files import S3Location
 
 from lamp_py.aws.s3 import file_list_from_s3, file_list_from_s3_date_range
@@ -51,10 +51,10 @@ class FilteredHyperJob(HyperJob):
     def output_processed_schema(self) -> pyarrow.schema:
         return self.processed_schema
 
-    def create_parquet(self, _: None) -> None:
+    def create_parquet(self, _: DatabaseManager | None) -> None:
         self.update_parquet(None)
 
-    def update_parquet(self, _: None) -> bool:
+    def update_parquet(self, _: DatabaseManager | None) -> bool:
         return self.create_tableau_parquet(num_days=self.rollup_num_days)
 
     # pylint: disable=R0914, R0912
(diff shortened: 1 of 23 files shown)


src/lamp_py/tableau/jobs/filtered_hyper.py
```diff
@@ -27,7 +27,7 @@ def __init__(
         remote_output_location: S3Location,
         processed_schema: pyarrow.schema,
         tableau_project_name: str,
-        rollup_num_days: int = 7,  # default this to a week of data
+        rollup_num_days: Optional[int] = 7,  # default this to a week of data
```
huangh, line 30: we are generally moving towards the `int | None` syntax to eliminate importing of the Typing library unnecessarily. Lets not introduce them back in. 

```suggestion
        rollup_num_days: int | None = 7,  # default this to a week of data
```

src/lamp_py/tableau/jobs/gtfs_rail.py
```diff
@@ -47,18 +47,19 @@ def __init__(
     def output_processed_schema(self) -> pyarrow.schema:
         """Define GTFS Table Schema"""
 
-    def create_parquet(self, db_manager: DatabaseManager) -> None:
+    def create_parquet(self, db_manager: DatabaseManager | None) -> None:
         if os.path.exists(self.local_parquet_path):
             os.remove(self.local_parquet_path)
 
+        assert isinstance(db_manager, DatabaseManager)
```
huangh, line 54:
```suggestion
        if not isinstance(db_manager, DatabaseManager):
            raise TypeError("db_manager must be of type DatabaseManager for Rail Performance Manager")
```

huangh: if you like this, apply to all the other `Type | None` ... `assert(type, Type)` calls introduced in this PR. I think using the assert to check in running code isn't best practice. Though i'm sure i've done the same thing somewhere in this repo...

huangh (changes requested):
I'm deplying this to dev to check if there are any issues for the changes made to RPM, and we should watch it in staging as well since we have a lot of surface area for this PR.

PR #573, opened 2025-09-15
fix(tableau): address missed callsite for tableau methods refactor

huangh:
addressing issues introduced in https://github.com/mbta/lamp/pull/567

team member permissions mismatch in s3 resulting in untested call-site for refactored methods. we'll all get permissions and that should address this

---
  - https://app.asana.com/0/0/1211366080297135


runkelcorey (dismissed):
Looks good! What does the change to walletTransactions mean? Is it just a rename?

huangh:
> Looks good! What does the change to walletTransactions mean? Is it just a rename?

reverted it as unnecessary change when debugging - as its a trivial change, I will merge it on pipeline success with your prior approval, unless you re-approve first

PR #575, opened 2025-09-16
Soft fail dataframely errors and populate service_date

runkelcorey:
Asana Task: https://app.asana.com/1/15492006741476/project/1189492770004753/task/1211373568540575?focus=true

# Why do we need changes?

Currently, our code hard fails when it runs due to dataframely exceptions. However, none of these failures should result in a failure (at least so long as we are still in active development). One of the errors is that `service_date` is null for all records that don't have a GTFS schedule match. While I could have put that change in a new PR, I didn't because it affected so many records.

## Edits
1. Fills null `service_date`s by parsing `plan_start_dt`
2. Multiple dataframe schemas to fix primary keys
3. `dataframely` validation to pass the valid records and log the invalid records

## Adds
1. Test for the `timestamp_to_service_date` conversion

1. Testing suite
2. `run_bus_recent_events` runner finished (with logged dataframely exceptions)

1. Should I make a function for the dataframely logging? It could accept a dataframe to be validated and return the valid records while logging any exceptions as a side effect

---
  - https://app.asana.com/0/0/1211291821750031

### DIFF
--- a/analysis/vehicle_position_viewer.py
+++ b/analysis/vehicle_position_viewer.py
@@ -1,0 +1,84 @@
+import marimo
+
+__generated_with = "0.14.16"
+app = marimo.App(width="medium")
+
+
+@app.cell
+def _():
+    import marimo as mo
+    import polars as pl
+    import pyarrow
+    import pyarrow.dataset as ds
+    from pyarrow.fs import S3FileSystem
+    import plotly.express as px
+
+    return S3FileSystem, ds, mo, pl, px
+
+
+@app.cell
+def _():
+    vehicle_id = "G-10204"
+    trip_id = "71193618"
+    return trip_id, vehicle_id
+
+
+@app.cell
+def _(px):
+
+    def plot_lla(vp):
+        fig3 = px.scatter_map(
+            vp,
+            lat="vehicle.position.latitude",
+            lon="vehicle.position.longitude",
+            zoom=8,
+            height=300,
+            hover_data="vehicle.timestamp",
+        )
+        fig3.update_layout(mapbox_style="open-street-map")
+        fig3.update_layout(margin={"r": 0, "t": 0, "l": 0, "b": 0})
+        config2 = {"scrollZoom": True}
+        return fig3.show(config2)
+
+    return (plot_lla,)
+
+
+@app.cell
+def _(ds_helper, pl, vehicle_id):
+    vp_raw = ds_helper(
+        [
+            "s3://mbta-ctd-dataplatform-springboard/lamp/RT_VEHICLE_POSITIONS/year=2025/month=9/day=17/2025-09-17T00:00:00.parquet"
+        ],
+        (pl.col("id") == vehicle_id),
+    )
+    return (vp_raw,)
+
+
+@app.cell
+def _(mo, vp_raw):
+    mo.ui.dataframe(vp_raw)
+    return
+
+
+@app.cell
+def _(pl, plot_lla, trip_id, vp_raw):
+    plot_lla(vp_raw.filter(pl.col("vehicle.trip.trip_id") == trip_id))
+    return
+
+
+@app.cell
+def _(S3FileSystem, ds, pl):
+    def ds_helper(s3_uris, filter):
+        ds_paths = [s.replace("s3://", "") for s in s3_uris]
+        dset = ds.dataset(ds_paths, format="parquet", filesystem=S3FileSystem())
+        if filter is not None:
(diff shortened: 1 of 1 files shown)


src/lamp_py/bus_performance_manager/events_metrics.py
```diff
@@ -30,6 +35,21 @@ class BusEvents(CombinedSchedule, TransitMasterEvents, GTFSEvents):  # pylint: d
     direction_destination_headway_seconds = dy.Int64(nullable=True)
 
 
+def timestamp_to_service_date(timestamp_column: pl.Expr, service_date_start_hour: int = 3) -> pl.Expr:
```
huangh, line 38: I think this method will lead to bugs in the future - it needs more timezone checks. It's only incidentally working correctly, because `plan_start_dt` is calculated as a seconds of offset from `service_date`, which is an America/New_York Date. so the math in here is only right because the inputs are in the same timezone.

tests/bus_performance_manager/test_gtfs_rt_ingestion.py
```diff
@@ -106,9 +106,6 @@ def test_gtfs_rt_to_bus_events(exists_patch: mock.MagicMock) -> None:
 
     assert not bus_vehicle_events.is_empty()
 
-    for col, data_type in bus_vehicle_events.schema.items():
```
huangh, line 109: these are removed bc dataframely handles these checks

src/lamp_py/bus_performance_manager/events_metrics.py
huangh: I think we also need to make sure we're explicit on definitions - `plan_start_dt` is not a timestamp, it's a datetime. so having the method say convert timestamp is a little confusing.

runkelcorey: It turns out that Polars only supports checking timezones from the DataFrame and not from an expression. I'm going to abandon this function and just transform the dataframe, where we know the input type

src/lamp_py/bus_performance_manager/events_metrics.py
```diff
@@ -111,8 +116,16 @@ def enrich_bus_performance_metrics(bus_df: pl.DataFrame) -> dy.DataFrame[BusEven
 
     :return BusEvents:
     """
+    process_logger = ProcessLogger("enrich_bus_performance_metrics")
+
     bus_df = (
         bus_df.with_columns(
+            pl.coalesce(
+                pl.col("service_date"),
+                pl.when(pl.col("plan_start_dt").dt.hour() < 3)
```
huangh, line 125: Let's not add more magic numbers to determine service_date. 
This change introducing a bug wrt `service_date_from_timestamp` where we had previously set service_date hour to end at 4am EST. However, i think that one operates on the "YYYYMMDD" string, whereas you've coerced it into a date.

huangh: perhaps just factor out the 3 here and the 4 there and put it in a (TM) good place so we have one spot to look for for global configs.

huangh: i responded in line accidentally, so it's below https://github.com/mbta/lamp/pull/575#discussion_r2356546378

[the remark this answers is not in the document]
runkelcorey: iirc we deliberately did not want to change the bus service date start hour from 3 to 4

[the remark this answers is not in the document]
huangh: iIrc (lol) I thought that was only because `service_date` as provided handled it for us. 

Either way, you're right that 3 should be the split, but I think we might consider dynamically setting it in a future ticket based on the schedule.

```

stop_events.select('plan_start_dt').sort(by="plan_start_dt", descending=False)
shape: (202_484, 1)
┌────────────────────────────────┐
│ plan_start_dt                  │
│ ---                            │
│ datetime[μs, America/New_York] │
╞════════════════════════════════╡
│ 2025-09-14 03:20:00 EDT        │
│ 2025-09-14 03:20:00 EDT        │
│ 2025-09-14 03:20:00 EDT        │
│ 2025-09-14 03:20:00 EDT        │
│ 2025-09-14 03:20:00 EDT        │
│ …                              │
│ 2025-09-15 02:19:00 EDT        │
│ 2025-09-15 02:19:00 EDT        │
│ 2025-09-15 02:19:00 EDT        │
│ 2025-09-15 02:19:00 EDT        │
│ 2025-09-15 02:19:00 EDT        │
└────────────────────────────────┘
```

PR #576, opened 2025-09-16
New PR template

runkelcorey:
Asana Task: https://app.asana.com/1/15492006741476/project/1189492770004753/task/1211373568540579?focus=true

# Why do we need changes?

Because I think this template encourages better review and I am done looking for a previous PR to copy.

## Adds
1. A section for motivation
2. A section for changes
    - subsections for edits, removals, and additions
3. A section for validation steps
4. A section directing reviewers questions

They weren't! You will :)

1. Is this too prescriptive?

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,32 +1,32 @@
 Asana Task: <asana_ticket_url>
 
-# Why do we need changes?
+## Why do we need changes?
 
 A sentence describing what motivated this PR. Maybe it's 2 sentences long if it's complex.
 
 
-# What changes does this PR propose?
+## What changes does this PR propose?
 
-## Adds
+### Adds
 1. A new thing (remove this section if you didn't add anything)
 
-## Edits
+### Edits
 1. List
 2. The
 3. Changes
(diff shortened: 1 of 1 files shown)


runkelcorey:
> I'm down for updating the template in general, but I have opinions. Some of this templates questions are rehashing what should be in the Asana ticket. I also don't agree with the add/edit/remove, because that's what the Files Changed view is for. I would replace this with a general "Changes" section that's just logical groupings of what changes were done.
> 
> So file/change level - already in Files Changed logical change - should be per git commit, and there's already context here overall grouping - what I would consider reasonable to stick in the PR text. How/why? - Asana
> 
> I'd be more willing for a section for "Technical Details" for things tried, and why choices were made, from a sw perspective. And then we can leave most of the "why" in Asana.

Good points! I'll remove the Why section and change the Changes section. I think the changes section and questions for the reviewers is the place to raise what was tried and any reservations about the choices made. That keeps the issues at hand closely linked to the changes that they impact. Thoughts?

huangh (approved):
we can iterate 👍

PR #585, opened 2025-09-24
feat(bus): correct join logic for overload and multi-route trip ids

huangh:
[Asana Task](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1210774178459682?focus=true)
## Background 

Overload trips - denoted by suffix "-OL" - are trips that run under the same trip id/route, that are "overloaded" i.e. extra buses running to improve service. The suffix appears in the GTFS data, but not in TM. In TM, these trips are all categorized under the same trip_id. The differentiator is vehicle_label. 

Split route/variants - denoted by suffix "_1" or "_2" - are trips that are special variants/combinations of multiple routes. In TM, these all appear under the same trip id. In GTFS, the two halves of the trip i.e. 57 bus -> 66 bus are given the "_1/_2" suffix. 

Example: 
[trip_part_1](https://www.mbta.com/schedules/57/line?schedule_direction%5Bdirection_id%5D=1&schedule_direction%5Bvariant%5D=57-E-1)
[trip_part_2](https://www.mbta.com/schedules/57/line?schedule_direction%5Bdirection_id%5D=1&schedule_direction%5Bvariant%5D=66-C-1)

- Strip off "-OL\d" and "_1" or "_2" from GTFS, make this a new join column. Join with TM based on this column. 
- ASOF join on sequence over trip_id and vehicle label will handle the the join properly, so we have all of these trips represented. 
- improve bus_runner script to save off intermediates based on flags for easier validation
- 

Unit tests, runner script, notebook


src/lamp_py/bus_performance_manager/events_gtfs_schedule.py
```diff
@@ -339,7 +339,17 @@ def bus_gtfs_schedule_events_for_date(service_date: date) -> pl.DataFrame:
         "departure_seconds",
         "parent_station",
     ]
-    stop_events = stop_events.drop(drop_columns).rename({"trip_id": "plan_trip_id"})
+    stop_events = (
+        stop_events.drop(drop_columns)
+        .rename({"trip_id": "plan_trip_id"})
+        .with_columns(
+            pl.col("plan_trip_id")
+            .str.replace(r"_\d", "")
```
runkelcorey, line 347: why not call `remove_overload_and_special_route_suffix` so that logic is centrally editable?

tests/bus_performance_manager/test_gtfs.py
```diff
@@ -99,8 +99,7 @@ def test_gtfs_events_for_date(exists_patch: mock.MagicMock) -> None:
     # drop new lines in expected_bus_events
     # read it in
     # run this
-    # bus_events.filter(pl.col.plan_trip_id.is_in(expected_bus_events['plan_trip_id'].unique())).sort(by=["plan_trip_id", "stop_sequence"]).write_csv('bus_test_gtfs.csv')
-
+    # bus_events.filter(pl.col.plan_trip_id.is_in(expected_bus_events['plan_trip_id'].unique())).sort(by=["plan_trip_id", "stop_sequence"]).write_csv('tests/bus_performance_manager/bus_test_gtfs.csv')
     # CSV Bus events
     expected_bus_events = pl.read_csv(
         os.path.join(current_dir, "bus_test_gtfs.csv"),
```
runkelcorey, line 105: can you write this idempotently? like using a try/except so that we don't have this commented out code hanging out in the test?

src/lamp_py/bus_performance_manager/events_gtfs_rt.py
```diff
@@ -26,6 +26,22 @@ class GTFSEvents(BusTrips):
     gtfs_arrival_dt = dy.Datetime(nullable=True, time_zone="UTC")
     latitude = dy.Float64(nullable=True)
     longitude = dy.Float64(nullable=True)
+    trip_id_suffix_removed_lamp_key = dy.String(nullable=True)
```
runkelcorey, line 29: what if we overwrote `trip_id` instead of adding a new column? I think your change is simple enough that users can backtrack and find the original trip ID if they wanted to

runkelcorey (changes requested):
Shouldn't this PR also change `join_rt_to_schedule` so that the TransitMaster trips get appropriately joined to the overloaded trips?

src/lamp_py/bus_performance_manager/events_gtfs_rt.py
```diff
@@ -26,6 +26,22 @@ class GTFSEvents(BusTrips):
...
+
+    @dy.rule()
+    def _no_split_trips1() -> pl.Expr:
+        return ~pl.col("trip_id_suffix_removed_lamp_key").str.ends_with("_1")
+
+    @dy.rule()
+    def _no_split_trips2() -> pl.Expr:
+        return ~pl.col("trip_id_suffix_removed_lamp_key").str.ends_with("_2")
+
+    # pylint: enable=no-method-argument
```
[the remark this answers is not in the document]
huangh: I kept it contains is because ends_with didn't have a regex option, and i already strip -OL\d in the code so this is just a sanity check. the -OL also has a number at the end of it too so the suggestion is not correct.

I don't think there are any 3 section special variants that would require _3.

huangh:
> Shouldn't this PR also change `join_rt_to_schedule` so that the TransitMaster trips get appropriately joined to the overloaded trips?

...yes. good catch

src/lamp_py/bus_performance_manager/events_gtfs_rt.py
huangh: shouldn't your validation do that for us?

src/lamp_py/bus_performance_manager/combined_bus_schedule.py
```diff
@@ -45,12 +48,20 @@ def join_tm_schedule_to_gtfs_schedule(gtfs: pl.DataFrame, tm: TransitMasterTable
     process_logger = ProcessLogger("join_tm_schedule_to_gtfs_schedule")
     process_logger.log_start()
 
-    tm_schedule = tm.tm_schedule.collect().filter(
-        pl.col("TRIP_SERIAL_NUMBER").cast(pl.String).is_in(gtfs["plan_trip_id"].unique().implode())
-    )
+    tm_schedule = tm.tm_schedule.filter(
+        pl.col("TRIP_SERIAL_NUMBER")
+        .cast(pl.String)
+        .is_in(gtfs.with_columns(remove_rare_variant_route_suffix("plan_trip_id"))["plan_trip_id"].unique().implode())
```
huangh, line 54: this is key - without removing the suffixes, the tm_schedule trip_ids get filtered out because they are 12345 and not 12345_1, 12345_2 (i.e. 12345_1/2 are not in tm_schedule, so 12345 never gets filtered. 

This filter is here to slim down tm_schedule considerably - 40m rows -> ~100k is because the tm_schedule has ALL trip_ids from forever, and not just the ones that are active.

src/lamp_py/bus_performance_manager/events_gtfs_rt.py
runkelcorey, line 1: clean 🧼

runkelcorey (approved):
nicely done

PR #591, opened 2025-10-01
fix(tableau) update dataframely schema and pyarrow bus_schema in tableau to match 

huangh:
Asana Task: [ticket](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1211342873330825?focus=true)

Recent changes to schemas in bus_performance_manager metrics generation have not flowed down to the tableau job. Want to keep the tableau job schema in place, as it is essentially an agreed upon contract with our downstream consumers for what columns they want and expect, so I think it's a good idea to leave that other schema in place, and just add a test to make sure we're generating it correctly. 

Changes in:
https://github.com/mbta/lamp/pull/566
https://github.com/mbta/lamp/pull/577

- Direction string column missing from bus schema - add this back in. It is an expected field for the bus performance metrics  
- fix fallout

- test added to compare schema and ensure we catch schema drift in the future


tests/tableau/test_bus_schema.py
```diff
@@ -0,0 +1,15 @@
+from lamp_py.bus_performance_manager.events_metrics import BusPerformanceMetrics
+from lamp_py.tableau.conversions.convert_bus_performance_data import apply_bus_analysis_conversions
+from lamp_py.tableau.jobs.bus_performance import bus_schema
+import polars as pl
+import pyarrow
+
+
+def test_bus_parquet_convertable_to_bus_tableau_columns() -> None:
```
runkelcorey, line 8: nit
```suggestion
def test_bus_parquet_convertible_to_bus_tableau_columns() -> None:
```

runkelcorey (changes requested):
One typo.

Another way to go about this would be to remove `bus_schema` and then change [these lines](https://github.com/mbta/lamp/blob/653ffa6ccd64107a600dcd32edd49d128f7dd6cc/src/lamp_py/tableau/jobs/bus_performance.py#L126-L128) to
```python
def output_processed_schema(self) -> pyarrow.schema:
        return BusEvents.pyarrow_schema()
```
But I sense you don't want to do that b/c you want to catch our schema drifting with a failed test. If this test fails, is the idea that we alert OpsAnalytics/anyone else who uses this data?

huangh:
> One typo.
> 
> Another way to go about this would be to remove `bus_schema` and then change [these lines](https://github.com/mbta/lamp/blob/653ffa6ccd64107a600dcd32edd49d128f7dd6cc/src/lamp_py/tableau/jobs/bus_performance.py#L126-L128) to
> 
> ```python
> def output_processed_schema(self) -> pyarrow.schema:
>         return BusEvents.pyarrow_schema()
> ```
> 
> But I sense you don't want to do that b/c you want to catch our schema drifting with a failed test. If this test fails, is the idea that we alert OpsAnalytics/anyone else who uses this data? Or do you imagine a different action that we take?

If I continue yakshaving this, I would do something like that. But the tableau schema is subtly different than the BusPerformanceMetrics schema, so it wouldn't be the exact same. There are a set of transforms i need to do (strip timezones, some other things... that make them not match exactly. 

I'm doing minimal changes to just move forward, but I need to:

- make the tableau schema a dataframely schema
- add timezone stripping to the universal converter
- move bus_tableau creation over to the universal converter
- delete all this conversion code
- ??
- profit

runkelcorey:
Okay, I'll approve if you fix that typo

runkelcorey (approved):

PR #592, opened 2025-10-01
bug(bus) Compute lat/lon for each record

runkelcorey:
I realized while investigating https://app.asana.com/1/15492006741476/project/1210964140773280/task/1211479998902404?focus=true that lat/lons are prone to be duplicates for duplicate `trip_id`, `stop_sequence`, `service_date`s. This is a problem for debugging but nothing downstream because duplicate records are filtered out by dataframely.

1. Computes lat/lon fields for each group of
```
"service_date",
"trip_id",
"stop_sequence",
"route_id",
"stop_id",
"direction_id",
"start_time",
"vehicle_id",
"vehicle_label",
```
instead of on each group of
```
"service_date",
"trip_id",
"stop_sequence",
```

2. Fixes typo in dataframely check that each first stop sequence has a `gtfs_departure_dt`
3. Adds `stop_id` as a primary key in `GTFSEvents`

1. Debugged run_bus_recent_events runner, saw differences in the lat/lon fields in duplicate records

1. None

### DIFF
--- a/src/lamp_py/bus_performance_manager/events_gtfs_rt.py
+++ b/src/lamp_py/bus_performance_manager/events_gtfs_rt.py
@@ -268,7 +268,6 @@ def positions_to_events(vehicle_positions: pl.DataFrame) -> dy.DataFrame[GTFSEve
 
     valid = logger.log_dataframely_filter_results(*GTFSEvents.filter(vehicle_events))
 
-
     logger.log_complete()
 
     return valid


src/lamp_py/bus_performance_manager/events_gtfs_rt.py
```diff
@@ -254,63 +255,17 @@ def positions_to_events(vehicle_positions: pl.DataFrame) -> dy.DataFrame[GTFSEve
             .then(pl.col("vehicle_timestamp"))
             .max()
             .alias("gtfs_departure_dt"),
+            pl.col("latitude").get(pl.col("vehicle_timestamp").arg_min()).alias("latitude"),
```
huangh, line 258: I know this is exactly how it was before, but is this even correct? feels like we're going to get the lat from the first timestamp, and the lon from the last timestamp, and then have a frankenstein lat,lon coordinate.

huangh (dismissed):

PR #597, opened 2025-10-03
fix(tableau/bus) cleanup tableau uploading for bus

huangh:


huangh:
these changes were made in https://github.com/mbta/lamp/pull/596 and https://github.com/mbta/lamp/pull/591

PR #722, opened 2026-03-27
feat(ingestion): add new metadata field to TripsUpdated

runkelcorey:
Asana Task: https://app.asana.com/1/15492006741476/project/1189492770004753/task/1213810187567346?focus=true

Adds `uiVersion` to `metadata` for `TripUpdatesRecord`.

1. Updates `test_append_records` to test an extra, unexpected field in the new Glides records.
2. Tested `test_append_records` to test a missing, nullable field in the remote Glides records

1. Do you see an elegant way to a test case to handle a missing, nullable field in the remote Glides records? Or should I do a new test b/c this one is growing big

### DIFF
--- a/src/lamp_py/ingestion/glides.py
+++ b/src/lamp_py/ingestion/glides.py
@@ -20,8 +20,10 @@ from lamp_py.runtime_utils.remote_files import (
     S3_SPRINGBOARD,
 )
 
-RFC3339_DATE_REGEX = r"^20(?:([1-3][0-9]-[0-1][0-9]-[0-3][0-9]))"  # up to 2039-19-39
-RFC3339_DATETIME_REGEX = RFC3339_DATE_REGEX + r"[T ]([0-2][0-9]:[0-5][0-9]:[0-5][0-9](?:\.\d+)?)(Z|[\+-]\d{2}:\d{2})?$"
+RFC3339_DATE_REGEX = r"^[0-9]{4}-[01][0-9]-[0-3][0-9]"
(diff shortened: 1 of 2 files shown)


src/lamp_py/ingestion/glides.py
```diff
@@ -65,7 +65,9 @@ class GlidesRecord(dy.Schema):
     id = dy.String()
     type = dy.String()
     time = dy.Datetime(  # in %Y-%m-%dT%H:%M:%S%:z format before serialization
-        min=datetime(2024, 1, 1), max=datetime(2039, 12, 31), time_unit="ms"  # within Python's serializable range
+        min=datetime(2024, 1, 1),
+        max=datetime(2039, 12, 31),
+        time_unit="ms",  # within Python's serializable range
```
runkelcorey: this is just formatting

huangh (changes requested):
i think we have technically incorrect strictness that violates upstream schema. probably match theirs exactly is the best move. 

same with the `time` field...also those restrictions make it not RFC3339 compliant

huangh (approved):

andrewdolce (approved):
Looks good to me 🚀 

> I think the real question is how we treat optional upstream. We have chosen - by the limits of our dataframe library - to represent optional as null. for downstream, it doesn't matter - no data. for upstream - if someone asks if it was not available b/c optional, or it was there but was filled in with a null, I think that is a question to send back to glides.

FWIW, Glides currently should never emit an explicit null for this field, which I believe matches the intention in the spec.

PR #724, opened 2026-04-02
feat(ingestion): ingest `state_of_charge_timestamp` and `_percentage`

runkelcorey:
[🚎 Ingest BEB state_of_charge_percentage and state_of_charge_timestamp](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1213794561554639?focus=true)

1. Adds `state_of_charge_timestamp` and `state_of_charge_percentage` to `BUS_VEHICLE_POSITIONS`
2. Refactors BusLoc schema using dataframely
3. Updates ingestion to coerce existing data to use specified schema

1. Updated BusLoc ingestion test to use dataframely schema for record generation
2. Deployed to `dev`, saw [no errors](https://mbta.splunkcloud.com/en-US/app/search/search?q=%7C%20%40spl2%20search%20index%3Dlamp-dev%20error_type%20NOT%20ValidationError&display.page.search.mode=fast&dispatch.sample_ratio=1&workload_pool=&earliest=-60m%40m&latest=now&sid=1775144331.85354) and expected data in `arn:aws:s3:::mbta-ctd-dataplatform-springboard/lamp/BUS_VEHICLE_POSITIONS/year=2026/month=4/day=2/2026-04-02T00:00:00.parquet`

1. Should we add logic (in a different PR) to handle the case where the specified schema represents a subset of the existing records? If we rolled back this change, then I think our ingestion would break because the specified schema wouldn't contain all the columns of the input dataset


tests/ingestion/test_gtfs_rt_converter.py
```diff
@@ -256,42 +262,51 @@ def test_rt_trip_file_conversion() -> None:
...
+    assert timestamp
+    assert timestamp.month == vehicle_timestamp.month
+    assert timestamp.year == vehicle_timestamp.year
+    assert timestamp.day == vehicle_timestamp.day
+
+    assert filename == str(incoming_file)
+
+    assert table
+    table = pl.DataFrame(flatten_table_schema(table))
+    assert (
```
huangh, line 301: 👍

src/lamp_py/ingestion/config_busloc_vehicle.py
```diff
@@ -1,14 +1,116 @@
...
+                    inner={
+                        "position": position,
+                        "location_source": dy.String(nullable=True),
+                        "timestamp": dy.UInt64(nullable=True),
+                        "trip": trip_descriptor,
+                        "vehicle": vehicle_descriptor,
+                        "operator": dy.Struct(
+                            inner={
+                                "id": dy.String(nullable=True),
+                                "first_name": dy.String(metadata={"reader_roles": ["OperatorName"]}, nullable=True),
```
huangh, line 38: this bit feels like scope creep for this PR. I see the intent with the metadata...but is this the granularity/how we plan to handle our pii architecture with these tags?

src/lamp_py/ingestion/config_busloc_vehicle.py
```diff
@@ -1,14 +1,116 @@
...
+                        "state_of_charge_timestamp": dy.UInt64(nullable=True),
+                    }
+                ),
+            }
+        ),
+        min_length=1,
+    )
+
+
+class BusLocVehicleTable(dy.Schema):
```
huangh, line 65: was it an active choice to define to define a completely separate flat version of the nested struct, or should we call one of the many `flatten` implementations we already have to autogen this? this is a lot of LOC for low benefit. 

Or we drop the nested implementation and pump the json straight to flat. I don't think we want both, as a general rule (and if you agree, we should apply to all the other ingestion structs to clean up)

src/lamp_py/ingestion/gtfs_rt_structs.py
```diff
@@ -1,71 +1,80 @@
...
-    ]
+
+class GTFSRealtime(dy.Schema):
+    """Fields common to all GTFS Realtime messages."""
+
+    header = dy.Struct(
+        inner={
+            "gtfs_realtime_version": dy.String(),
+            "incrementality": dy.String(),
+            "timestamp": dy.UInt64(),
```
huangh: this is missing `feed_version` - though i think this was added really recently so we might not have updated our internal docs to reflect. 

https://gtfs.org/documentation/realtime/reference/#message-feedheader

src/lamp_py/ingestion/gtfs_rt_structs.py
```diff
@@ -1,71 +1,80 @@
...
-        ("odometer", pyarrow.float64()),
-    ]
+
+class GTFSRealtime(dy.Schema):
+    """Fields common to all GTFS Realtime messages."""
+
+    header = dy.Struct(
+        inner={
+            "gtfs_realtime_version": dy.String(),
+            "incrementality": dy.String(),
```
huangh: I think the only valid string is currently `FULL_DATASET`, although `DIFFERENTIAL` is also (and the only other) allowed value. this is actually an enum, not a string. i'm not sure what's going on upstream to stringify, but...we should handle accordingly

src/lamp_py/ingestion/config_busloc_vehicle.py
```diff
@@ -1,14 +1,116 @@
...
 )
 
 
+class BusLocVehicleRecord(GTFSRealtime):
+    """Each VehiclePosition message generated by BusLoc."""
+
+    entity = dy.List(
+        inner=dy.Struct(
+            inner={
+                "id": dy.String(min_length=1),
```
huangh, line 26: does nullable=False not imply min_length>=1? Why did we choose min_length over just non-nullable aka required?

huangh (changes requested):

src/lamp_py/ingestion/config_busloc_vehicle.py
runkelcorey: `""`

src/lamp_py/ingestion/config_busloc_vehicle.py
runkelcorey: as a general rule, I actually disagree: eventually, I want dataframely schemas for every dataset we materialize. `BusLocVehicleRecord` represents our incoming JSONs and `BusLocVehicleTable` represents our springboard tables. They are different because our springboard datasets are not necessarily perfectly flat versions of incoming records.

in this case, I initially tried to create this class using `lamp_py.utils.dataframely.unnest_columns` but the springboard dataset does not flatten `consist` so I couldn't use that function. Our other `flatten` implementations don't provide a dataframely schema

src/lamp_py/ingestion/gtfs_rt_structs.py
runkelcorey: yes but that's unfortunately not what BusLoc provides. if you try
```python
pl.read_ndjson("s3://mbta-ctd-dataplatform-archive/lamp/delta/2026/04/07/2026-04-07T00:00:00Z_https_mbta_busloc_s3.s3.amazonaws.com_prod_VehiclePositions_enhanced.json.gz").select("header").unnest("header")
```
you'll get back
|timestamp	|gtfs_realtime_version	|incrementality|
|-|-|-|
|i64	|str	|i64|
|1775519999	|"2.0"	|0|

runkelcorey: @dks-mbta is updating the spec for busloc somehwere on your backlog?

src/lamp_py/ingestion/gtfs_rt_structs.py
runkelcorey: also, unfortunately, not what BusLoc provides. let me remove this struct (for now) and make a header just for BusLoc since I think it's the only GTFS-RT message that won't match the spec

src/lamp_py/ingestion/config_busloc_vehicle.py
runkelcorey: it's what I did [in Glides](https://github.com/mbta/lamp/blob/6a9944756fc623f6870cc512323e725b0f9c4895/src/lamp_py/ingestion/glides.py#L29-L30) but, since that's slipping down the priority list, maybe we shouldn't touch this for now? I can remove here and in our existing code?

My plan was:

1. tag the column(s) with a descriptive role
2. write column(s) in each role to a new prefix in `lamp-restricted`, eg. `lamp-restricted/OperatorName/BUS_VEHICLE_POSITIONS/etc.parquet`
2. map the prefix to an IAM policy

src/lamp_py/ingestion/config_busloc_vehicle.py
huangh: Fair. Thinking on what you said here, I'll modify my complaint. I think the real issue here might be that we're using dataframely to validate json, when we should probably be using a json validator for a json schema. I think we've somewhat discussed this before where a lack of a value in json we are interpreting as nulls in the data frame schema, but we know that's not quite right. 

I think that's the smell I'm smelling. What do you think

src/lamp_py/ingestion/gtfs_rt_structs.py
huangh: Following from above, this header message does match spec, because this field is optional. it's just that dataframely is strict and doesn't allow.

src/lamp_py/ingestion/config_busloc_vehicle.py
runkelcorey: It seems like to do this, we'd

1. Write [pydantic](https://docs.pydantic.dev/latest/#pydantic-examples) classes
2. Use [polyfactory](https://polyfactory.litestar.dev/latest/usage/declaring_factories.html) to create test data

I'm on board with validating our incoming JSONs but it's hacky unless we do it for all GTFS-RT messages because of how `converter.detail.import_schema` is called. For that reason, are you ok if we punt it to its own ticket?

huangh (approved):

src/lamp_py/ingestion/gtfs_rt_structs.py
dks-mbta: @runkelcorey We have an inbound Busloc ticket to address this.

PR #728, opened 2026-04-10
fix(ingestion): hash method only ever returned the new portion of data

huangh:
from #724 

no backfill for live data required as this had not been deployed to prod


runkelcorey (approved):
thank you so much for this. I added [a ticket](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1214000364253089?focus=true) to harden our ways of preventing this happens again

PR #730, opened 2026-04-13
chore(ingestion): Delete light_rail_gps code

runkelcorey:

Deletes unused light rail GPS code. This code showed up in some searches I was doing and I thought "why not cut our codebase down a bit?"

Does not delete the light rail GPS remote file location, which is referenced by lightswitch.

1. `pylint` and `pytest` don't complain
2. dev ingestion still ingests

None.


huangh (approved):
i'm ok with this in general, but seeing as how type 10's are coming in imminently, we should at least fully tie this in with an asana task bookkeeping so we don't lose track of this when it comes back for type 10. 

or...we can leave it since it's been cruft for a year, and will be re-enabled with type 10 (i think?)

runkelcorey:
Created [a new task for Type 10 data](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1214072337230339?focus=true)

PR #731, opened 2026-04-14
fix(rail): fix(rail): Remove ferry (`route_type` `4`) from RPM

runkelcorey:
[Expand enumeration to enable Ferry Data to enter LAMP](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1213911472207618)

1. Filters out ferry trips from Rail PM. In the short term, keeping ferry in RPM may have given our users quick access to ferry service metrics. However...it's not rail. Plus, I want to avoid creating more dependencies on RPM, especially since we have no users looking at ferry afaik.

1. Updates tests to expect no ferry trips. I changed the tests by checking which routes, patterns, trips, and stop times were ferry related and removing those. Tests passing indicates the only changes were ferry.

1. Do you agree that we ought to remove ferry from RPM?


tests/performance_manager/test_performance_manager.py
```diff
@@ -442,14 +442,14 @@ def test_static_tables(
     # processing the static tables our db tables should have these many record
     # counts.
     row_counts = {
-        StaticTrips: 11709,
-        StaticRoutes: 24,
+        StaticTrips: 10993,
```
huangh, line 445: have we been including ferry this whole time? i didn't think there was any ferry data to take in, so where's this discrepancy coming from?

runkelcorey: We've had ferry in the _scheduled_ routes, patterns, and trips

huangh (approved):
silly 🚢

PR #732, opened 2026-04-14
feat(ingestion): Add row count to logged metadata

runkelcorey:
[🦺 make it harder to delete all the data](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1214000364253089)

For GTFS-RT files uploaded to s3:
1. Adds row count of file so that these metrics appear in Splunk
2. Adds [serialized] size of file

1. Deployed to `dev`, saw that ingestion didn't stop and that row count matched row count calculated from lightswitch

1. Should i add this to the other converters?


huangh:
lets test the ergonomics of this and the alert in gtfs first, then move it to others when needed. this will already affect 99% of our data. the missing bits not checked are gtfs-static, glides, and transitmaster.

huangh (approved):

PR #735, opened 2026-05-04
chore(ci): Run zizmor on each PR

runkelcorey:
[Asana Task](https://app.asana.com/1/15492006741476/project/1113179098808463/task/1214348962087612?focus=true)

Updates existing workflows to address Zizmor's concerns using:
* regular persona
* a new action [I made in mbta/actions](https://github.com/mbta/actions/compare/feat-zizmor)

1. This PR running *should* validate them 🤞
2. It's possible some of our actions are broken after this and require minor remediation. I can take that gamble

1. I only fixed errors thrown by the auditor persona. That solved all of the dependency pins and the more egregious permissions / secret inheritance but leaves comments like
```
warning[excessive-permissions]: overly broad permissions
   --> .github/workflows/deploy-base.yaml:8:1
    |
  8 | / on:
  9 | |   workflow_call:
 10 | |     inputs:
 11 | |       environment:
...   |
183 | |           webhook-url: ${{ secrets.SLACK_WEBHOOK }}
184 | |           job-status: ${{ job.status }}
    | |________________________________________^ default permissions used due to no permissions: block
    |
    = note: audit confidence → Medium

help[undocumented-permissions]: permissions without explanatory comments
  --> .github/workflows/deploy-base.yaml:70:7
   |
70 |       id-token: write
   |       ^^^^^^^^^^^^^^^ needs an explanatory comment
   |
   = note: audit confidence → High

help[anonymous-definition]: workflow or action definition without a name
   --> .github/workflows/deploy-base.yaml:8:1
    |
  8 | / on:
  9 | |   workflow_call:
 10 | |     inputs:
 11 | |       environment:
...   |
183 | |           webhook-url: ${{ secrets.SLACK_WEBHOOK }}
184 | |           job-status: ${{ job.status }}
    | |________________________________________^ this workflow
    |
    = note: audit confidence → High
    = tip: use 'name: ...' to give this workflow a name
```
How do we feel about these? Personally, I'm satisfied by how constrained the permissions now are but YMMV.


.github/workflows/ci-yaml.yaml
```diff
@@ -10,11 +10,19 @@ on:
...
+  contents: read
+
 jobs:
   build:
     name: Validate YAML actions
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9  # v3
```
skyqrose: actions/checkout is up to v6 now, and I think there are some reasons to upgrade. Maybe it was old versions of checkout had an EOL dependency or something? Might as well do that upgrade now.

There are other updates, too. cache is up to v5. I haven't checked beyond that.

.github/workflows/ci-yaml.yaml
```diff
@@ -10,11 +10,19 @@ on:
...
+        with:
+          persist-credentials: false
       - run: yamllint . -f parsable --strict
+      - uses: mbta/actions/zizmor@e335b8b5c95e2bef83d0af86058279ba996e572d
```
skyqrose: I don't think putting it in mbta/actions really adds much. It's two steps, checkout and zizmor, and you already have to checkout here anyway.

It does set a couple defaults, but idk if having better defaults for those is worth the extra layer of dependency management.

.github/workflows/ci-yaml.yaml
```diff
@@ -10,11 +10,19 @@ on:
...
+        with:
+          minimum_issue_severity: medium
```
skyqrose: The PR description says "regular persona", but `mbta/actions/zizmor` says its default is "auditor". ([Plain zizmor's default is "regular".](https://github.com/zizmorcore/zizmor-action#persona))

What's your opinion on regular vs auditor, for this project and for TID projects in general?

And any opinions on other inputs?

.github/workflows/ci_python.yaml
```diff
@@ -19,6 +9,10 @@ on:
       - 'poetry.lock'
       - '.github/workflows/ci_python.yaml'
       - '.github/python_deps/action.yaml'
+  workflow_call:
```
skyqrose, line 12: This seems like a meaningful change, to not run CI on push anymore. Was there something that zizmor recommended that made you change this?

.github/workflows/deploy-prod.yaml
```diff
@@ -35,5 +43,17 @@ jobs:
...
+      AWS_ROLE_ARN: ${{ secrets.AWS_ROLE_ARN }}
+  run-lightswitch:
+    needs: deploy
+    uses: ./.github/workflows/run-task.yaml
+    with:
+      environment: prod
+      task: Lightswitch
+    permissions:
+      contents: read
+      id-token: write
```
skyqrose, line 59: There are some lightswitch changes in this PR.

- Here and in `deploy-staging.yml`:
  - Changes `deploy-lightswitch: false` to `deploy-lightswitch: true`
  - Adds a new `run-lightswitch` job
- Deletes `lightswitch_deploy_run.yaml`

Why? Was it prompted by zizmor?

.github/dependabot.yaml
```diff
@@ -8,3 +8,5 @@ updates:
       time: "08:00"
       timezone: "America/New_York"
     open-pull-requests-limit: 5
+    cooldown:
+      default-days: 7
```
skyqrose, line 12: <3 This is something we should be doing in all of our repos.

Did zizmor remind you about it, or did you just notice it while looking through the .github files?

analysis/export_hr_dataset.qmd
```diff
@@ -0,0 +1,256 @@
+---
+title: "Exporting [a lot of] Heavy Rail TripUpdates"
```
skyqrose: I don't think this belongs in this PR.

.github/workflows/deploy-base.yaml
skyqrose, line 67: > I only fixed errors thrown by the auditor persona. That solved all of the dependency pins and the more egregious permissions / secret inheritance but leaves comments like

```
warning[excessive-permissions]: overly broad permissions
   --> .github/workflows/deploy-base.yaml:8:1
    |
  8 | / on:
  9 | |   workflow_call:
 10 | |     inputs:
 11 | |       environment:
...   |
183 | |           webhook-url: ${{ secrets.SLACK_WEBHOOK }}
184 | |           job-status: ${{ job.status }}
    | |________________________________________^ default permissions used due to no permissions: block
    |
    = note: audit confidence → Medium

help[undocumented-permissions]: permissions without explanatory comments
  --> .github/workflows/deploy-base.yaml:70:7
   |
70 |       id-token: write
   |       ^^^^^^^^^^^^^^^ needs an explanatory comment
   |
   = note: audit confidence → High

help[anonymous-definition]: workflow or action definition without a name
   --> .github/workflows/deploy-base.yaml:8:1
    |
  8 | / on:
  9 | |   workflow_call:
 10 | |     inputs:
 11 | |       environment:
...   |
183 | |           webhook-url: ${{ secrets.SLACK_WEBHOOK }}
184 | |           job-status: ${{ job.status }}
    | |________________________________________^ this workflow
    |
    = note: audit confidence → High
    = tip: use 'name: ...' to give this workflow a name
```

Here's some general opinions on this:

In general, we should follow zizmor's guidance. If there's a reason that we think zizmor is stricter than want to be, that's fine, as long as we document it (either at the location, or in [Notion](https://www.notion.so/mbta-downtown-crossing/Zizmor-GHA-audit-350f5d8d11ea8021bc7bc0c3fd6de6a4)).

We should also make sure that the output is clean. i.e. any warnings we choose to ignore are actually suppressed, rather than left in the output for a human to see an ignore, so that if there are new and relevant errors, they don't get lost in the visual noise.

For the three specific warnings.

1. This does look like a real warning that we should follow (principle of least-privilege).
2. I see the reasoning. You don't want to leave an obsolete permission unlocked forever just because you don't remember why it was there to begin with. But it could turn into busywork. This rule is under the `--pedantic` option. I think it's up to the team whether they want to turn pedantic rules on. From my armchair, I'd recommend it, but I'm not the one implementing it.
3. This one is also under `pedantic`. The docs aren't super clear, but here's where it shows up: In the [actions UI](https://github.com/mbta/lamp/actions/runs/24725791567/job/72326347532), this job appears under its yaml key `deploy` instead of a human-written name like "Deploy to Staging" has. It's up to the team whether pedantic rules are on or off.

GitHub actions run. Job name: 'Deploy to Staging / deploy'

Based on your experience, if you have a recommendation for other teams on whether to have pedantic rules on or off, please document it in Notion.

.github/workflows/ci-yaml.yaml
runkelcorey: I was thinking that
1. it makes migrating a teeny bit easier across the org
2. running a single action across repos opens more options for y'all to enforce usage of zizmor

The only drawback I see is that choosing the settings yourself can folks reflect on their actions but I feel like the act of remediating those issues does the same.

Do you think it makes y'all's work easier (now or in the future)?

.github/workflows/ci-yaml.yaml
runkelcorey: I chose `auditor` because I don't understand what `persona` changes while `min-severity` had more legible differences. So, these defaults gave me a better sense that I knew what I was excluding vs. including

analysis/export_hr_dataset.qmd
runkelcorey: yeah big oops

.github/workflows/ci_python.yaml
runkelcorey: yes! zizmor said that `workflow_run` is a [dangerous trigger](https://docs.zizmor.sh/audits/#dangerous-triggers)
>because they run in the context of the target repository rather than the fork repository, while also being typically triggerable by the latter. This can lead to attacker controlled code execution or unexpected action runs with context controlled by a malicious fork.

The probability of an attack like that in Lamp is probably not huge but this repo is open-source so someone could open a PR. I also like this change from an architectural perspective because refactoring as `workflow_call` makes them easier to version and control

.github/workflows/ci-yaml.yaml
skyqrose: The action would make it easier to enforce how people use zizmor. But I'm less worried about that and more worried about enforcing that people use zizmor, which the action doesn't help at all with. We'd still need to enforce that people use our action instead.

Updating also isn't easier cuz each repo would still need to update the version of the action that they use.

Also, with the action, other teams need to learn the interface of zizmor and also the interface of our custom action. It works well if you can hide some implementation details behind our action, but you can't really in this case.

So I'm leaning towards the extra layer not making it any easier.

.github/workflows/deploy-prod.yaml
runkelcorey: This change was prompted by zizmor but motivated by another practical consideration. `lightswitch_deploy_run.yaml` also used the `workflow_run` trigger, so zizmor flagged it for remediation. When considering how to remediate, I found it a bit more concise to consolidate it as a step in deployment rather than its own workflow. The effect is that lightswitch is built and its task run more often but that change is probably good because the current state causes problems in some current edge cases.

.github/dependabot.yaml
runkelcorey: tbh never knew the setting existed! zizmor [autofixed](https://docs.zizmor.sh/usage/#auto-fixing-results) it

.github/workflows/ci-yaml.yaml
skyqrose: Oh, the docs don't explain how persona, min-severity, and min-confidence interact at all. It appears they are three entirely separate scales, each issue is marked with a persona, severity, and confidence. And it only returns issues that pass all the filters you give it. That's... weird.

My guess is that persona is meant to be based on both the severity and confidence. E.g. for the regular persona, a high-severity issue might be included at medium-confidence, but it wouldn't include a low-severity issue unless it had high-confidence. (Those are thresholds I made up, I haven't verified this idea.)

If my guess is right, then I think using personas is more likely to give you relevant results than setting up your own severity and confidence filters, and also the docs make it seem like the intended way to use the tool, so I'd lean towards using personas. But that's just a guess and a suggestion; it's still the team's choice as to what they think is relevant for them.

.github/workflows/ci_python.yaml
skyqrose: That rule is about `workflow_run` (which I see was removed from `deploy-staging.yml`). The trigger that was removed from this file is `push`, which is safe.

.github/workflows/ci-yaml.yaml
runkelcorey: ohhhh I thought y'all could require some workflows to be run across an org but apparently you can only do that with [third-party integrations](https://docs.github.com/en/code-security/concepts/code-scanning/about-code-scanning). Okay, I'll make a new workflow that's just in lamp 🫡

.github/workflows/ci_python.yaml
runkelcorey: yup! but removing `workflow_run` from `deploy-staging` while still requiring `ci_python` to succeed before deployment means refactoring to trigger `ci_python` from `deploy-staging` instead of triggering `deploy-staging` from `ci_python`'s run

ultimately, I want
1. `ci_python` to succeed
2. `deploy-staging` to run

skyqrose: We talked offline about this:
1. I mixed up workflow_dispatch and workflow_call.
2. I didn't see that it was referenced from deploy-staging, and swapping `push` for `workflow_call` was a way to be sure ci passes before the deploy.

The change makes sense and is fine.

I think most teams run ci and deploy-staging in parallel on push to main, which is simpler, but has one less layer of protection if bad code somehow makes its way to main, so this approach is good, too.

.github/workflows/ci-yaml.yaml
skyqrose: Oh, I didn't realize that was an option. Zizmor does integrate with GitHub, so maybe we could do that, but only if we pay for the extra advanced security option, which we don't yet.

So for now, every repo has to add the workflow one-by-one. But that's okay cuz they also need to do a code change in the repo to fix any issues, so the marginal extra effort of adding the workflow is pretty small.

If we do ever use that option in the future, we can set up a shared action (or whatever else is needed) then.

.github/workflows/deploy-base.yaml
runkelcorey: Okay, I'll update the Notion!

skyqrose (commented):
LGTM from my perspective, though I'm not giving an official PR approval cuz that should come from the lamp team

.github/workflows/deploy-prod.yaml
```diff
@@ -6,13 +6,18 @@ on:
...
 jobs:
   validate-tag:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     steps:
       - name: Check Semantic Version Format
         run: |
-          TAG_NAME="${{ github.ref_name }}"
+          TAG_NAME="${GITHUB_REF_NAME}"
```
huangh, line 20: notes: https://docs.github.com/en/actions/reference/workflows-and-actions/variables#default-environment-variables

.github/dependabot.yaml
huangh: this has been making the news recently..good change

huangh (approved):
only one way to find out! lgtm

PR #736, opened 2026-05-05
fix(aws): Don't overwrite parquet files if they have fewer records

runkelcorey:
[Asana Task](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1214494536112678?focus=true)

1. Adds `replace_remote_parquet` which:
* checks if the remote file exists
    * if not, calls `upload_file`
    * if so, checks if the local files has at least as many records as the remote file
        * if not calls `upload_file`
        * if so, replaces file
2. Replaces `upload_file` with `replace_remote_parquet` in Glides ingestion (the guilty party) and alerts ingestion

1. New unit test for `replace_remote_parquet`
2. Runs on `dev`; I reverted implementing this in GTFS-RT ingestion because currently our ingestion fails this test by uniquing the existing dataset

1. Do you agree that triggering an error is a good way to handle this issue?


src/lamp_py/aws/s3.py
runkelcorey, line 1: All changes until the last function are just `ruff`

src/lamp_py/ingestion/convert_gtfs_rt.py
runkelcorey, line 1: `ruff` changes only—I can revert this if you really want to. Previous commits had changes to this file

src/lamp_py/ingestion/glides.py
```diff
@@ -249,7 +249,7 @@ def append_records(self) -> None:
 
     def upload_records(self) -> None:
         """Upload local parquet file to remote storage."""
-        upload_file(file_name=self.local_path, object_path=self.remote_path)
+        replace_remote_parquet(file_name=self.local_path, object_path=self.remote_path)
```
runkelcorey, line 252: this is the only substantive change, the rest are `ruff`

tests/aws/test_s3_utils.py
```diff
@@ -1,30 +1,31 @@
-# pylint: disable=[W0621, W0611]
+# pylint: disable=[W0621, W0611, R0913, R0917]
```
runkelcorey, line 1: too many [positional] arguments, which I don't really care for when it's test fixtures

src/lamp_py/aws/s3.py
```diff
@@ -743,3 +741,37 @@ def read_parquet_chunks(
...
+    """
+    process_logger = ProcessLogger("replace_remote_parquet", file_name=file_name, object_path=object_path)
+    process_logger.log_start()
+
+    object_path = object_path.replace("s3://", "")
+
+    if object_exists(object_path):
+        existing_row_count = pq.read_metadata(object_path, filesystem=fs.S3FileSystem()).num_rows
+        new_row_count = pq.read_metadata(file_name).num_rows
+        assert new_row_count >= existing_row_count, f"{new_row_count} < {existing_row_count}; cancelling upload."
```
huangh: I think we shouldn't use asserts to check this - just log error and set the error code and return false. i think if the assert fails in prod, we have to do a lot to make sure it's bubbled up and caught somewhere, so easier to just log error instead imo

src/lamp_py/aws/s3.py
```diff
@@ -743,3 +741,37 @@ def read_parquet_chunks(
...
+        process_logger.add_metadata(existing_row_count=existing_row_count, new_row_count=new_row_count)
+
+    result = upload_file(file_name, object_path, extra_args)
+    assert result, "upload_file failed"
```
huangh: same as above

tests/aws/test_s3_utils.py
```diff
@@ -154,3 +154,39 @@ def test_move_bad_objects(s3_stub, caplog):  # type: ignore
...
+    pl.int_range(0, local_records, eager=True).to_frame().write_parquet(local_file.as_posix())
+
+    remote_file = tmp_path / "remote_foo.parquet"
+    pl.int_range(0, 10, eager=True).to_frame().write_parquet(remote_file.as_posix())
+
+    with patch("lamp_py.aws.s3.object_exists", return_value=remote_file_exists):
+        with patch("pyarrow.fs.S3FileSystem", return_value=LocalFileSystem()):
+            with patch("lamp_py.aws.s3.upload_file", return_value=upload_succeeds):
+                with raises:
+                    replace_remote_parquet(local_file.as_posix(), "s3://" + remote_file.as_posix())
```
huangh: comment: we should really start using types or pathlib/something to specify the s3 resources and local paths so we don't have to do string joining stuff. its fragile

huangh (changes requested):

PR #738, opened 2026-05-05
feat(lightswitch): add helper function for converting unix epoch seconds to timestamp with timezone

rudiejd:
Asana Task: [🔮 Don't filter out light rail trains that have been stopped for more than 10 minutes](https://app.asana.com/1/15492006741476/project/584764604969369/task/1207658140982515?focus=true)

Problem: 

As demonstrated by the macro I created in this PR, the DuckDB syntax for converting unix seconds since  epoch to a datetime with timezone is not very ergonomic. I spend a lot of time recreating this logic in my notebook. 

Solution:

Add a `to_timestamptz` macro, similar to `to_timestamp`, which converts unix seconds since epoch to a datetime with timezone information. Default to using `America/New_York` as the IANA time zone. 

Added new tests for Lightswitch

1. Is this something that should be contributed upstream to DuckDB?

### DIFF
--- a/tests/publishing/test_lightswitch.py
+++ b/tests/publishing/test_lightswitch.py
@@ -110,7 +110,7 @@ def test_totimestamptz(duckdb_con: duckdb.DuckDBPyConnection) -> None:
     with duckdb_con:
         register_to_timestamptz(duckdb_con)
         timestamp_tz = duckdb_con.execute("SELECT to_timestamptz(0, 'Etc/UTC')").pl().row(0)[0]
-        assert timestamp_tz == datetime.fromtimestamp(0, tz=ZoneInfo('Etc/UTC'))
+        assert timestamp_tz == datetime.fromtimestamp(0, tz=ZoneInfo("Etc/UTC"))
 
 
 def test_register_effective_gtfs_timestamps(
--- a/tests/publishing/test_lightswitch.py
+++ b/tests/publishing/test_lightswitch.py
@@ -1,8 +1,10 @@
+from zoneinfo import ZoneInfo
 import os
 from contextlib import nullcontext
 from logging import ERROR
 from pathlib import Path
 from typing import Generator
+from datetime import datetime
 
 import polars as pl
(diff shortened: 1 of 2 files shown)


tests/publishing/test_lightswitch.py
```diff
@@ -102,6 +105,14 @@ def test_register_read_ymd(
             )
 
 
+def test_totimestamptz(duckdb_con: duckdb.DuckDBPyConnection) -> None:
+    """It converts unix timestamp to datetime with timezone"""
+    with duckdb_con:
```
rudiejd, line 110: somehow this passes locally but fails in CI 🤔 

the timezone of the datetime returned by `from_timestamp` becomes `Etc/UTC` instead of `America/New_York` when run in CI

rudiejd: I think this is probably because duckdb is somehow not getting the system timezones on the GitHub CI runner. I'll just use Etc/UTC for the test

runkelcorey: Is it possible that the `:: TIMSTAMPTZ` cast isn't working? When I run `SELECT (to_timestamp(0) AT TIME ZONE 'America/Los_Angeles') :: TIMESTAMPTZ`, I get `1969-12-31 16:00:00-05`. When I run  `SELECT (to_timestamp(0) AT TIME ZONE 'America/New_York') :: TIMESTAMPTZ` I get `1969-12-31 19:00:00-05`—different times but same time zone (my system's)

rudiejd:
I just realized that this function isn't necessary... maybe a misunderstanding on my part. 

from the docs:
image
image


I think I just didn't have ICU loaded when `TO_TIMESTAMP` was not returning a `TIMESTAMP WITH TIME ZONE`. still confusing, but doesn't warrant a helper

runkelcorey:
Frankly not great docs and not very intuitive for TIMESTAMPTZ to always be UTC

PR #739, opened 2026-05-07
feat(deps): Randomize test order and test seed

runkelcorey:

This PR adds `pytest-randomly`, which randomizes the order that tests run as well as the seed, and `pytest-order`, which allows us to control the order of tests in specific cases (eg. when tests write their state to the test performance manager database and need to retrieve that state later).

1. New tests?
2. Used a runner?
3. Doesn't need validation? Maybe it's just documentation.

1. Focus on this thing, even though it seems minor!


runkelcorey:
There's another refactor afoot

PR #743, opened 2026-05-08
feat(deps): update dataframely & add pytest plugins

runkelcorey:

1. Adds `pytest-randomly`, which randomizes the order that tests run as well as the seed
2. Adds`pytest-order`, which allows us to control the order of tests in specific cases (eg. when tests write their state to the test performance manager database and need to retrieve that state later).
3. Updates dataframely
4. Replaces bespoke functions with dataframely equivalents

1. CI
5. Ran glides ingestion locally to ensure that schema updates aligned with reality

1. None.


src/lamp_py/bus_performance_manager/combined_bus_schedule.py
```diff
@@ -30,6 +27,14 @@ class CombinedBusSchedule(GTFSBusSchedule):
     waiver_remark = TransitMasterSchedule.waiver_remark
 
 
+class CombinedBusSchedule(CombinedBusScheduleCore):
```
huangh, line 30: what is the logic for breaking this out? i'm sure there is some...is this a place for some documentation or a markdown table showing why each of these fields are here and not there? I don't understand the delineation between what is core and what is not-core.

src/lamp_py/bus_performance_manager/events_gtfs_schedule.py
```diff
@@ -16,10 +16,9 @@ class BusBaseSchema(dy.Schema):
     route_id = dy.String(primary_key=True)
 
 
-class GTFSBusSchedule(BusBaseSchema):
+class GTFSBusScheduleCore(BusBaseSchema):
```
huangh, line 19: so there's Base -> Core -> Schedule?

src/lamp_py/bus_performance_manager/events_tm.py
```diff
@@ -18,18 +18,23 @@
...
     tm_actual_arrival_dt = dy.Datetime(nullable=True, time_zone="UTC")
     tm_actual_departure_dt = dy.Datetime(nullable=True, time_zone="UTC")
     tm_actual_arrival_time_sam = dy.Int64(nullable=True)
     tm_actual_departure_time_sam = dy.Int64(nullable=True)
-    vehicle_label = dy.String(nullable=True)
-    tm_stop_sequence = dy.Int64(primary_key=True)
     tm_pullout_id = dy.String(primary_key=True)
 
 
+class TransitMasterEvents(TransitMasterEventsCore):
```
huangh, line 31: this has overlap with `CombinedBusSchedule`. Should `CombinedBusSchedule` multiple inherit from these other events cores?

src/lamp_py/tableau/jobs/glides.py
```diff
@@ -1,38 +1,44 @@
...
-    trip_service_date = dy.Date(alias="data.tripUpdates.tripKey.serviceDate", nullable=True)
-
-
-class OperatorSignInsTableau(OperatorSignInsTable):  # type: ignore[misc, valid-type]
-    """Glides Operator Sign-Ins data, transformed for Tableau consumption."""
-
-    input_timestamp = dy.Datetime(time_unit="ms", alias="data.metadata.inputTimestamp", nullable=True)
-    signed_in_at = dy.Datetime(time_unit="ms", alias="data.signedInAt", nullable=True)
-    time = dy.Datetime(time_unit="ms", alias="time", nullable=True)
+OperatorSignInsTableau: Type[GlidesRecord] = type(
```
huangh, line 32: what does this additional typing do for you that the prior implementation didn't do? Is there some formatting weirdness that happened here? Also, It smells like there's an extra comma that led it to turn GlidesRecord into a tuple...which seems like really strange syntax if it's actual proper.  `(GlidesRecord,),`???

Tl;dr - i have no idea what's going on here. (my) skill issue.

analysis/flashback_performance.py
runkelcorey, line 1: black

huangh (changes requested):
A lot of these changes I'm unsure what they're for. I think I see the core dataframely change to allow with_nullable to be removed, but the rest of the changes are a little mysterious. 

I think the organization of the dy schemas needs explaining, because they seem to be overlapping now and I don't understand why any particular schema gets `core` designation or not.

runkelcorey:
thanks for that comment. my dependencies drifted from our prod and errors that I was seeing locally were not the same as I saw in CI. let me (again) close this in favor of another

PR #744, opened 2026-05-08
fix(aws): do not overwrite remote parquets if fewer records

runkelcorey:
[🐞 Glides trip updates truncated in prod](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1214494536112678)

1. Adds `replace_remote_parquet` which:
* checks if the remote file exists
    * if not, calls `upload_file`
    * if so, checks if the local files has at least as many records as the remote file
        * if so, calls `upload_file`
        * if not, logs error and returns `False`
2. Replaces `upload_file` with `replace_remote_parquet` in Glides ingestion (the guilty party) and alerts ingestion

1. New unit test for `replace_remote_parquet`
2. Runs on `dev`; I reverted implementing this in GTFS-RT ingestion because currently our ingestion fails this test by uniquing the existing dataset

1. None.

### DIFF
--- a/src/lamp_py/aws/s3.py
+++ b/src/lamp_py/aws/s3.py
@@ -7,9 +7,9 @@ from datetime import date, datetime, timezone
 from io import BytesIO
 from threading import current_thread
 from typing import (
+    IO,
     Callable,
     Dict,
-    IO,
     Iterator,
     List,
     Optional,
@@ -20,12 +20,12 @@ from typing import (
 import boto3
 import botocore
 import botocore.exceptions
-from botocore.exceptions import ClientError
 import pandas
 import pyarrow as pa
 import pyarrow.compute as pc
-import pyarrow.parquet as pq
 import pyarrow.dataset as pd
+import pyarrow.parquet as pq
+from botocore.exceptions import ClientError
 from pyarrow import Table, fs
 from pyarrow.util import guid
 
@@ -124,7 +124,7 @@ def download_file(object_path: str, file_name: str) -> bool:
 
 def delete_object(del_obj: str) -> bool:
     """
-    delete s3 object
+    Delete s3 object
 
     :param del_obj - expected as 's3://my_bucket/object' or 'my_bucket/object'
 
@@ -158,7 +158,7 @@ def delete_object(del_obj: str) -> bool:
 
 def object_metadata(obj: str) -> Dict[str, str]:
     """
-    retrieve s3 object Metadata
+    Retrieve s3 object Metadata
 
     Will throw if object does not exist
 
@@ -193,7 +193,7 @@ def object_metadata(obj: str) -> Dict[str, str]:
 
 def object_exists(obj: str) -> bool:
     """
-    check if s3 object exists
+    Check if s3 object exists
 
     will raise on any error other than "NoSuchKey"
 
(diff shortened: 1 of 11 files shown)


src/lamp_py/aws/s3.py
```diff
@@ -743,3 +742,40 @@ def read_parquet_chunks(
...
+    """
+    Overwrite an existing object with a new file if the new file has at least as many rows as the existing file.
+
+    If the existing file does not exist, the new file will be uploaded.
+
+    :param file_name: local parquet file path to upload
+    :param object_path: S3 object path to upload to (including bucket); if not empty, the existing file must be Parquet
+    :param extra_args: additional arguments passed to upload_file
+
+    :return: None
```
huangh:
```suggestion
    :return: True if uploaded, False otherwise
```

src/lamp_py/aws/s3.py
```diff
@@ -743,3 +742,40 @@ def read_parquet_chunks(
...
+    process_logger = ProcessLogger("replace_remote_parquet", file_name=file_name, object_path=object_path)
+    process_logger.log_start()
+
+    object_path = object_path.replace("s3://", "")
+
+    try:
+        if object_exists(object_path):
+            existing_row_count = pq.read_metadata(object_path, filesystem=fs.S3FileSystem()).num_rows
+            new_row_count = pq.read_metadata(file_name).num_rows
+            assert new_row_count >= existing_row_count, f"{new_row_count} < {existing_row_count}; cancelling upload."
```
huangh: i'm back to feeling this is a potential footgun. asserts aren't meant to be used this way, also they can be environment-variabled off if run in non-debug mode. Which we don't do...but some enterprising engineer might. 


```suggestion
            if new_row_count >= existing_row_count:
                
                raise LampInvalidReplaceError
     
     
     # and down below
     except LampInvalidReplaceError:
         # handle expected exception
         process_logger.log_failure("{new_row_count} < {existing_row_count}; cancelling upload.")
     except Exception as e:
         # handle unexpected exception
         process_logger.log_failure(e)
```

tests/aws/test_s3_utils.py
```diff
@@ -154,3 +153,36 @@ def test_move_bad_objects(s3_stub, caplog):  # type: ignore
             found_error = True
 
     assert found_error
+
+
+@pytest.mark.parametrize(
+    ["remote_file_exists", "local_records", "upload_succeeds", "log_text"],
+    [
+        (True, 10, True, "existing_row_count=10"),
+        (True, 5, True, "<"),
```
huangh: what is "<"?

tests/aws/test_s3_utils.py
```diff
@@ -154,3 +153,36 @@ def test_move_bad_objects(s3_stub, caplog):  # type: ignore
             found_error = True
 
     assert found_error
+
+
+@pytest.mark.parametrize(
+    ["remote_file_exists", "local_records", "upload_succeeds", "log_text"],
+    [
+        (True, 10, True, "existing_row_count=10"),
```
huangh: suggestion: the AI (bless our overlords) have put me onto pytest.param(
           blah, blah, blah, id="forward"
        ),

which i really like because it localizes the "id" with the test case, which makes it easier to see what the intent of a test case should be. 

so suggest to use pytest.param, but i would like to see the id field used.

huangh (changes requested):

huangh (approved):

PR #745, opened 2026-05-11
feat(tests): randomize pytest order and seed

runkelcorey:
In this PR, I avoided updating dataframely. Turns out there's a regression that caused the errors I saw and I just assumed they were intentional.

1. Adds `pytest-randomly`, which randomizes the order that tests run as well as the seed
2. Adds`pytest-order`, which allows us to control the order of tests in specific cases (eg. when tests write their state to the test performance manager database and need to retrieve that state later).
3. Updates Glides schemas to pass tests and get closer to reality
4. Updates VehiclePositions schema to require min length of 1 to pass tests and get closer to reality

1. CI
5. Ran glides ingestion locally to ensure that schema updates aligned with reality

1. None.

### DIFF
--- a/analysis/flashback_performance.py
+++ b/analysis/flashback_performance.py
@@ -25,6 +25,7 @@ def _():
 
     from lamp_py.flashback.events import StopEvents
     from lamp_py.aws.s3 import file_list_from_s3_with_details
+
     return StopEvents, alt, datetime, mo, pl, timedelta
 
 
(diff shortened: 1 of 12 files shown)


analysis/flashback_performance.py
runkelcorey, line 1: black .

huangh (approved):

PR #748, opened 2026-05-13
feat(ingestion): Add `automaticStartTime` and `automaticEndTime` to Glides trip updates

runkelcorey:
Asana Task: [🚋 add automaticStartTime/EndTime to Glides TripUpdates](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1212437463053858?focus=true)

Adds 2 new fields following https://github.com/mbta/schemas/pull/19 and https://github.com/mbta/glides/pull/3319.

1. Ran Glides ingestion locally using `ctd-glides-prod`, saw 0 errors

1. Should the Tableau runner transform this field out of a string format? We could transform to an [Interval](https://tableau.github.io/hyper-db/lang_docs/py/tableauhyperapi.html#tableauhyperapi.TypeTag.INTERVAL)

### DIFF
--- a/src/lamp_py/ingestion/glides.py
+++ b/src/lamp_py/ingestion/glides.py
@@ -127,6 +127,8 @@ class TripUpdatesRecord(GlidesRecord):
                         "endLocation": location,
                         "startTime": dy.String(nullable=True),  # can be "unset" string :(
                         "endTime": dy.String(nullable=True),  # can be "unset" string :(
+                        "automaticStartTime": dy.String(nullable=True, regex=r"unset|" + GTFS_TIME_REGEX),
+                        "automaticEndTime": dy.String(nullable=True, regex=r"unset|" + GTFS_TIME_REGEX),
                         "cars": dy.String(nullable=True),  # an array of objects
                         "revenue": dy.String(nullable=True),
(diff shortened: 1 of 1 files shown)


src/lamp_py/ingestion/glides.py
```diff
@@ -127,6 +127,8 @@ class TripUpdatesRecord(GlidesRecord):
                         "endLocation": location,
                         "startTime": dy.String(nullable=True),  # can be "unset" string :(
                         "endTime": dy.String(nullable=True),  # can be "unset" string :(
+                        "automaticStartTime": dy.String(nullable=True, regex=r"unset|" + GTFS_TIME_REGEX),
+                        "automaticEndTime": dy.String(nullable=True, regex=r"unset|" + GTFS_TIME_REGEX),
```
cho-jos: question: since `startTime` and `endTime` can also be `"unset"`, do those fields need the  `regex=r"unset|" + GTFS_TIME_REGEX` parameter as well? I guess I'm not sure why those would not have the extra param, but the automatic times do

runkelcorey: good catch! they ought to

huangh (approved):

PR #754, opened 2026-05-21
fix(ingestion): Don't treat GTFS_ARCHIVE as parquet

runkelcorey:
#744 caused a [prod ingestion failure](https://mbta.splunkcloud.com/en-US/app/search/show_source?sid=1779372643.126561&offset=0&latest_time=1779332400) when it treated `GTFS_ARCHIVE.db.gz` as a parquet file.

Uses `upload_file` for `GTFS_ARCHIVE.db.gz`.

Didn't. Seems pretty clear from the error.

1. Is there an easy unit test that you see I could add here? The only way I see to do it would be to create a bunch more test files and directories

### DIFF
--- a/src/lamp_py/ingestion/compress_gtfs/gtfs_to_parquet.py
+++ b/src/lamp_py/ingestion/compress_gtfs/gtfs_to_parquet.py
@@ -220,9 +220,9 @@ def compress_gtfs_file(gtfs_table_file: str, schedule_details: ScheduleDetails) 
         pl.concat(
             (old_records, same_records, new_records),
             how="diagonal",
-        ).filter(pl.col("gtfs_end_date") > pl.col("gtfs_active_date")).write_parquet(
-            export_path, use_pyarrow=True, statistics=True
-        )
+        ).filter(
+            pl.col("gtfs_end_date") > pl.col("gtfs_active_date")
+        ).write_parquet(export_path, use_pyarrow=True, statistics=True)
     else:
         #
         # no partition file exists (current or last)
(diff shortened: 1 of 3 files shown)


huangh (dismissed):
Just make sure the rest of the replace_parquet callsites are appropriate. And maybe can check the file extension to only operate on .parquet

src/lamp_py/ingestion/compress_gtfs/gtfs_to_parquet.py
```diff
@@ -297,9 +295,15 @@ def gtfs_to_parquet() -> None:
         year_path = os.path.join(gtfs_tmp_folder, year)
         pq_folder_to_sqlite(year_path)
         for file in os.listdir(year_path):
-            replace_remote_parquet(
-                file_name=os.path.join(year_path, file),
-                object_path=os.path.join(compressed_gtfs.s3_uri, year, file),
-            )
+            if os.path.splitext(file)[1] == ".parquet":
```
huangh, line 298: Wouldn't this be better inside the function? So it can't be misused going forward?

huangh (changes requested):
Last bit

src/lamp_py/ingestion/compress_gtfs/gtfs_to_parquet.py
runkelcorey: I added [a check inside the file](https://github.com/mbta/lamp/pull/754/changes#diff-7b95df5a4edeceae9ed404a940a84a2148f2dd4b063d405549b46eac1503d7a5R771-R772) but I don't think generalizing the function to handle to non-parquet files makes sense. What would that function do that `upload_file` doesn't?

huangh: missed that change in the ruff changes. my bad.

huangh (approved):

PR #756, opened 2026-05-26
feat(rail): export `vehicle_events` and `vehicle_trips` tables as Parquet

runkelcorey:
Asana Task: [🔎 decide on CR tasks - rewrite in polars vs cr queries](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1213810187567352?focus=true)

Writes 2 new tables from Performance Manager:
- `vehicle_events`
- `vehicle_trips`

These are big tables and will slow the updates to all our other Rail PM datasets. But if we want to migrate Rail PM out of a database and into our data lake architecture, we can do so by

1. migrating the subway and commuter rail jobs to the tableau publisher
2. refactoring the SQL queries to run on the parquet exports that Rail PM creates
3. repeating this process with the `vehicle_events` and `vehicle_trips` tables

I neglected to refactor the `HyperRtRail` class because I think spending time improving it doesn't make sense if we're trying to deprecate it.

Writing out these tables may also make it easier to incrementally validate our migration out of database land.

- [x] ran successfully on dev
- [x] verified that runtime of the entire Rail PM reads loop increased by less than 1.5x
- [x] verified that I can run the CR query on the output files using DuckDB

1. Am I overengineering this migration? Should we just refactor the Rail PM code all at once? I fear the validation that comes with that


src/lamp_py/tableau/jobs/rt_rail.py
```diff
@@ -488,3 +488,97 @@ def output_processed_schema(self) -> pyarrow.schema:
...
+class HyperRtVehicleEvents(HyperRtRail):
+    """Export the table `vehicle_events` as a Parquet to migrate off database."""
+
+    def __init__(
+        self,
+        **hyper_job_args: Any,
+    ) -> None:
+        HyperRtRail.__init__(
+            self,
+            ">=",
```
huangh: I hate myself for doing this.

src/lamp_py/tableau/jobs/rt_rail.py
```diff
@@ -488,3 +488,97 @@ def output_processed_schema(self) -> pyarrow.schema:
...
+            RouteType.LIGHT_RAIL,
+            **hyper_job_args,
+        )
+
+        self.table_query = "SELECT * FROM vehicle_trips %s;"
```
huangh: I think the other tables we extract like this we do in daily increments? I fear the prod data will be way larger and this job running once a day will be unworkably long

huangh (commented):
This runs only once a day, right? I think it's a reasonable way forward to be able to see what's in this table. But I think the prod ones are gigantic though...

huangh:
Confirm if it'll blow up in staging or prod. Make adjustments to filter by day if so..

Or convert this to an adhoc so we can see what we're dealing with outside of the standard pipeline.

src/lamp_py/tableau/jobs/rt_rail.py
runkelcorey: agreed

src/lamp_py/tableau/pipeline.py
```diff
@@ -40,6 +42,8 @@
 
 
 PERFORMANCE_MANAGER_JOBS: List[HyperJob] = [
+    Prod_RailMetrics_VehicleEvents_LongTerm,
```
runkelcorey: nit: move these below Glides

huangh: agree. i think to the bottom as they are not operationally used

huangh (approved):

PR #761, opened 2026-06-04
feat(ingestion): Add test coverage for alerts

runkelcorey:
Asana Task: [🔔 handle new & old schemas when ingesting alerts](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1215213383773439)

To safely make changes to alerts, this PR adds consistency and test coverage. Specifically, it:

1. adds dataframely schemas for alerts
2. uses those schemas for testing
3. refactors existing GTFSRTDetail to include abstract methods for table and record schemas

A future PR will actually change the schemas. Eventually, it would be nice to have all RT files typed by dataframely for safer changes.

1. Tests pass
2. Added new files to test files to ensure that the dataframely schemas are compatible with real data

1. None.


src/lamp_py/ingestion/gtfs_rt_detail.py
```diff
@@ -42,3 +57,16 @@ def table_sort_order(self) -> Optional[List[Tuple[str, str]]]:
         TODO: perform additional experiments to optimize sort order of all parquet file types  # pylint: disable=fixme
         """
         return None
+
+
+class GTFSRealtime(dy.Schema):
+    """Abstract GTFS-RT schema. Each feed message contains a header and a list of entities, which can be either TripUpdates, VehiclePositions, or Alerts."""
```
huangh: I know this was just moved, but there's nothing abstract about this class. it's just a base class.

src/lamp_py/ingestion/gtfs_rt_detail.py
```diff
@@ -1,13 +1,18 @@
 from abc import ABC
 from abc import abstractmethod
-from typing import Optional, List, Tuple
+from typing import Generic, Optional, List, Tuple, TypeVar
```
huangh: looks like `TypeVar` is old news. https://realpython.com/python312-typing/#generic-classes-and-functions

src/lamp_py/ingestion/gtfs_rt_detail.py
```diff
@@ -1,13 +1,18 @@
...
 
+import dataframely as dy
 import pyarrow
 
 from lamp_py.ingestion.utils import flatten_table_schema
 
 
-class GTFSRTDetail(ABC):
+TableSchemaT = TypeVar("TableSchemaT", bound=dy.Schema)
```
huangh: binding to just a dy.schema isn't doing anything useful in a typing sense here for you. can you assert that the bound schema is a flat schema somehow?

src/lamp_py/ingestion/gtfs_rt_detail.py
```diff
@@ -19,6 +24,16 @@ def transform_for_write(self, table: pyarrow.table) -> pyarrow.table:
         """modify table schema before write to parquet"""
         return flatten_table_schema(table)
 
+    @property
+    def table_schema(self) -> Optional[type[TableSchemaT]]:
```
huangh: do we expect all instances of `GTFSRTDetail` to implement `table_schema` and `record_schema`? if so, needs     @abstractmethod

runkelcorey: not yet. when/if they ever do, I'll implement it

src/lamp_py/ingestion/gtfs_rt_detail.py
runkelcorey: 😬

tests/ingestion/test_gtfs_rt_converter.py
```diff
@@ -44,8 +44,8 @@ def mock_upload_file(file_name: str, object_path: str, _: dict | None = None) ->
 
 
 def gtfs_rt_factory(
-    schema: type[GTFSRealtime], dy_gen: dy.random.Generator, timestamp: datetime
-) -> dy.DataFrame[GTFSRealtime]:
+    schema: type[FeedMessage], dy_gen: dy.random.Generator, timestamp: datetime
+) -> dy.DataFrame[FeedMessage]:
     """Generate an infinite stream of GTFS-RT dataframes that conform to the provided schema."""
```
huangh: (non-blocking)

the df length is determined by the override length or the sample method default struct.sample(n=1)..so this method as written will never generate an infinte list, only a single record. 

i think it's fine for unit testing, but if the intent is to use it to generate more than that, i don't think this will work.

huangh (dismissed):
why are we doing both bus and alerts at once? should the PR title change to reflect this?

tests/ingestion/test_gtfs_rt_converter.py
runkelcorey: good catch; this adapted the generator I originally used when refactoring ingestion

runkelcorey:
> why are we doing both bus and alerts at once? should the PR title change to reflect this?

`RtBusVehicleDetail` isn't functionally changing. I'm just moving some functionality out of the `config_busloc_vehicle` so that `RtAlertsDetail` imports it from the already shared locations of `gtfs_rt_detail` and `gtfs_rt_structs`

PR #762, opened 2026-06-09
fix(ingestion): Allow trip_descriptor to be null

runkelcorey:
Splunk [logged](https://mbta.splunkcloud.com/en-US/app/search/search?q=search%20index%3Dlamp-staging%20ArrowInvalid&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=&earliest=1781035810.67&latest=1781035810.68&sid=1781041265.26494) an error processing an Alert b/c it required the `trip_descriptor` to be not null.

What it says in the title.

Didn't. Seems trivial.

None.


huangh (approved):

huangh:
Is it supposed to be nullable per spec? Feels like there's almost 2 sets of validator. Valid according to spec, and valid according to we use it by convention. One is a fail fail and one is a warn? Idk

PR #764, opened 2026-06-15
feat(ingestion): Stop deduplicating data

runkelcorey:

1. Remove all the hashing
2. Remove all the deduplicating

1. Updated tests
2. Ran successfully on dev

1. What impacts would you expect on Rail PM?


huangh (changes requested):
I think this is crash looping performance manager.

PR #765, opened 2026-06-18
feat(ingestion): allow `cause_detail` and `effect_detail` to be Translations

runkelcorey:
[🔔 handle new & old schemas when ingesting alerts](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1215213383773439)

Implements temporary logic until Alerts Manager changes schema that

1. allows for `cause_detail` and `effect_detail` to either be strings or `Translations`
2. regardless of the input, outputs a table with
    - `cause_detail.translation`
    - `effect_detail.translation`
    - `cause_detail`
    - `effect_detail`

Not in scope for this PR is [updates to performance manager to actually read the English translations from Alerts in the new schema](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1215214638711739).

1. Updated PM alerts test to ensure that these additions wouldn't cause problems for PM
2. Adds ingestion test to check what happens if data with the new schema is appended to data with the old schema
3. Soaking in dev w/o errors
4. No PM Alerts errors in dev

1. Are you okay with this so long as I create a ticket to clean it up after migration?

### DIFF
--- a/src/lamp_py/ingestion/config_rt_alerts.py
+++ b/src/lamp_py/ingestion/config_rt_alerts.py
@@ -114,8 +114,8 @@ class ProposedAlertsRecord(FeedMessage):
                         "effect": dy.String(nullable=True),
                         "effect_detail": translated_string,
                         "url": translated_string,
-                        "header_text": translated_string,
-                        "description_text": translated_string,
+                        "header_text": with_nullable(translated_string, nullable=False),
+                        "description_text": with_nullable(translated_string, nullable=False),
                         "severity_level": dy.String(nullable=True),
                         "severity": dy.UInt16(nullable=True),  # MBTA Enhanced field
                         "created_timestamp": dy.UInt64(nullable=True),  # MBTA Enhanced field
--- a/src/lamp_py/ingestion/config_rt_alerts.py
+++ b/src/lamp_py/ingestion/config_rt_alerts.py
@@ -75,14 +75,77 @@ class AlertsRecord(FeedMessage):
         min_length=1,
     )
 
+class ProposedAlertsRecord(FeedMessage):
+    """Each Alert message from GTFS-RT."""
+
+    entity = dy.List(
+        inner=dy.Struct(
+            inner={
+                "id": dy.String(min_length=1),
+                "alert": dy.Struct(
+                    inner={
+                        "active_period": dy.List(
+                            dy.Struct(
+                                inner={
+                                    "start": dy.UInt64(nullable=True),
+                                    "end": dy.UInt64(nullable=True),
+                                }
+                            ),
+                            nullable=True,
+                        ),
+                        "informed_entity": dy.List(
+                            dy.Struct(
+                                inner={
+                                    "agency_id": dy.String(nullable=True),
+                                    "route_id": dy.String(nullable=True),
+                                    "route_type": dy.Int32(nullable=True),
+                                    "direction_id": dy.UInt8(nullable=True),
+                                    "trip": with_nullable(trip_descriptor, nullable=True),
+                                    "stop_id": dy.String(nullable=True),
+                                    "facility_id": dy.String(nullable=True),
+                                    "activities": dy.List(dy.String(), nullable=True),  # MBTA Enhanced field
+                                }
+                            ),
+                            nullable=True,
+                        ),
+                        "cause": dy.String(nullable=False),
+                        "cause_detail": translated_string,
+                        "effect": dy.String(nullable=True),
+                        "effect_detail": translated_string,
+                        "url": translated_string,
+                        "header_text": translated_string,
+                        "description_text": translated_string,
+                        "severity_level": dy.String(nullable=True),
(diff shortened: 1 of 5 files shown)


src/lamp_py/ingestion/config_rt_alerts.py
```diff
@@ -76,13 +76,81 @@ class AlertsRecord(FeedMessage):
...
+                                    "trip": with_nullable(trip_descriptor, nullable=True),
+                                    "stop_id": dy.String(nullable=True),
+                                    "facility_id": dy.String(nullable=True),
+                                    "activities": dy.List(dy.String(), nullable=True),  # MBTA Enhanced field
+                                }
+                            ),
+                            nullable=True,
+                        ),
+                        "cause": dy.String(nullable=False),
+                        "cause_detail": translated_string,
```
huangh, line 113: for all these translated_strings that were string(`nullable=True`) before, does the type translated_string let it be `null`?

edit: i see how translated string is implemented - actually is the other way around - have to set the strings that are _not_ nullable.

src/lamp_py/ingestion/convert_gtfs_rt.py
```diff
@@ -334,7 +334,18 @@ def gz_to_pyarrow(self, filename: str) -> Tuple[Optional[datetime], str, Optiona
...
+            except pyarrow.ArrowTypeError as e:
+                if self.config_type == ConfigType.RT_ALERTS:
+                    table = pyarrow.Table.from_pylist(
+                        json_data["entity"],
+                        schema=pyarrow.schema(
+                            [v.pyarrow_field(k) for k, v in ProposedAlertsRecord.entity.inner.inner.items()]  # type: ignore[attr-defined]
+                        ),
+                    )
+                else:
+                    raise e
```
huangh, line 348: catch and print this

src/lamp_py/ingestion/config_rt_alerts.py
```diff
@@ -76,13 +76,81 @@ class AlertsRecord(FeedMessage):
...
+                        "effect": dy.String(nullable=True),
+                        "effect_detail": translated_string,
+                        "url": translated_string,
+                        "header_text": translated_string,
```
huangh: header_text is required per spec. need to set nullable=false for this

src/lamp_py/ingestion/config_rt_alerts.py
```diff
@@ -76,13 +76,81 @@ class AlertsRecord(FeedMessage):
...
+                        "description_text": translated_string,
```
huangh: required

src/lamp_py/ingestion/config_rt_alerts.py
```diff
@@ -76,13 +76,81 @@ class AlertsRecord(FeedMessage):
...
     """Flattened Alerts data."""
 
     alert_cause = dy.String(alias="alert.cause")
     alert_cause_detail = dy.String(nullable=True, alias="alert.cause_detail")
+    alert_cause_detail_translation = with_alias(
+        translated_string.inner["translation"], "alert.cause_detail.translation"
+    )
     alert_effect = dy.String(nullable=True, alias="alert.effect")
     alert_effect_detail = dy.String(nullable=True, alias="alert.effect_detail")
+    alert_effect_detail_translation = with_alias(
```
huangh, line 151: if we're going to flatten the input into a table, why not break out the translation language here as well? perhaps you just need to get a single translation.language column out, and then expect...(really assume..) that all of the strings are translated into the same language in the same alert. 

as it stands, this alert_effect_detail_translation is a struct that will require further unpacking, yes?

huangh (commented):
looks reasonable - decide if any of my comments are valid or if you can live with it - i think the only consequential change would be to further flatten the translation.language out, but if you've already been testing and are okay with the ergonomics of it, it's not a dealbreaker for me. 

i could see arguments against doing what i suggested..

src/lamp_py/ingestion/config_rt_alerts.py
runkelcorey: yeah, I think that assumes too much and requires even more changes for PM

src/lamp_py/ingestion/config_rt_alerts.py
runkelcorey: good catch

src/lamp_py/ingestion/convert_gtfs_rt.py
runkelcorey: won't this be caught by L371 and then logged by https://github.com/mbta/lamp/blob/05e19ae09e1c3e28b05df3c6f568394f2202e8e4/src/lamp_py/ingestion/convert_gtfs_rt.py#L236-L242?

huangh (approved):
lgtm. last thing non-blocking- should we have some sort of logging when the schema gets expanded or some new input comes in? we want to know that it's happened and perhaps adjust the default schema of record to incude these changes eventually? otherwise our documented dataframely definitions will get outdated.

PR #767, opened 2026-06-23
fix(bus): Add `plan_stop_departure_dt` to headway calculations

runkelcorey:
Asana Task: [🚌 Add plan_stop_departure_dt to ordering of headways](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1213441045749769?focus=true)

1. Makes all datetimes UTC (so they are combinable)
2. Adds `plan_stop_departure_dt` to the calculation of headways

1. Wrote a new test dataframely class with a rule that any row with non-empty inputs should have non-empty headways

None.

### DIFF
--- a/src/lamp_py/bus_performance_manager/combined_bus_schedule.py
+++ b/src/lamp_py/bus_performance_manager/combined_bus_schedule.py
@@ -145,7 +145,7 @@ def join_tm_schedule_to_gtfs_schedule(
         )
         .with_columns(
             pl.col("plan_stop_departure_dt")
-            .sub(pl.col("service_date").cast(pl.Datetime))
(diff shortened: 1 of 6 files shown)


huangh:
Did you confirm that all these date times that were tz-less before were all tz-less UTC, and we haven't assigned UTC to EST times? 

I think there's a test in there somewhere that expects the correct relative difference between time fields, but we should ensure that nothing has changed here.

runkelcorey:
> Did you confirm that all these date times that were tz-less before were all tz-less UTC, and we haven't assigned UTC to EST times?

yep!

huangh (approved):

PR #771, opened 2026-06-30
feat(flashback): Use current time for new/updated event timestamps

runkelcorey:
Asana Task: [📸 update `timestamp` to reflect publishing time](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1213954104994853)

Generates the publication timestamp from the event loop rather than deriving it from the events. This lets us compare when flashback evaluated the event to when the event happened and infer from the difference how long it takes events to reach us.

1. Updated tests

None.

### DIFF
--- a/src/lamp_py/flashback/events.py
+++ b/src/lamp_py/flashback/events.py
@@ -1,5 +1,4 @@
 from datetime import datetime, timedelta
-from zoneinfo import ZoneInfo
 
 import dataframely as dy
 import polars as pl
@@ -73,6 +72,7 @@ def unnest_vehicle_positions(vp: dy.DataFrame[VehiclePositions]) -> dy.DataFrame
 def update_records(
     existing_records: dy.DataFrame[StopEvents],
     new_records: dy.DataFrame[StopEvents],
+    publication_timestamp: datetime,
     max_record_age: timedelta,
 ) -> dy.DataFrame[StopEvents]:
     """Return a DataFrame of recent stops using VehiclePositions."""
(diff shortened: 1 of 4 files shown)


src/lamp_py/flashback/events.py
```diff
@@ -119,9 +119,9 @@ def update_records(
                     .over(["start_date", "route_id", "trip_id", "vehicle_id"])
                     .gt(pl.col("stop_sequence")),
                     pl.col("departed").is_null(),
-                ).then(pl.col("timestamp_right").max().over(["start_date", "route_id", "trip_id", "vehicle_id"])),
+                ).then(publication_timestamp.timestamp()),
                 "timestamp",
-                "timestamp_right",
+                publication_timestamp.timestamp(),
             ).alias("timestamp"),
```
huangh, line 127: i'm not sure if this logic makes sense? or am i just reading it wrong

if the case is not true:

coalesce: nothing, "timestamp", publication_timestamp.timestamp() -> "timestamp" is furthest left non-null

if the case is true:

coalesce: publication_timestamp.timestamp(), "timestamp", publication_timestamp.timestamp() -> publication_timestamp.timestamp() is furthest left non-null

So what's the point of the 2nd publication_timestamp?

huangh (changes requested):
i think there's some weird timestamp logic now. 

also, we have all these times in america/new_york..do we also have the epoch time representation? at a glance, it's easier to work with the epoch time without a library, but providing only datetimes would require more tooling to make it work. 

basically is there a need for both the human readable and the machine computable versions to both be here? 

and should the timestamp have `_est` notation? is it typical for times out of our api to be in eastern?

src/lamp_py/flashback/events.py
huangh: new comments clarified my confusion

huangh (approved):

PR #777, opened 2026-07-14
fix(bus): Remove gtfs_last_in_transit_dt from headway calculation

runkelcorey:
Asana Task: [🚌 Add plan_stop_departure_dt to ordering of headways](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1213441045749769)

Removes `gtfs_last_in_transit_dt` from the calculation of headways. After consulting with Ops Analytics, we decided that it was better to remove this field for interpretability at the cost of fewer non-null headways.

1. Ran `bus_events` runner using 2026-01-02, which previously generated issues with negative headways, and saw noen
2. Tests from #767 still succeed

None.


src/lamp_py/bus_performance_manager/events_metrics.py
```diff
@@ -238,21 +243,21 @@ def calculate_derived_bus_performance_metrics(
                     order_by=pl.coalesce(
                         "stop_departure_dt",
                         "stop_arrival_dt",
-                        "gtfs_last_in_transit_dt",
+                        "plan_stop_departure_dt",
                     ),
                 )
             ).alias("route_direction_headway_seconds"),
```
huangh: Are we sure we want to comingle planned and actual headways here? Do we want to put a flag in here noting which source it got the headway from? Wouldn't this result in us reporting planned headways as actual?

runkelcorey: Ops Analytics thinks this calculation better represents the "true" headway, for better or worse. I should change the data dictionary tho and may change the name

huangh: are you making this change? or are we good with it as opsanalytics has determined this is the best way

huangh: i don't want to be blocking this. we've talked, change is desired by downstream. i'm gonna approve this and you can push your comment/name change/data dictionary (if made) w/o further review.

huangh (dismissed):

PR #779, opened 2026-07-15
feat(lightswitch): rename columns to have underscores

rudiejd:
Asana Task: ad-hoc

Problem:
Column names in lightswitch have periods in them. But since period is a reserved character in most SQL dialects, this forces many SQL clients that work with lightswitch/DuckDB to quote the column names. This is bad for the developer experience, and makes it so that people can't use autocomplete.

Solution:
Rename columns in lightswitch to have an underscore instead of a column when the catalog is deployed. 

Added new tests

1. How should we communicate this change to users? 
2. Is it alright that `lamp.read_ymd` will now return different column names than the views?


src/lamp_py/publishing/lightswitch.py
```diff
@@ -2,6 +2,7 @@
 from typing import List
 
 import duckdb
+import pandas as pd
```
runkelcorey: is this used? I don't see it (and, if it is, can you switch to polars?)
```suggestion
```

runkelcorey (changes requested):
one nit

src/lamp_py/publishing/lightswitch.py
rudiejd: it was being used for typing because of duckdb `.to_df()`. I can convert to polars though

runkelcorey (approved):

PR #781, opened 2026-07-20
fix(lightswitch): don't use periods when creating parquet views

rudiejd:
Asana Task: n/a

Ensures that we don't create views of parquet with periods in the column names, and also fixes an error that was preventing us from using the parquet views.

Existing tests - new tests are no longer needed. 

1. Is it alright to run two queries when creating all the views?

### DIFF
--- a/tests/publishing/test_lightswitch.py
+++ b/tests/publishing/test_lightswitch.py
@@ -14,7 +14,6 @@ from lamp_py.publishing.lightswitch import (
     register_read_ymd,
     register_effective_gtfs_timestamps,
     create_schemas,
-    rename_columns_with_periods,
 )
 from lamp_py.runtime_utils.remote_files import S3Location
 from tests.test_resources import rt_vehicle_positions, tm_route_file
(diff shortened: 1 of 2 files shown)


tests/publishing/test_lightswitch.py
runkelcorey, line 1: Can you update the existing tests to check if views are created with `_` in lieu of `.`?

runkelcorey (changes requested):

tests/publishing/test_lightswitch.py
```diff
@@ -90,18 +94,24 @@ def test_register_read_ymd(
     directory_name: str,
     raises: pytest.RaisesExc,
 ) -> None:
-    """It raises errors on directories that don't exist."""
+    """It raises errors on directories that don't exist, and renames columns to have underscores instead of periods"""
```
runkelcorey, line 97: nit
```suggestion
    """It raises errors on directories that don't exist and renames columns to have underscores instead of periods."""
```

runkelcorey (approved):
I'll happily approve again if you commit that nit

tests/publishing/test_lightswitch.py
rudiejd: done!

PR #790, opened 2026-08-12
fix: add more specific DELTA partition for lamp ingestion,

huangh:
original assumptions were that the only partition in `incoming` was `incoming/lamp/delta`, but now there is `incoming/lamp/railds` which breaks our original assumption. specify `delta` for lamp ingestion pipeline.


runkelcorey (dismissed):
dev deployment will validate this

PR #792, opened 2026-08-13
Docs data dictionary

runkelcorey:
Asana Task: [🚌 up-to-date data dictionary](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1213995356092512?focus=true)

1. Adds a function to print a dataframely schema as a Polars DataFrame
2. Updates schemas and the data dictionary

A future PR could regenerate these docs each time Lightswitch is rebuilt.

N/A: just needs review.

1. Are these definitions accurate enough?


Data_Dictionary.md
```diff
@@ -269,65 +269,72 @@ LAMP_ALL_Bus_Events & LAMP_RECENT_Bus_Events have the same data dictionary.\
...
+| `latitude`                                   | double                | false          | true        | Earliest `IN_TRANSIT_TO` latitude for vehicle from GTFS-RT [VehiclePosition](https://gtfs.org/realtime/reference/#message-vehicleposition).                                                                                                          | - `allow_inf`: `False`                              |
+|                                              |                       |                |             |                                                                                                                                                                                                                                                      | - `allow_nan`: `False`                              |
```
huangh, line 322: Why are there spaces here?

src/lamp_py/utils/dataframely.py
```diff
@@ -36,3 +37,27 @@ def unnest_columns(columns: dict[str, dy.Column]) -> dict[str, dy.Column]:
         else:
             new_schema.update({name: col})
     return new_schema
+
+
+def schema_as_frame(schema: dy.Schema) -> pl.DataFrame:
```
huangh, line 42: This is clean 🧼

huangh (approved):
I like it a lot. 

One bit of validation you might want to add is dy rule that ensures that all fields have a metadata.definition so we don't forget in the future.

Data_Dictionary.md
aprokosch: It appears that the `allow_nan` field at the far right has spilled onto a new row when logically it belongs to the previous row. This appears to affect both `latitude` and `longitude`.
Side note: I'm a little sad that they didn't spell the attribute `allow_NaN`

PR #793, opened 2026-08-14
fix(bus): Stop imputing `last_in_transit_dt`

runkelcorey:
We and Ops Analytics realized that this calculation imputes values when `last_in_transit_dt` is `null`.

Adds an if-else to prevent imputing `null` values.

Trivial.

None.

### DIFF
--- a/src/lamp_py/bus_performance_manager/events_gtfs_rt.py
+++ b/src/lamp_py/bus_performance_manager/events_gtfs_rt.py
@@ -263,9 +263,9 @@ def positions_to_events(vehicle_positions: pl.DataFrame) -> dy.DataFrame[GTFSEve
         .with_columns(
             (pl.col("service_date").cast(pl.Datetime) + pl.duration(seconds=pl.col("start_time"))).alias("start_dt"),
             pl.col("gtfs_stop_sequence").count().over(partition_by=["trip_id", "vehicle_id"]).alias("stop_count"),
-            pl.min_horizontal("gtfs_last_in_transit_dt", "gtfs_arrival_dt").alias(
-                "gtfs_last_in_transit_dt"
-            ),  # cap last_in_transit_dt at arrival_dt
+            pl.when(pl.col("gtfs_last_in_transit_dt").is_not_null()).then(
+                pl.min_horizontal("gtfs_last_in_transit_dt", "gtfs_arrival_dt")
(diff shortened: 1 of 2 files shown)


src/lamp_py/runtime_utils/remote_files.py
```diff
@@ -145,7 +145,7 @@ def s3_uri(self) -> str:
 )
 
 # published by LAMP
-bus_events = S3Location(bucket=S3_PUBLIC, prefix=os.path.join(LAMP, "bus_vehicle_events"), version="1.2")
+bus_events = S3Location(bucket=S3_PUBLIC, prefix=os.path.join(LAMP, "bus_vehicle_events"), version="1.2.1")
```
huangh, line 148: Are we able to bump the bus dataset? Or is that still broken

src/lamp_py/bus_performance_manager/events_gtfs_rt.py
```diff
@@ -263,9 +263,9 @@ def positions_to_events(vehicle_positions: pl.DataFrame) -> dy.DataFrame[GTFSEve
         .with_columns(
             (pl.col("service_date").cast(pl.Datetime) + pl.duration(seconds=pl.col("start_time"))).alias("start_dt"),
             pl.col("gtfs_stop_sequence").count().over(partition_by=["trip_id", "vehicle_id"]).alias("stop_count"),
-            pl.min_horizontal("gtfs_last_in_transit_dt", "gtfs_arrival_dt").alias(
-                "gtfs_last_in_transit_dt"
-            ),  # cap last_in_transit_dt at arrival_dt
+            pl.when(pl.col("gtfs_last_in_transit_dt").is_not_null()).then(
+                pl.min_horizontal("gtfs_last_in_transit_dt", "gtfs_arrival_dt")
```
huangh, line 267: If last in transit were null, then it would just take gtfs arrival dt. I don't see how this change affects that logic. Do you want to check if gtfs arrival dt is also non-null?

huangh: Ah. You only want a last in transit value to be available if there was already a last in transit, and you want it to be null otherwise? 

Can we add a comment showing the analysis that explains why this is a better desired state? Why is populating gtfs-arrival here as fallback not desirable?

huangh (approved):

PR #795, opened 2026-08-18
feat(bus): Keep original values for `gtfs_arrival_dt` and `gtfs_departure_dt`

runkelcorey:
Working group needs to be able to reconstruct the `stop_arrival_dt` and `stop_departure_dt` from its components.

Refactors to defer creating `stop_arrival_dt` and `stop_departure_dt` until the last stage.

1. Moved and updated tests
2. Ran bus runner locally

None.


src/lamp_py/bus_performance_manager/events_metrics.py
```diff
@@ -204,15 +211,36 @@ def calculate_derived_bus_performance_metrics(
...
+                        "gtfs_departure_dt",
+                        "tm_actual_departure_dt",
+                        pl.col("gtfs_first_in_transit_dt")
+                        .shift(-1)
+                        .over(partition_by=["trip_id", "tm_pullout_id"], order_by="stop_sequence"),
+                    )
+                )  # use the first in transit dt from the previous stop)
+                .when(pl.col("point_type").eq(pl.lit("end"))) # endpoints
+                .then(pl.lit(None)) # no departure time
+                .otherwise(  # midpoints
```
mreginald, line 225: Hi @runkelcorey, I hope that orientation went well! We tried to do an assessment of the potential changes by comparing staging to prod, and I think a middle ground would be the most useful thing for us to have to test, but we aren't ready to move this proposal into prod yet. 

When you're back, can you please put the following changes into staging for midpoints (lines 225 to 232), so that the stop_departure_dt is the result of coalescing:
1. min_horizontal(`tm_actual_departure_dt`, `gtfs_departure_dt`)
2. the next `gtfs_first_in_transit_dt `(exactly what's in lines 230 to 232, but this is what would be coalesced, instead of including this in the previous step where you take the earliest of the various timestamps. 

_We did some analysis we can talk more about next week, and in the cases where there actually IS a `tm_actual_departure `or a `gtfs_departure_dt`, taking the next stop's `gtfs_first_in_transit_dt `reduced accuracy, so we want to test out just having this be a fallback when the data is null, instead of treating it as one of the equally viable options in the min_horizontal step._ 

3. `gtfs_last_in_transit_dt `NOT `gtfs_first_in_transit_dt` (the final option for the coalesce in line 234)

_The final fallback for the stop departure should be the last in_transit_to record approaching the stop instead of the first recorded timestamp approaching the stop_

Let me know if you want to huddle about any of these proposals before implementing!

PR #801, opened 2026-09-10
feat(performance_manager): publish per-carriage occupancy in subway export

devinmatte:
Asana Task: external contributor

Adds per-carriage occupancy to `subway-on-time-performance-v1`

```
vehicle_consist       "1958|1959|1915|1914"
occupancy_status      "FEW_SEATS_AVAILABLE|STANDING_ROOM_ONLY|STANDING_ROOM_ONLY|STANDING_ROOM_ONLY"
occupancy_percentage  "8|21|15|25"
```

Sampled from the record that sets `vp_stop_timestamp`, so it describes the vehicle on arrival. Includes an `RPM_VERSION` bump.

Unit tests in `tests/performance_manager/test_l0_rt_vehicle_positions.py`; I don't have access to anything else as an external contributor.

1. Based on the code, this data exists in other datasets, just not public ones. Our other current path to getting this data is to listen to the SSE feed and record our own duplicated dataset. That would be substantially less reliable, so we'd prefer this data exist in the LAMP feeds. Is there a better way to access this information alongside the rest of the subway performance data?


runkelcorey:
Thanks @devinmatte! Actions [show](https://app.asana.com/1/15492006741476/project/1208949713596462/task/1218558009019500?focus=true) related unit tests are failing. Based on the logs, I'm guessing test data doesn't contain that field. Do you thin you can fix this?

devinmatte:
Hi @runkelcorey looks like data from 2023 and older fails because it doesn't have any occupancy data, along with the test data needing an update. Pushed a commit to fix it. I wasn't able to run those tests locally so didn't catch it.

